PR #334, opened 2021-07-28
Implement `Withdraw` in the client.

enriquefynn:
Just as the `Deposit` instruction has a client-side implementation, this PR implements the client-side for `Withdraw`.
It's mainly useful for testing, as withdrawals will be done in the browser UI.

### DIFF
--- a/cli/src/helpers.rs
+++ b/cli/src/helpers.rs
@@ -654,6 +654,7 @@ pub fn command_withdraw(
             solido.get_stake_authority(opts.solido_program_id(), opts.solido_address())?;
 
         // Get heaviest validator.
+        // TODO(ruuda): Use `CliError` to handle the following error.
         let heaviest_validator = get_validator_to_withdraw(&solido.validators).map_err(|_| {
             MaintenanceError::new(
                 "The instance has no active validators to withdraw from.".to_owned(),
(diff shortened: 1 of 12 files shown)


cli/src/helpers.rs
```diff
@@ -606,3 +614,88 @@ pub fn command_deposit(
...
+            &config.signer.pubkey(),
+            &solido.st_sol_mint,
+        );
+
+        let stake_authority =
+            solido.get_stake_authority(opts.solido_program_id(), opts.solido_address())?;
+
+        // Get heaviest validator.
+        let heaviest_validator = get_validator_to_withdraw(&solido.validators).map_err(|_| {
+            MaintenanceError::new(
```
ruuda, line 658: It’s not really a maintenance error. I added a catch-all `CliError` in #343 that we can use for this. If you want to merge first you can also add a
```rs
// TODO(ruuda)
```
comment here and then I’ll change it as part of #343.

program/src/error.rs
```diff
@@ -146,6 +146,10 @@ pub enum LidoError {
     /// There is a validator that has more stake than the selected one.
     #[error("ValidatorWithMoreStakeExists")]
     ValidatorWithMoreStakeExists = 37,
+
+    /// The instance has no validators.
+    #[error("EmptySetOfValidators")]
+    EmptySetOfValidators = 38,
```
ruuda, line 152: We already have `NoActiveValidators = 27`. That name is outdated, we don’t need an “active” stake account for the validator, `EmptySetOfValidators` is a better name. But we should merge the two.

ruuda (approved):
LGTM with two comments.

program/src/error.rs
enriquefynn: Great, changed it to `EmptySetOfValidators`

PR #343, opened 2021-08-03
Allow selecting different Ledger keys

ruuda:
@mkaczanowski noticed that whichever usb://ledger url he tried, Solido would always try to use the same key, even for a url that includes a `?key=0` query parameter. I suspect this is because we used `DerivationPath::default()`. This pull request changes that to parse the derivation path from the url as well, which is in line with what Solana's clap-utils do. I hope this will resolve the issue.


enriquefynn (approved):
LGTM!
Thanks for the better errors and bugfix.

enriquefynn:
Just clippy is complaining about some things

ruuda:
I’ll wait for confirmation from @mkaczanowski that this actually fixes the problem before merging.

enriquefynn (changes requested):
I'll test this further, seems not to be working

enriquefynn (approved):
It works, I was confused because Solana prints the wrong public key for the signing.

cli/src/main.rs
```diff
@@ -335,21 +336,40 @@ fn merge_with_config_and_environment(
...
-                        .into_os_string()
-                        .into_string()
-                        .expect("Should have failed before"),
-                )
-                .expect("Failed reading URL."),
-                DerivationPath::default(),
+                locator,
+                derivation_path,
                 &hw_wallet,
                 false,    /* Confirm public key */
```
ruuda, line 374: @mkaczanowski should we change this to true? Do you think it was helpful?

mkaczanowski: yeah, that would be nice. People would know which account they use for signing

PR #344, opened 2021-08-03
Warn that transactions may have executed on error

ruuda:
We just ran into this while testing on testnet, where the RPC returned an error, but the transaction did execute. Add a warning for this particular error, to point this out to the user. I don't know if there are other cases where a transaction may have executed though, should we always warn?

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -224,19 +224,37 @@ impl<'a> SnapshotConfig<'a> {
         signers: &T,
     ) -> snapshot::Result<()> {
         let transaction = self.sign_transaction(instructions, signers)?;
-        let _signature = match self.output_mode {
+        let signature_result = match self.output_mode {
             OutputMode::Text => {
                 // In text mode, we can display a spinner.
                 self.client
-                    .send_and_confirm_transaction_with_spinner(&transaction)?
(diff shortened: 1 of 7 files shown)


cli/src/main.rs
```diff
@@ -214,19 +214,37 @@ impl<'a> SnapshotConfig<'a> {
...
+
+        // Warn the user for one particular footgun.
+        match signature_result {
+            Err(ref err) if error::might_have_executed(err) => {
+                eprintln!(
+                    "Warning: The RPC returned an error, but the transaction \
+                     might still have been executed. Check the signer's address \
+                     on a block explorer before continuing. Beware that timestamps \
+                     shown on explorer.solana.com can be off by weeks, check the slot \
+                     number to confirm whether a transaction is recent."
```
enriquefynn, line 239: Would be nice to have the transaction id. Is this easy to do?!

ruuda: No, it’s not possible :crying_cat_face: I wanted to include it, but the “transaction id” is the `Signature` returned by [`send_transaction`](https://docs.rs/solana-client/1.7.9/solana_client/rpc_client/struct.RpcClient.html#method.send_transaction), you only get it in the success case, not in the error case.

enriquefynn (approved):
LGTM!

PR #345, opened 2021-08-03
Specify mint and Solido keys.

enriquefynn:
Adds the possibility of specifying the minter and Solido keys for the Solido initialization. Also creates the command `show-authorities` to show the Solido authorities given a known public key. When initializing the Solido instance with an existing key, the initialization step changes slightly. Now it is also required to make a new private key for the Solido instance and pass it in the initialization method.

Adds extra check for the mint, now we require that the mint has the Solido authority for minting and hasn't minted any tokens yet (supply is zero).

### DIFF
--- a/cli/src/helpers.rs
+++ b/cli/src/helpers.rs
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: 2021 Chorus One AG
 // SPDX-License-Identifier: GPL-3.0
 
-use std::fmt;
+use std::{fmt, path::PathBuf};
 
 use serde::Serialize;
 use solana_program::{pubkey::Pubkey, system_instruction};
@@ -12,19 +12,21 @@ use solana_sdk::{
 
 use lido::{
     balance::get_validator_to_withdraw,
+    find_authority_program_address,
     metrics::LamportsHistogram,
     state::{Lido, RewardDistribution, Validator},
     token::{Lamports, StLamports},
     util::serialize_b58,
-    MINT_AUTHORITY, RESERVE_ACCOUNT,
+    MINT_AUTHORITY, RESERVE_ACCOUNT, REWARDS_WITHDRAW_AUTHORITY, STAKE_AUTHORITY,
 };
 
 use crate::{
     config::{
-        AddRemoveMaintainerOpts, AddValidatorOpts, CreateSolidoOpts, DepositOpts, ShowSolidoOpts,
-        WithdrawOpts,
+        AddRemoveMaintainerOpts, AddValidatorOpts, CreateSolidoOpts, DepositOpts,
+        ShowSolidoAuthoritiesOpts, ShowSolidoOpts, WithdrawOpts,
     },
     error::MaintenanceError,
+    get_signer,
 };
 use crate::{
     multisig::{get_multisig_program_address, propose_instruction, ProposeInstructionOutput},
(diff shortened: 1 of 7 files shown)


cli/src/main.rs
```diff
@@ -308,6 +312,12 @@ fn main() {
             let output = result.ok_or_abort_with("Failed to show Solido data.");
             print_output(output_mode, &output);
         }
+        SubCommand::ShowAuthorities(solido_pubkey) => {
+            let result =
+                config.with_snapshot(|_config| command_show_solido_authorities(&solido_pubkey));
+            let output = result.ok_or_abort_with("Failed to show Solido data.");
```
glottologist, line 318: Nit:  Shouldn't this be a little more specific to the sub command, i.e. "Failed to show authorities for Solido instance"

cli/src/helpers.rs
```diff
@@ -93,17 +98,27 @@ pub fn command_create_solido(
     config: &mut SnapshotConfig,
     opts: &CreateSolidoOpts,
 ) -> Result<CreateSolidoOutput> {
-    let lido_keypair = Keypair::new();
+    let lido_signer = {
+        if opts.solido_key_path() != &PathBuf::default() {
+            // If we've been given a solido private key, use it to create the solido instance.
+            let signer = get_signer(opts.solido_key_path().clone());
+            signer
```
ruuda, line 105: This can be just `get_signer` without binding it to a variable, right?

cli/src/helpers.rs
```diff
@@ -126,30 +141,37 @@ pub fn command_create_solido(
...
-    // Solana transaction size limits are so low that we need to break our
-    // instructions down into multiple transactions. So set up the mint first,
-    // then continue.
-    let signers = &[&st_sol_mint_keypair, config.signer];
-    config.sign_and_send_transaction(&instructions[..], signers)?;
-    instructions.clear();
-    eprintln!("Did send mint init.");
+    let st_sol_mint_pubkey = {
+        if opts.mint_address() != &Pubkey::default() {
+            // If we've been given a minter address, return its public address.
```
ruuda, line 146: s/public address/public key/?

cli/src/helpers.rs
```diff
@@ -483,6 +505,74 @@ pub fn command_show_solido(
...
+
+    #[serde(serialize_with = "serialize_b58")]
+    pub mint_authority: Pubkey,
+
+    #[serde(serialize_with = "serialize_b58")]
+    pub rewards_withdraw_authority: Pubkey,
+}
+
+impl fmt::Display for ShowSolidoAuthorities {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
ruuda, line 530: The program id and reserve account are part of the struct, but not printed here. Is that intentional?

program/src/logic.rs
```diff
@@ -31,6 +32,40 @@ pub(crate) fn check_rent_exempt(
     }
     Ok(())
 }
+
+pub(crate) fn check_mint(
```
ruuda, line 36: I would add a doc comment that we only need to check the mint during initialization of the Solido instance; otherwise people might think this check is missing in other instruction processor functions.

tests/test_solido.py
```diff
@@ -105,6 +106,52 @@ def approve_and_execute(
...
+    developer_account_owner.pubkey,
+    '--multisig-address',
+    multisig_instance,
+    '--solido-key-path',
+    solido_test_account.keypair_path,
+    '--mint-address',
+    mint_address.pubkey,
+    keypair_path=test_addrs[0].keypair_path,
+)
+
```
ruuda, line 154: I would add a comment below that we don’t use the instance created above in the remainder of the test, that it’s only used to test creating an instance with a separate mint.

ruuda (approved):
:ok_hand: Very nice! I have a few small comments.

cli/src/helpers.rs
enriquefynn: The reserve account is printed, I think is ok that the program id is not printed, since the user has to provide it to call the cmd.

cli/src/main.rs
enriquefynn: Thanks! That was copy-pasta mistake. Changed it.

cli/src/helpers.rs
enriquefynn: Yes, also the following one. Changed it.

ruuda (approved):

PR #346, opened 2021-08-04
Airdrops more times in case of error.

### LINKED ISSUE #342: Test often fail on CI in airdrop step
Builds are flaky, they often fail like this:

```
  export PATH="$HOME/.local/share/solana/install/active_release/bin:$PATH"
  validator=$(tests/start_test_validator.py)
  
  # Perform initial Solana setup.
  solana-keygen new --no-bip39-passphrase --silent
  solana config set --url http://127.0.0.1:8899
  solana airdrop 500.0
  
  tests/coverage.py
  kill $validator
  shell: /usr/bin/bash -e {0}
  env:
    CARGO_TERM_COLOR: always

Wrote new keypair to /home/runner/.config/solana/id.json
Config File: /home/runner/.config/solana/cli/config.yml
RPC URL: http://127.0.0.1:8899 
WebSocket URL: ws://127.0.0.1:8900/ (computed)
Keypair Path: /home/runner/.config/solana/id.json 
Commitment: confirmed 
Requesting airdrop of 500 SOL
Error: RPC request error: cluster version query failed: error sending request for url (http://127.0.0.1:8899/): error trying to connect: tcp connect error: Connection refused (os error 111)
```

I don’t understand why this happens, because `start_test_validator.py` waits for the RPC to be available. I tried to reproduce locally by running this many times:

    v=$(tests/start_test_validator.py) && solana airdrop 500.0 && kill $v && rm -fr test-ledger

But so far I did not manage to reproduce, while on CI it seems to happen in about 1/4 of the runs. :thinking:

enriquefynn:
Our tests are failing because sometimes the airdrop fails. This PR tries to mitigate this by retrying and waiting some time in-between.
Closes #342

### DIFF
--- a/tests/airdrop_lamports.sh
+++ b/tests/airdrop_lamports.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+for i in `seq 1 5`
+do
+    solana airdrop 500.0
+    if [ $? -eq 0 ]
+    then
+        break
+    else
+        echo "Airdrop failed, awaiting 2s before asking again..."
+        sleep 2
+    fi
+done
--- a/tests/airdrop_lamports.sh
+++ b/tests/airdrop_lamports.sh
@@ -6,11 +6,13 @@ do
     then
         break
     else
-        echo "Airdrop failed, awaiting 2s before asking again..."
-        sleep 2
         if [ $i -eq 5 ]
         then
             echo "Airdrop failed after trying 5 times, giving up."
+            exit 1
+        else
+            echo "Airdrop failed, awaiting 2s before asking again..."
+            sleep 2
         fi
     fi
 done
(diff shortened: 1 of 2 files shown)


glottologist (approved):

tests/airdrop_lamports.sh
```diff
@@ -0,0 +1,18 @@
...
+        if [ $i -eq 5 ]
+        then
+            echo "Airdrop failed after trying 5 times, giving up."
+            exit 1
+        else
+            echo "Airdrop failed, awaiting 2s before asking again..."
+            sleep 2
+        fi
+    fi
+done
```
ruuda, line 18: This borders on what I think is still acceptable in Bash before I would reach for a sane language :see_no_evil: But in this case it’s fine.

ruuda (approved):

PR #349, opened 2021-08-06
Limit withdrawal amount per withdrawal to 10% of the validator balance

### LINKED ISSUE #336: Put a maximum on the amount to withdraw per withdrawal
One challenge with our current withdrawal implementation, is that users who withdraw can disturb the stake distribution. We already enforce that withdrawals have to withdraw from the validator furthest above its target balance, but a withdrawal could still drain almost the entire account, and leave the pool unbalanced.

@vshvsh had a great idea to fix this: limit the amount that you can withdraw per withdrawal. Users can still withdraw as much as they like, but they have to break up the amount into multiple withdrawals, which Solido then forces to go to different validators, and this will keep the stake roughly balanced.

When picking the limit, we need to make a trade-off between how much imbalance we tolerate (larger limits mean more imbalance), and how much inconvenience we want to cause (smaller limits mean users need to do more separate withdrawals).

If splitting the withdrawals is a burden though, we could automatically do this for the user from the CLI or frontend.

Proposal: Let’s set the max withdraw amount to 10% of the validator’s balance. With 30 validators, that means if we have 300 depositors who all deposited roughly the same amount, then they can withdraw without having to split up their transaction.

ruuda:
Closes #336.

The change to the program is fairly small, but in order to test it, I refactored the withdrawal tests a bit so there is a shared setup part that can be reused by multiple tests.

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -26,7 +26,7 @@ use crate::{
         ExchangeRate, FeeRecipients, Lido, Maintainers, RewardDistribution, Validator, Validators,
         LIDO_CONSTANT_SIZE, LIDO_VERSION,
     },
-    token::{Lamports, StLamports},
+    token::{Lamports, Rational, StLamports},
     vote_instruction, MINIMUM_STAKE_ACCOUNT_BALANCE, MINT_AUTHORITY, RESERVE_ACCOUNT,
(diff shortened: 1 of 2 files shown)


enriquefynn (approved):
LGTM!

Issue #347, opened 2021-08-05
Validator removal and stake rebalancing


ruuda:
Meeting notes from discussion with @enriquefynn and @glottologist.

* Validator removal is “easy” if we can do unstaking (back to the reserve): then we just only allow removing validators that have no stake. (And no unclaimed credits.) Those are safe to remove.
* So the real problem is unstaking.
* This is also needed anyway for active rebalancing (restoring the stake distribution after a withdrawal, when no new deposits come in).
* We can do a two-stage removal:
  1. Set a flag on the validator to indicate that it should be removed. This should also prevent new stake for that validator. It could also be used by the maintenance bot to signal that it should unstake all funds from this validator.
  2. Actually remove the validator once its stake is zero.
* We can turn `validator.weight` into a boolean to indicate removal, because we don’t plan to have different weights for validators any more anyway.

That covers validator removal, but how do we do unstaking?

 * For the same reason that we need multiple stake accounts for activating stake (the warmup period), we need multiple for deactivating stake. We can use the same trick with the seeds to make a list of deactivating stake accounts.
 * A new `Unstake` instruction could split off some stake from an active stake account into the end of the list of deactivating stake accounts.
 * A new instruction (`WithdrawInactiveStake`?) could withdraw the inactive stake from these stake accounts back into the reserve.
 * Let’s keep `Unstake` gated to maintainers for now, so we can keep the rebalancing logic in the maintenance daemon; we can make that permissionless at a later time.

Post-meeting thought: do we actually need this `WithdrawInactiveStake`, and a new list of stake accounts? `UpdateValidatorBalance` already withdraws inactive stake into the reserve! And for user withdrawals, it doesn’t really matter if they split off from an active or deactivating stake account, it only saves them the trouble if they receive a deactivating stake account. But can deactivating stake be split at all? So maybe just an `Unstake` instruction is sufficient? We may need to change `UpdateValidatorBalance` to bump the start seed if it leaves an account empty.

* If we have `Unstake`, then the maintenance bot can do active rebalancing. For any validators below target, it stakes with them, and for validators above target, it unstakes.
* Problem: What if a user withdraws, which disturbs the balance and makes most validators be above target and one below target? Then the bot would issue an `Unstake` for the other validators, which is the right thing to if nothing happens for the remainder of the epoch. But if a new deposit comes in, staking that may already resolve the imbalance, and then we shouldn’t have deactivating stake, because it takes time to re-activate, and in that time we are missing out on rewards.
[...]

PR #351, opened 2021-08-09
Active flag

enriquefynn:
Steps towards validator's removal #347.
Replaces the validator's weight with the boolean flag `active`.
When the `active` flag is `false`, the validator cannot receive stake deposits. The maintenance bot should also ignore inactive validators and not try to stake more Sol to them.

### DIFF
--- a/program/src/balance.rs
+++ b/program/src/balance.rs
@@ -31,28 +31,33 @@ pub fn get_target_balance(
     );
 
     let total_delegated_lamports: token::Result<Lamports> = validators
-        .iter_entries()
+        .iter_active()
         .map(|v| v.stake_accounts_balance)
         .sum();
 
     let total_lamports = total_delegated_lamports.and_then(|t| t + undelegated_lamports)?;
 
-    let total_weights: u64 = validators.iter_entries().map(|v| v.weight.0 as u64).sum();
-
     // We only want to target validators that are not in the process of being
     // removed. For now, those are all the validators. Once we add validator
     // removal, we need to take the removal flag into account here.
-    let num_active_validators = validators.len() as u64;
+    let num_active_validators = validators.iter_active().count() as u64;
 
-    // Target a weighted distribution. A division by zero means that
-    // `total_weights` is zero, that means there are no active validators.
+    // If `num_active_validators` is zero, that means there are no active
+    // validators.
+    if num_active_validators == 0 {
+        return Err(LidoError::EmptySetOfValidators);
+    }
+    // Target an uniform distribution. A division by zero cannot happen because
+    // of the test above.
     for (target, validator) in target_balance.iter_mut().zip(validators.iter_entries()) {
-        *target = (total_lamports
-            * Rational {
-                numerator: validator.weight.0 as u64,
-                denominator: total_weights,
-            })
-        .map_err(|_| LidoError::EmptySetOfValidators)?
+        if !validator.inactive {
+            *target = (total_lamports
+                * Rational {
+                    numerator: 1,
+                    denominator: num_active_validators,
+                })
+            .expect("Does not happen, error caught when testing if `num_active_validators == 0`")
+        }
     }
 
     // The total lamports to distribute may be slightly larger than the total
@@ -66,22 +71,27 @@ pub fn get_target_balance(
     let mut remainder = (total_lamports - total_lamports_distributed)
         .expect("Does not underflow because we distribute at most total_lamports.");
 
-    assert!(remainder.0 < num_active_validators);
+    assert!(num_active_validators == 0 || remainder.0 < num_active_validators);
 
-    // Distribute the remainder among the first few validators, give them one
-    // Lamport each. This does mean that the validators early in the list are
-    // in a more beneficial position because their stake target is one Lamport
-    // higher, but to put that number into perspective, the transaction fee per
-    // signature is 10k Lamports at the time of writing. Also, there is a minimum
-    // amount we can stake, so in practice, validators will never be as close to
-    // their target that the one Lamport matters anyway.
-    for target in target_balance.iter_mut() {
+    // Distribute the remainder among the first few active validators, give them
+    // one Lamport each. This does mean that the validators early in the list
+    // are in a more beneficial position because their stake target is one
+    // Lamport higher, but to put that number into perspective, the transaction
+    // fee per signature is 10k Lamports at the time of writing. Also, there is
+    // a minimum amount we can stake, so in practice, validators will never be
+    // as close to their target that the one Lamport matters anyway.
+    // for target in target_balance.iter_mut() {
+    for (target, validator) in target_balance.iter_mut().zip(validators.iter_entries()) {
         if remainder == Lamports(0) {
             break;
         }
-        *target = (*target + Lamports(1))
-            .expect("Does not overflow because per-validator balance is at most total_lamports.");
-        remainder = (remainder - Lamports(1)).expect("Does not underflow due to loop condition.");
+        if !validator.inactive {
+            *target = (*target + Lamports(1)).expect(
+                "Does not overflow because per-validator balance is at most total_lamports.",
+            );
+            remainder =
+                (remainder - Lamports(1)).expect("Does not underflow due to loop condition.");
+        }
     }
 
     // Sanity check: now we should have distributed all inputs.
@@ -90,6 +100,7 @@ pub fn get_target_balance(
         .cloned()
         .sum::<token::Result<Lamports>>()
         .expect("Does not overflow, is at most total_lamports.");
+
     assert_eq!(total_lamports_distributed, total_lamports);
 
     Ok(())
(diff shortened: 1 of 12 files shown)


program/src/balance.rs
```diff
@@ -31,28 +31,31 @@ pub fn get_target_balance(
     );
 
     let total_delegated_lamports: token::Result<Lamports> = validators
-        .iter_entries()
+        .iter_active()
```
ruuda, line 34: This needs to iterate all validators, not just the active ones. Right after if we flip the switch, the validator will still have active stake that is part of the Solido balance, it doesn’t disappear.

program/src/balance.rs
```diff
@@ -31,28 +31,31 @@ pub fn get_target_balance(
...
-    // removal, we need to take the removal flag into account here.
-    let num_active_validators = validators.len() as u64;
+    // removed.
+    let num_active_validators = validators.iter_active().count() as u64;
 
-    // Target a weighted distribution. A division by zero means that
-    // `total_weights` is zero, that means there are no active validators.
+    // No active validators.
+    if num_active_validators == 0 {
+        return Err(LidoError::EmptySetOfValidators);
```
ruuda, line 46: The error message here might be misleading now, because the set of validators can be non-empty and we still return this error. If the error is not used anywhere else, we can rename the error.

program/src/balance.rs
```diff
@@ -68,20 +71,25 @@ pub fn get_target_balance(
...
-    // their target that the one Lamport matters anyway.
-    for target in target_balance.iter_mut() {
+    // Distribute the remainder among the first few active validators, give them
+    // one Lamport each. This does mean that the validators early in the list
+    // are in a more beneficial position because their stake target is one
+    // Lamport higher, but to put that number into perspective, the transaction
+    // fee per signature is 10k Lamports at the time of writing. Also, there is
+    // a minimum amount we can stake, so in practice, validators will never be
+    // as close to their target that the one Lamport matters anyway.
+    // for target in target_balance.iter_mut() {
```
ruuda, line 81: This line should be deleted.

program/src/state.rs
```diff
@@ -640,16 +628,15 @@ pub struct Validator {
     /// Sum of the balances of the stake accounts.
     pub stake_accounts_balance: Lamports,
 
-    /// Weight of the validator. Used when calculating the stake amount for
-    /// keeping a weighted balance, also defines the validator's share of fees.
-    pub weight: Weight,
+    /// Controls if a validator is allowed to have new stake deposits.
+    /// When removing a validator, this flag should be set to `true`.
+    pub inactive: bool,
```
ruuda, line 633: With booleans, it is often simpler to avoid a double negation, then you don’t have to work all of that out in your head. In this case, we can change the field to `active: bool`, and have it default to true.

Some evidence for this, is that almost all places in the code where this field is referenced, now write `!validator.inactive`, and then you have to mentally make the step “not inactive ⇒ active”. It would be a bit simpler to just write `validator.active` in those places.

program/src/state.rs
```diff
@@ -640,16 +628,15 @@ pub struct Validator {
...
 }
 
 impl Validator {
-    pub fn new(fee_address: Pubkey, weight: Weight) -> Validator {
+    pub fn new(fee_address: Pubkey) -> Validator {
         Validator {
             fee_address,
-            weight,
             ..Default::default()
```
ruuda, line 640: Let’s also initialize the field explicitly (maybe let’s just initialize all fields explicitly and remove the `Default`), I find it a bit scary to rely on an implicit default for such an important field.

program/src/balance.rs
```diff
@@ -219,40 +228,44 @@ mod test {
...
-    fn get_target_balance_weighted_validators() {
-        let mut validators = Validators::new_fill_default(2);
-        validators.entries[0].entry.weight = Weight(1000);
-        validators.entries[1].entry.weight = Weight(3000);
+    fn get_target_balance_inactive_validator() {
+        // 200 Lamports delegated, but only one active validator,
+        // so all of the target should be with that one validator.
+        let mut validators = Validators::new_fill_default(3);
+        validators.entries[0].entry.stake_accounts_balance = Lamports(100);
+        validators.entries[1].entry.stake_accounts_balance = Lamports(0);
```
ruuda, line 238: If we set this to a non-zero value, that would have caught the bug above where inactive validators are not considered for the current balance.

program/src/balance.rs
```diff
@@ -219,40 +228,44 @@ mod test {
...
-        validators.entries[1].entry.weight = Weight(2000);
-        validators.entries[0].entry.stake_accounts_balance = Lamports(101);
-        validators.entries[1].entry.stake_accounts_balance = Lamports(99);
 
-        let mut targets = [Lamports(0); 2];
-        let undelegated_stake = Lamports(51);
+    #[test]
+    fn get_target_balance_all_inactive() {
+        // 200 Lamports delegated, but only one active validator,
+        // so all of the target should be with that one validator.
```
ruuda, line 257: This comment is not correct.

program/src/balance.rs
```diff
@@ -219,40 +228,44 @@ mod test {
...
+        assert_eq!(targets, [Lamports(200), Lamports(0), Lamports(200)]);
 
         assert_eq!(
             get_validator_furthest_below_target(&validators, &targets[..]),
-            (1, Lamports(75))
+            (0, Lamports(100))
         );
     }
-    #[test]
-    fn get_target_balance_weighted_validators_non_integer_multiple() {
```
ruuda, line 240: We should keep the non-integer multiple test, it was testing the distribution of the additional lamports at the end. In the new test above, with 400 lamports and 2 active validators, the total is a multiple of the number of validators, so there is no remainder to divide. We should also test the case where the total is not a multiple, and the target of some validators is therefore larger than that of others.

It should be possible to see this in the code coverage results too. You can run `tests/coverage.py` locally, and that will produce an html report where you can see which lines of the distribution logic are covered.

ruuda (commented):
The approach looks good! I have a few comments about the details.

program/src/state.rs
enriquefynn: I originally did it like that, but then I'd lose the default `entry` constructor from `AccountMap`. But we could also implement the custom default for a `Validator`.

program/src/balance.rs
enriquefynn: Oops... will do that, thanks!

enriquefynn: Wouldn't this test be equal to the `get_target_balance_works_for_non_integer_multiple`?

program/src/state.rs
ruuda: :+1: Let’s implement `Default` manually for `Validator` then.

program/src/balance.rs
ruuda: There is a small difference: when one of the validators is inactive, distributing the remainder should skip over those. To test it properly, we need to make sure then that the inactive validator(s) are early in the list, such that there is a validator behind an inactive one, that gets one additional lamport.

ruuda:
Let’s not merge this to `main` yet, with the testnet dry-run going on. At the moment, `main` is still compatible with the program that is deployed to devnet, but this would break that. We can make a new branch `next` or something like that for now, and continue development of v0.4.0-incompatible features on that branch. Then after the testnet run is complete, we can merge it into `main`.

enriquefynn:
> Let’s not merge this to main yet, with the testnet dry-run going on. At the moment, main is still compatible with the program that is deployed to devnet, but this would break that. We can make a new branch next or something like that for now, and continue development of v0.4.0-incompatible features on that branch. Then after the testnet run is complete, we can merge it into main.

Great, I modified the base branch, let's do it like that.

program/src/balance.rs
```diff
@@ -220,39 +226,61 @@ mod test {
...
+
     #[test]
-    fn get_target_balance_weighted_validators_non_integer_multiple() {
-        let mut validators = Validators::new_fill_default(2);
-        validators.entries[0].entry.weight = Weight(1000);
-        validators.entries[1].entry.weight = Weight(2000);
-        validators.entries[0].entry.stake_accounts_balance = Lamports(101);
-        validators.entries[1].entry.stake_accounts_balance = Lamports(99);
+    fn get_target_balance_works_with_inactive_for_integer_multiple() {
+        // 500 Lamports delegated, but only two active validator out of three.
```
ruuda, line 250: s/validator/validators/

program/src/state.rs
```diff
@@ -640,16 +626,15 @@ pub struct Validator {
     /// Sum of the balances of the stake accounts.
     pub stake_accounts_balance: Lamports,
 
-    /// Weight of the validator. Used when calculating the stake amount for
-    /// keeping a weighted balance, also defines the validator's share of fees.
-    pub weight: Weight,
+    /// Controls if a validator is allowed to have new stake deposits.
+    /// When removing a validator, this flag should be set to `true`.
```
ruuda, line 630: s/true/false/

ruuda (approved):
Found two typos, aside from those, LGTM! :shipit:

Issue #222, opened 2021-06-22
Check writable attributes from instructions


enriquefynn:
Some instructions may not have the right writable attributes, for example, in the `StakeDepositAccounts`, I believe that `deposit_authority`'s `is_writable` flag can be set to false.

glottologist:
Fixed in PR #352

PR #352, opened 2021-08-10
222 check writable attributes from instructions

glottologist:
I went through all the accounts and added comments against the is_writable fields to hopefully make it easier to reason about why the said accounts are writable.  

Additionally set the stake_authority.is_writable : false in StakeDepositAccounts as per original comment on #222

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -609,6 +609,7 @@ accounts_struct! {
         },
         pub st_sol_mint {
             is_signer: false,
+            // Is writable due to fee mint (spl_token::instruction::mint_to)
             is_writable: true,
         },
         pub mint_authority {
@@ -617,6 +618,7 @@ accounts_struct! {
         },
         pub validator_fee_st_sol_account {
             is_signer: false,
+            // Is writable due to fee mint (spl_token::instruction::mint_to)
             is_writable: true,
         },
         const spl_token = spl_token::id(),
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -429,6 +429,7 @@ accounts_struct! {
         // Needs to be writable so we withdraw from it.
         pub validator_vote_account {
             is_signer: false,
+            // Is writable due to withdraw to reserve (vote_instruction::withdraw)
             is_writable: true,
         },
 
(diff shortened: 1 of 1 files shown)


program/src/instruction.rs
```diff
@@ -151,18 +151,26 @@ accounts_struct! {
...
         },
         pub recipient {
             is_signer: false,
+            // Is writable due to mint to (spl_token::instruction::mint_to) recipient from
+            // st_sol_mint
             is_writable: true,
         },
         pub st_sol_mint {
             is_signer: false,
+            // Is writable due to mint to (spl_token::instruction::mint_to) recipient from
```
ruuda, line 166: It is writable because of the `mint_to`, but not due to the recipient, it’s because the mint stores the total supply.

program/src/instruction.rs
```diff
@@ -214,11 +224,14 @@ accounts_struct! {
         // Stake account to withdraw from.
         pub source_stake_account {
             is_signer: false,
+            // Is writable due to spliti stake (solana_program::stake::instruction::split)
```
ruuda, line 227: s/spliti/split/

program/src/instruction.rs
```diff
@@ -408,13 +429,15 @@ accounts_struct! {
         // Needs to be writable so we withdraw from it.
         pub validator_vote_account {
             is_signer: false,
+            // Is writable due to withdraw to reserve (vote_instruction::withdraw)
```
ruuda, line 432: There’s already a comment about it three lines above; let’s remove that comment then.

ruuda (approved):
Nice! LGTM with three nits.

program/src/instruction.rs
glottologist: image

But the non-read-only accounts In mint_to (above) are the st_sol_mint and the recipient.

program/src/instruction.rs
glottologist: fixed

program/src/instruction.rs
glottologist: fixed

enriquefynn (approved):
LGTM!

PR #355, opened 2021-08-11
Pretty-print `SignerError`s

ruuda:
In particular, add a hint about the Ledger "blind signing" setting, to make the error a bit more actionable for the user.

@enriquefynn I don’t have a Ledger at hand at the moment, can you verify that this prints the right hint when the "blind signing" setting is disabled?

### DIFF
--- a/cli/src/error.rs
+++ b/cli/src/error.rs
@@ -10,6 +10,8 @@ use solana_program::instruction::InstructionError;
 use solana_program::program_error::ProgramError;
 use solana_program::pubkey::PubkeyError;
 use solana_sdk::pubkey::Pubkey;
+use solana_sdk::signer::presigner::PresignerError;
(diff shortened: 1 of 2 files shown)


enriquefynn:
The hint works when "blind signing" is off. It prints a nice error msg.
When the "blind signing" is on and the user rejected providing the public key, there's an error at `cli/src/main.rs:424:14`: `.expect("Failed to contact remote wallet"),` in the `generate_remote_keypair` function.

enriquefynn (approved):
Maybe we can also pretty print `RemoteWalletError`?! But this feels like a rabbit-hole we're entering. I'd merge this as it is for now.

ruuda:
> Maybe we can also pretty print RemoteWalletError?!

Feel free to open a pull request :innocent:

PR #356, opened 2021-08-11
Fix support for usb://ledger keypair path without ?key query param

ruuda:
With #343, we added support for `?key=` derivation paths in `usb://ledger` keypair paths, but I broke support for the bare `usb://ledger` path. This should fix that, and it adds a test to ensure that we can parse both.

@fkbenjamin, can you check whether you can use the `usb://ledger` path again with this version?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3633,6 +3633,7 @@ dependencies = [
  "borsh",
  "bs58 0.4.0",
  "clap 3.0.0-beta.2",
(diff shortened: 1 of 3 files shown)


fkbenjamin:
Yes, can confirm that this branch works for the "usb://ledger" key path!

enriquefynn:
Great! Just some clippy complaints.

PR #357, opened 2021-08-11
Add token transfer to the client.

### LINKED ISSUE #320: Add multisig support for SPL token transfers
@FelixLutsch mentioned that for the incentive program, we may need the multisig to distribute wrapped LDO tokens. To support that, we need:

* A command to propose an SPL token transfer to the multisig
* A way to display it in `multisig show-transaction`

enriquefynn:
Add a new subcommand `token` to the multisig subcommand. In the future we may support more token instructions, e.g., burning, creating. For now we support only token transfers.

Note: when doing just the `multisig token` params, there's a clap error: "`Option::unwrap()` on a `None` value".
I'm still verifying what can be done, but it works well with the `--help` flag. 🤷 

Closes #320

### DIFF
--- a/cli/src/multisig.rs
+++ b/cli/src/multisig.rs
@@ -31,7 +31,7 @@ use multisig::instruction as multisig_instruction;
 use crate::config::ConfigFile;
 use crate::config::{
     ApproveOpts, CreateMultisigOpts, ExecuteTransactionOpts, ProposeChangeMultisigOpts,
-    ProposeUpgradeOpts, ShowMultisigOpts, ShowTransactionOpts,
+    ProposeUpgradeOpts, ShowMultisigOpts, ShowTransactionOpts, TransferTokenOpts,
 };
 use crate::error::{Abort, AsPrettyError};
 use crate::print_output;
@@ -42,6 +42,12 @@ use crate::{SnapshotClientConfig, SnapshotConfig};
 pub struct MultisigOpts {
     #[clap(subcommand)]
     subcommand: SubCommand,
+}
+
+#[derive(Clap, Debug)]
+pub struct TokenOpts {
+    #[clap(subcommand)]
+    subcommand: TokenSubCommand,
 }
 
 impl MultisigOpts {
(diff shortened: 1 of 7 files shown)


cli/src/multisig.rs
```diff
@@ -534,6 +574,24 @@ impl fmt::Display for ShowTransactionOutput {
...
+                match token_instruction {
+                    TokenInstruction::Transfer {
+                        from_address,
+                        to_address,
+                        amount,
+                    } => {
+                        writeln!(f, "It transfers tokens owned by the multisig.")?;
+                        writeln!(f, "    From address: {}", from_address)?;
+                        writeln!(f, "    To address:   {}", to_address)?;
+                        writeln!(f, "    Amount:       {}", amount)?;
```
ruuda, line 588: Would be good to clarify that this is the amount of the smallest denomination. E.g. if you do it for stSOL, it’s not the amount of stSOL but the amount of stLamports.

The mint also stores the number of decimals, maybe we can use that to format, but on the other hand, maybe that makes it ambiguous too :confused: 

Just clarifying that it is the smallest denomination should work. Also, should we print the token mint address of these accounts, for convenience?

program/tests/tests/withdrawals.rs
```diff
@@ -4,6 +4,7 @@
 #![cfg(feature = "test-bpf")]
 
 use bincode::deserialize;
+use solana_program::program_error::ProgramError;
```
ruuda, line 7: Is this change needed?

tests/util.py
```diff
@@ -193,7 +193,7 @@ def create_vote_account(
     return test_account
 
 
-def create_spl_token(owner_keypair_fname: str, minter: str) -> str:
+def create_spl_token_account(owner_keypair_fname: str, minter: str) -> str:
```
ruuda, line 196: :+1: the new name is much better!

ruuda (approved):
:+1: I have one small suggestion, but aside from that, this looks good to me.

Is the problem with the unwrap fixed? Nothing weird stood out to me in the PR.

cli/src/multisig.rs
enriquefynn: I like printing the mint address, I'll add these changes.
Something like "Amount (in smallest denomination): " would work?

program/tests/tests/withdrawals.rs
enriquefynn: Good catch, I was testing some things and that was left.

cli/src/multisig.rs
ruuda: :+1: yeah that works. If you like to keep then aligned, we could also do “Amount: X of the token’s smallest denomination”

Issue #177, opened 2021-06-15
Inspect why error no more includes a human-readable message 


malikankit:
Cross-referencing comment (3 occurences) in `tests/test_multisig.py`

```
    # TODO: Previously the error included a human-readable message, why does it
    # only include the error code now? Something to do with different Anchor
    # versions?
 ```

**Comment by ruuda** [comment]
I think it’s a problem with Anchor after all, maybe a bug introduced in a later version.

```rust
    let program_error = ProgramError::from(multisig::ErrorCode::AlreadyExecuted);
    println!("{:?}", program_error);
    let anchor_error = multisig::Error::from(program_error);
    println!("{:?}", anchor_error);
```
prints
```
Custom(105)
ProgramError(Custom(105))
```
while I would have expected it to print
```
Custom(105)
ErrorCode(AlreadyExecuted)
```
it seems that converting back the error code to the structured error is broken.

ruuda:
I believe I found the cause of this: https://github.com/project-serum/anchor/pull/354#issuecomment-897717786

ruuda:
#360 adds a workaround for this problem, so let’s close the issue.

PR #359, opened 2021-08-12
Update to Solana 1.7.8, Serum Multisig 0.6.0, update some dependencies

### LINKED ISSUE #294: Borsh dependency is out of date

### LINKED ISSUE #358: Resolve `cargo audit` issues
I just ran `cargo-audit`, and we do depend on a few libraries that have known security vulnerabilities. Some could in theory affect how the CLI connects to the RPC, so if you connect to a public RPC instead of your own validator node, then this might be a problem. I don’t know if it is in practice, but it’s easier to just update.

Some other issues pointed out by it are not relevant to us, but still, if `cargo audit` reporting issues scares auditors, it’s easier to just resolve these issues, than to figure out if they are actually a problem for us.

We should also run `cargo audit` on CI in some for or another. On the one hand, I don’t like that external things can break our build. But on the other hand, I don’t know of a better way to do this at the moment. We do have “Dependabot alerts — Receive alerts of new vulnerabilities that affect your dependencies.” enabled in the repository security settings, and I just turned on “Dependabot security updates — Easily upgrade to non-vulnerable dependencies.” as well. (I hope it’s not as spammy as the regular dependabot.) But so far these haven’t reported anything, so apparently they don’t check all the things that `cargo audit` checks.

ruuda:
This updates many things. Our dependencies put quite some constraints on each other, so we have to update a lot at once.

 * The `multisig` submodule, from a fork of v0.4.0, to a fork on top of v0.6.0. I rebased the changes that we have. Some of them have been integrated upstream. It looks like the Serum Multisig program may be preparing for a release as a crate, so perhaps in the future we can pull it in that way.
   * This version renames the crate from `multisig` to `serum-multisig`, some changes to our code are needed to accommodate that.
   * This versions bumps the `anchor-lang` dependency from 0.4.4 to 0.13.0.
   * That in turn bumps the `solana-sdk` dependency to `=1.7.8`.
   * This version also depends on a newer version of `borsh`.
  
* Solana from 1.7.3 to 1.7.8. The latest version at the time of writing is 1.7.10, but because `anchor-lang` has an `=1.7.8` dependency, Cargo refuses any other version than 1.7.8 :confused: So we are stuck with whatever Anchor uses. (If needed, we could fork Anchor and override this, but for now, 1.7.8 is fine.)
  * This version also depends on a newer version of `borsh`.
  * Validation rewards change slightly. Probably they changed the computation, and the rounding errors are different now. I updated the tests to match the new behavior.

* Update `borsh` from 0.8 to 0.9.1. Fixes #294.

* Update transitive dependencies that had issues according to `cargo audit`. This output of `cargo audit` changes as follows:
  * Vulnerabilities: from 5 to 0.
  * Unmaintained crate warnings: from 3 to 2 (`net2` via `solana-client` and `failure` via `solana-sdk` remain).
 This fixes #358.

* Delete a few unused dependencies from `Cargo.toml`.

* Targets the `next` branch because it is incompatible with v0.4.0 that is currently deployed on testnet.


program/tests/tests/collect_validator_fee.rs
```diff
@@ -86,7 +86,7 @@ async fn test_collect_validator_fee() {
     // configuration yields, so we have to deal with it.
     context.update_exchange_rate().await;
     let rewards = context.collect_validator_fee(validator.vote_account).await;
-    assert_eq!(rewards, Lamports(1246_030_107_210));
+    assert_eq!(rewards, Lamports(1_246_030_615_387));
```
enriquefynn, line 89: Is this because of https://github.com/solana-labs/solana/issues/18942?

enriquefynn (approved):
LGTM, with some minor observations.

enriquefynn:
Some multisig tests are also failing.

program/tests/tests/collect_validator_fee.rs
ruuda: Might be, I’m not sure ... if you would like to know, you can bisect Solana :see_no_evil:

ruuda:
> Some multisig tests are also failing.

Ah! I was hoping that this also fixed #177 ... but it turns out the error codes are just different now, they still don’t deserialize from error codes :crying_cat_face:

PR #368, opened 2021-08-16
Make listen an option to run-maintainer

ruuda:
I noticed while reviewing our deployment of the maintainer that the listen address is a positional argument rather than one with a long-form flag. I think this was an oversight, make it `--listen` instead.


glottologist (approved):
LGTM!

PR #370, opened 2021-08-17
Add `Deactivate` instruction

ruuda:
This implements part 1 of #364: adding a `Deactivate` instruction.

I will open a follow-up pull request to allow proposing a multisig transaction for this.

Issue #364, opened 2021-08-16
Add instruction to deactivate a validator, add command to propose a multisig transaction for it


ruuda:
Parent issue: #347.

After #351, validators now have an `active` flag. It is true initially, and when it is false, new deposits to this validator are blocked, and its stake balance target becomes zero.

What we still need to make this flag useful:

* Add an instruction to set the flag to `false` for a given validator, gated on manager approval.
* Add a command to the CLI to propose a multisig transaction to call that instruction.

ruuda:
This is done, see the two pull requests referenced above.

PR #371, opened 2021-08-17
Add `solido deactivate-validator` subcommand

ruuda:
This is a follow-up to #370 and implements the final part of #364:

* Add support for the `Deactivate` instruction in `solido multisig show-transaction`.
* Add `solido deactivate-validator` to propose a `Deactivate` to the multisig.
* Test it in `test_solido.py`.
* As a drive-by fix, fix some of the command descriptions.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -359,6 +359,30 @@ cli_opt_struct! {
 }
 
 cli_opt_struct! {
+    DeactivateValidatorOpts {
+        /// Address of the Solido program.
+        #[clap(long, value_name = "address")]
+        solido_program_id: Pubkey,
+
+        /// Account that stores the data for this Solido instance.
+        #[clap(long, value_name = "address")]
+        solido_address: Pubkey,
+
+        /// Address of the validator vote account.
+        #[clap(long, value_name = "address")]
+        validator_vote_account: Pubkey,
+
+        /// Multisig instance.
+        #[clap(long, value_name = "address")]
+        multisig_address: Pubkey,
+
+        /// Address of the Multisig program.
+        #[clap(long)]
+        multisig_program_id: Pubkey,
+    }
+}
+
+cli_opt_struct! {
     AddRemoveMaintainerOpts {
         /// Address of the Solido program.
         #[clap(long, value_name = "address")]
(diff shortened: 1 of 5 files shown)


cli/src/config.rs
```diff
@@ -358,6 +358,30 @@ cli_opt_struct! {
...
+        #[clap(long, value_name = "address")]
+        validator_vote_account: Pubkey,
+
+        /// Multisig instance.
+        #[clap(long, value_name = "address")]
+        multisig_address: Pubkey,
+
+        /// Address of the Multisig program.
+        #[clap(long)]
+        multisig_program_id: Pubkey,
```
enriquefynn, line 381: Is there a reason this doesn't have the attr `value_name = "address"`?

cli/src/main.rs
```diff
@@ -129,10 +129,13 @@ REWARDS
...
     AddValidator(AddValidatorOpts),
 
+    /// Deactivates a validator and initiates the removal process
+    DeactivateValidator(DeactivateValidatorOpts),
+
     /// Adds a maintainer to the Solido instance
     AddMaintainer(AddRemoveMaintainerOpts),
 
-    /// Adds a maintainer to the Solido instance
+    /// Removes a maintainer from the Solido instance
```
enriquefynn, line 138: Good catch 😅

cli/src/multisig.rs
```diff
@@ -18,11 +18,10 @@ use solana_sdk::signature::{Keypair, Signature, Signer};
...
-use lido::instruction::ChangeRewardDistributionMeta;
-use lido::instruction::LidoInstruction;
-use lido::instruction::RemoveMaintainerMeta;
+use lido::instruction::{
+    AddMaintainerMeta, AddValidatorMeta, ChangeRewardDistributionMeta, DeactivateValidatorMeta,
+    LidoInstruction, RemoveMaintainerMeta,
+};
 use lido::state::FeeRecipients;
 use lido::state::Lido;
 use lido::state::RewardDistribution;
```
enriquefynn, line 27: Can you also merge the rest of these modules?

enriquefynn (approved):
LGTM with some comments

cli/src/config.rs
ruuda: This was an oversight, fixed!

cli/src/multisig.rs
ruuda: Done.

PR #373, opened 2021-08-18
Implement the `Unstake` instruction.

enriquefynn:
Add a seeds for the unstake accounts and change some variables to cope with the modifications.
Changes:

- Split `VALIDATOR_STAKE_ACCOUNT` in `VALIDATOR_STAKE_ACCOUNT` and `VALIDATOR_UNSTAKE_ACCOUNT`.
- Function to split stake accounts `split_stake_account`, so it can be reused in `process_withdraw` and `process_unstake`.
- Bugfix (please check that): the inactive stake calculation at `StakeAccount::from_delegated_account` was subtracted twice, once from the `activating_lamports` and another for the `deactivating_lamports`. In other words, when the stake is being deactivated, the `activating_lamports` is not subtracted.
- New type `SeedRange` that is reused for `stake_seeds` and `unstake_seeds` for the `Validator`'s state.
- Split `find_stake_account_address` into `find_stake_account_address` and `find_unstake_account_address`. Helpers for finding stake accounts.
- Simple test to unstake.

The ability to retrieve funds from the inactive stake accounts will be implemented on top of this PR. Maybe we can modify the `WithdrawInactiveStake` instruction to do so.

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -367,18 +367,16 @@ accounts_struct! {
         // the given validator, with seed `stake_seeds.stake_accounts_seed_end - 1`.
         pub source_stake_account {
             is_signer: false,
-            // Is writable due to split (`stake_program::intruction::split`) of stake_account_end
-            // into stake_account_merge_into under the condition that they are not equal
+            // Is writable due to split (`stake_program::intruction::split`).
             is_writable: true,
         },
         // Destination stake account is the last unstake stake account that will
-        // receive the split of the funds.  Determined by the program-derived
+        // receive the split of the funds. Determined by the program-derived
         // stake account for the given validator, with seed
         // `unstake_seeds.stake_accounts_seed_end`.
         pub destination_stake_account {
             is_signer: false,
-            // Is writable due to the first two instructions from split
-            // (`stake_program::intruction::split`).
+            // Is writable due to the first two instructions from split.
             is_writable: true,
         },
         // Stake authority, to be able to split the stake.
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -364,7 +364,7 @@ accounts_struct! {
         },
         // Source stake account is the last active stake account that we'll try
         // to unstake from.  Determined by the program-derived stake account for
-        // the given validator, with seed `stake_seeds.stake_accounts_seed_end - 1`.
+        // the given validator, with seed `stake_seeds.begin`.
         pub source_stake_account {
             is_signer: false,
             // Is writable due to split (`stake_program::intruction::split`).
@@ -372,8 +372,7 @@ accounts_struct! {
         },
         // Destination stake account is the last unstake stake account that will
         // receive the split of the funds. Determined by the program-derived
-        // stake account for the given validator, with seed
-        // `unstake_seeds.stake_accounts_seed_end`.
+        // stake account for the given validator, with seed `unstake_seeds.end`.
         pub destination_stake_account {
             is_signer: false,
             // Is writable due to the first two instructions from split.
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -304,7 +304,7 @@ accounts_struct! {
         // For a `StakeDeposit` where we temporarily create an undelegated
         // account at `stake_account_end`, but immediately merge it into
         // `stake_account_merge_into`, this must be set to the program-derived
-        // stake account for the validator, with seed `stake_accounts_seed_end
+        // stake account for the validator, with seed `stake_seed.end
         // - 1`. For a `StakeDeposit` where we create a new stake account, this
         // should be set to the same value as `stake_account_end`.
         pub stake_account_merge_into {
@@ -314,7 +314,7 @@ accounts_struct! {
             is_writable: true,
         },
         // Must be set to the program-derived stake account for the given
-        // validator, with seed `stake_accounts_seed_end`.
+        // validator, with seed `stake_seeds.end`.
         pub stake_account_end {
             is_signer: false,
             // Is writable due to transfer (system_instruction::transfer) from reserve_account to
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -362,7 +362,7 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
-        // Source stake account is the last active stake account that we'll try
+        // Source stake account is the oldest active stake account that we'll try
         // to unstake from.  Determined by the program-derived stake account for
         // the given validator, with seed `stake_seeds.begin`.
         pub source_stake_account {
@@ -370,7 +370,7 @@ accounts_struct! {
             // Is writable due to split (`stake_program::intruction::split`).
             is_writable: true,
         },
-        // Destination stake account is the last unstake stake account that will
+        // Destination stake account is the oldest unstake stake account that will
         // receive the split of the funds. Determined by the program-derived
         // stake account for the given validator, with seed `unstake_seeds.end`.
         pub destination_stake_account {
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -362,33 +362,36 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
-        // Source stake account is the first, possibly active stake account that we'll try to unstake from.
-        // Determined by the
+        // Source stake account is the last active stake account that we'll try
+        // to unstake from.  Determined by the program-derived stake account for
+        // the given validator, with seed `stake_seeds.stake_accounts_seed_end - 1`.
         pub source_stake_account {
             is_signer: false,
-            // Is writable due to merge (stake_program::intruction::merge) of stake_account_end
+            // Is writable due to split (`stake_program::intruction::split`) of stake_account_end
             // into stake_account_merge_into under the condition that they are not equal
             is_writable: true,
         },
-        // Must be set to the program-derived stake account for the given
-        // validator, with seed `stake_accounts_seed_end`.
+        // Destination stake account is the last unstake stake account that will
+        // receive the split of the funds.  Determined by the program-derived
+        // stake account for the given validator, with seed
+        // `unstake_seeds.stake_accounts_seed_end`.
         pub destination_stake_account {
             is_signer: false,
-            // Is writable due to transfer (system_instruction::transfer) from reserve_account to
-            // stake_account_end and stake program being initialized
-            // (stake_program::instruction::initialize)
+            // Is writable due to the first two instructions from split
+            // (`stake_program::intruction::split`).
             is_writable: true,
         },
+        // Stake authority, to be able to split the stake.
         pub stake_authority {
             is_signer: false,
             is_writable: false,
         },
+        // Required to call `solana_program::stake::instruction::deactivate_stake`.
         const sysvar_clock = sysvar::clock::id(),
+        // Required to call cross-program.
         const system_program = system_program::id(),
-        const sysvar_rent = sysvar::rent::id(),
+        // Required to call `stake_program::intruction::split`.
         const stake_program = stake_program::program::id(),
-        const stake_history = stake_history::id(),
-        const stake_program_config = stake_program::config::id(),
     }
 }
 
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -56,10 +56,7 @@ pub enum LidoInstruction {
         #[allow(dead_code)] // but it's not
         amount: Lamports,
     },
-    /// Unstake from a validator to a new stake account
-    ///
-    /// Splits the stake account determined by the validator's inactive stake
-    /// seeds.
+    /// Unstake from a validator to a new stake account.
     Unstake {
         #[allow(dead_code)] // but it's not
         amount: Lamports,
@@ -365,12 +362,8 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
-        // For a `StakeDeposit` where we temporarily create an undelegated
-        // account at `stake_account_end`, but immediately merge it into
-        // `stake_account_merge_into`, this must be set to the program-derived
-        // stake account for the validator, with seed `stake_accounts_seed_end
-        // - 1`. For a `StakeDeposit` where we create a new stake account, this
-        // should be set to the same value as `stake_account_end`.
+        // Source stake account is the first, possibly active stake account that we'll try to unstake from.
+        // Determined by the
         pub source_stake_account {
             is_signer: false,
             // Is writable due to merge (stake_program::intruction::merge) of stake_account_end
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -56,7 +56,14 @@ pub enum LidoInstruction {
         #[allow(dead_code)] // but it's not
         amount: Lamports,
     },
-
+    /// Unstake from a validator to a new stake account
+    ///
+    /// Splits the stake account determined by the validator's inactive stake
+    /// seeds.
+    Unstake {
+        #[allow(dead_code)] // but it's not
+        amount: Lamports,
+    },
     /// Update the exchange rate, at the beginning of the epoch.
     ///
     /// This can be called by anybody.
@@ -337,6 +344,67 @@ pub fn stake_deposit(
     amount: Lamports,
 ) -> Instruction {
     let data = LidoInstruction::StakeDeposit { amount };
+    Instruction {
+        program_id: *program_id,
+        accounts: accounts.to_vec(),
+        data: data.to_vec(),
+    }
+}
+
+accounts_struct! {
+    UnstakeAccountsMeta, UnstakeAccountsInfo {
+        pub lido {
+            is_signer: false,
+            is_writable: true,
+        },
+        pub maintainer {
+            is_signer: true,
+            is_writable: false,
+        },
+        pub validator_vote_account {
+            is_signer: false,
+            is_writable: false,
+        },
+        // For a `StakeDeposit` where we temporarily create an undelegated
+        // account at `stake_account_end`, but immediately merge it into
+        // `stake_account_merge_into`, this must be set to the program-derived
+        // stake account for the validator, with seed `stake_accounts_seed_end
+        // - 1`. For a `StakeDeposit` where we create a new stake account, this
+        // should be set to the same value as `stake_account_end`.
+        pub source_stake_account {
+            is_signer: false,
+            // Is writable due to merge (stake_program::intruction::merge) of stake_account_end
+            // into stake_account_merge_into under the condition that they are not equal
+            is_writable: true,
+        },
+        // Must be set to the program-derived stake account for the given
+        // validator, with seed `stake_accounts_seed_end`.
+        pub destination_stake_account {
+            is_signer: false,
+            // Is writable due to transfer (system_instruction::transfer) from reserve_account to
+            // stake_account_end and stake program being initialized
+            // (stake_program::instruction::initialize)
+            is_writable: true,
+        },
+        pub stake_authority {
+            is_signer: false,
+            is_writable: false,
+        },
+        const sysvar_clock = sysvar::clock::id(),
+        const system_program = system_program::id(),
+        const sysvar_rent = sysvar::rent::id(),
+        const stake_program = stake_program::program::id(),
+        const stake_history = stake_history::id(),
+        const stake_program_config = stake_program::config::id(),
+    }
+}
+
+pub fn unstake(
+    program_id: &Pubkey,
+    accounts: &UnstakeAccountsMeta,
+    amount: Lamports,
+) -> Instruction {
+    let data = LidoInstruction::Unstake { amount };
     Instruction {
         program_id: *program_id,
         accounts: accounts.to_vec(),
--- a/program/src/logic.rs
+++ b/program/src/logic.rs
@@ -399,9 +399,6 @@ pub fn check_unstake_accounts(
         );
         return Err(LidoError::InvalidStakeAccount.into());
     }
-    // Does not underflow. If `stake_seeds.end == 0` the previous
-    // condition would have failed. And `stake_seeds.end` is always greater than
-    // `stake_seeds.begin`.
     let source_stake_seed = validator.entry.stake_seeds.begin;
     let destination_stake_seed = validator.entry.unstake_seeds.end;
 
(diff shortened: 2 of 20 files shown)


program/src/logic.rs
```diff
@@ -382,6 +383,130 @@ pub fn distribute_fees<'a, 'b>(
...
+
+    // If a validator doesn't have a stake account, it cannot be unstaked.
+    if validator.entry.stake_seeds.stake_accounts_seed_begin
+        == validator.entry.stake_seeds.stake_accounts_seed_end
+    {
+        msg!("Attempting to unstake from accounts in a validator that has no stake accounts.");
+        return Err(LidoError::InvalidStakeAccount.into());
+    }
+    // Does not underflow. If `stake_accounts_seed_end == 0` the previous
+    // condition would have failed.
```
ruuda, line 406: It could still be that `begin > end` and `end = 0`, but that is disallowed because we require that `end >= begin`. I think this is an invariant that is true at initialization and it is preserved by every instruction that modifies the seeds, and therefore it is true here, but it would be better to just confirm. (Although, if we are in a state where it’s false, then we are in deep trouble anyway.)

program/src/logic.rs
```diff
@@ -382,6 +383,130 @@ pub fn distribute_fees<'a, 'b>(
...
+    let source_stake_seed = validator.entry.stake_seeds.stake_accounts_seed_end - 1;
+    let destination_stake_seed = validator.entry.unstake_seeds.stake_accounts_seed_end;
+
+    let (source_stake_account, _) = Validator::find_stake_account_address(
+        program_id,
+        accounts.lido.key,
+        accounts.validator_vote_account.key,
+        source_stake_seed,
+    );
```
ruuda, line 415: Could we split this verification into calling the existing `Lido::check_stake_account` twice instead?

program/src/logic.rs
```diff
@@ -382,6 +383,130 @@ pub fn distribute_fees<'a, 'b>(
...
+    Ok(destination_bump_seed)
+}
+
+pub struct SplitStakeAccounts<'a, 'b> {
+    pub source_stake_account: &'a AccountInfo<'b>,
+    pub destination_stake_account: &'a AccountInfo<'b>,
+    pub authority: &'a AccountInfo<'b>,
+    pub system_program: &'a AccountInfo<'b>,
+    pub stake_program: &'a AccountInfo<'b>,
+}
```
ruuda, line 452: :+1: Good to make a struct for this.

program/tests/context.rs
```diff
@@ -881,6 +884,89 @@ impl Context {
             .expect("Failed to call StakeDeposit on Solido instance.")
     }
 
+    /// Unstake from the validator at `self.validator`.
+    pub async fn try_unstake(&mut self, amount: Lamports) -> transport::Result<()> {
+        // Where the new stake will live.
+        let solido = self.get_solido().await;
+        // solido.validators.entries[0].entry.get
```
ruuda, line 891: Can this be removed?

program/tests/context.rs
```diff
@@ -881,6 +884,89 @@ impl Context {
...
+            &id(),
+            &self.solido.pubkey(),
+            &validator.pubkey,
+            validator.entry.stake_seeds.stake_accounts_seed_begin,
+        );
+        let (destination_stake_account, _) = Validator::find_unstake_account_address(
+            &id(),
+            &self.solido.pubkey(),
+            &validator.pubkey,
+            validator.entry.unstake_seeds.stake_accounts_seed_begin,
```
ruuda, line 907: Shouldn’t this be the end? Initially `begin = end = 0`, but after one unstake, this will matter.

program/tests/context.rs
```diff
@@ -881,6 +884,89 @@ impl Context {
...
+        );
+
+        let stake_account_before = self.get_stake_account_from_seed(&validator.pubkey, 0).await;
+
+        send_transaction(
+            &mut self.context,
+            &mut self.nonce,
+            &[
+                // authorize_burn_instruction,
```
ruuda, line 916: Is this a leftover? As a comment, it is more confusing than clarifying :sweat_smile:

program/tests/context.rs
```diff
@@ -881,6 +884,89 @@ impl Context {
...
+        );
+        let unstake_account_balance = self.get_sol_balance(unstake_addr).await;
+        // We can't unstake all the balance if we ask for it, the split will not
+        // deactivate the rent.
+        if unstake_account_balance == amount {
+            let rent = self.get_rent().await;
+            let stake_rent = rent.minimum_balance(std::mem::size_of::<StakeState>());
+            deactivating_amount = (amount - Lamports(stake_rent)).unwrap();
+        }
+        assert_eq!(unstake_account.balance.deactivating, deactivating_amount);
```
ruuda, line 959: This feels like it should be part of a test, not the test context. I would keep the execution part in here, but move the additional balance checks into the test.

program/tests/context.rs
```diff
@@ -1219,6 +1305,43 @@ impl Context {
...
+            seed,
+        );
+
+        let clock = self.get_clock().await;
+        let stake_history = self.get_stake_history().await;
+        let stake_balance = self.get_sol_balance(stake_address).await;
+        let stake = self.get_stake_state(stake_address).await;
+        StakeAccount::from_delegated_account(stake_balance, &stake, &clock, &stake_history, seed)
+    }
+    pub async fn get_unstake_account_from_seed(
```
ruuda, line 1327: Nit: please leave a blank line between methods.

program/tests/tests/limits.rs
```diff
@@ -67,7 +67,7 @@ async fn test_max_validators_maintainers() {
 
     // The maximum number of validators that we can support, before Deposit or
     // StakeDeposit fails.
```
ruuda, line 69: Let’s also test `Unstake` here now then, to make sure that that’s not the limiting factor.

program/tests/tests/unstake.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    context.context.warp_to_slot(start_slot).unwrap();
+    context.update_exchange_rate().await;
+
+    let unstake_lamports = Lamports(1_000_000_000);
+    context.unstake(unstake_lamports).await;
+}
+
+#[tokio::test]
+async fn test_unstake_with_funded_destination_stake() {
```
ruuda, line 33: :+1: Great to test this.

program/tests/tests/unstake.rs
```diff
@@ -0,0 +1,58 @@
...
+    let (unstake_address, _) = Validator::find_unstake_account_address(
+        &crate::context::id(),
+        &context.solido.pubkey(),
+        &validator.pubkey,
+        0,
+    );
+    context.fund(unstake_address, Lamports(500_000_000)).await;
+    let unstake_lamports = Lamports(1_000_000_000);
+    context.unstake(unstake_lamports).await;
+}
```
ruuda, line 58: So two other cases I encountered while reviewing, that we should test:

* Unstaking the full amount in the stake account.
* Trying to unstake more than what’s in the stake account.

Also, we should test that the seed verification works, by trying to unstake to the wrong seed, and confirming that the transaction then fails.

ruuda (commented):
I did a first review pass. The overall changes look good, but this is a pretty important change that has impact on other parts of the code as well, even when they are not touched by this PR, so I don’t fully trust myself on this one, I will do another full pass later. But for now, here are my comments so far :)

program/src/state.rs
```diff
@@ -601,7 +607,24 @@ pub struct Validator {
     #[serde(serialize_with = "serialize_b58")]
     pub fee_address: Pubkey,
 
-    /// Start (inclusive) of the seed range for currently active stake accounts.
+    /// Seeds for active stake accounts.
+    pub stake_seeds: SeedRange,
```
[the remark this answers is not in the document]
enriquefynn: Borsh can't de/serialize it :/

program/src/state.rs
```diff
@@ -639,46 +659,66 @@ impl Validator {
...
+            stake_seeds: SeedRange::default(),
+            unstake_seeds: SeedRange::default(),
             stake_accounts_balance: Lamports(0),
             active: true,
         }
     }
 }
 
-impl PubkeyAndEntry<Validator> {
-    pub fn find_stake_account_address(
```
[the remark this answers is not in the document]
enriquefynn: There were two functions that did the same thing, I'll try to move everything to the validator's implementation.

program/src/instruction.rs
```diff
@@ -323,6 +328,66 @@ pub fn stake_deposit(
...
+            is_signer: true,
+            is_writable: false,
+        },
+        pub validator_vote_account {
+            is_signer: false,
+            is_writable: false,
+        },
+        // Source stake account is the last active stake account that we'll try
+        // to unstake from.  Determined by the program-derived stake account for
+        // the given validator, with seed `stake_seeds.stake_accounts_seed_end - 1`.
```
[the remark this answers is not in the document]
enriquefynn: Yeah, we have to unstake from the beginning. The conditions for the unstake are checked already on the `check_unstake_accounts`.

program/src/logic.rs
enriquefynn: If we do it like that, we'll have to pass the authority also for calculating stake/unstake addresses. I find it more complicated like that.

program/src/processor.rs
```diff
@@ -351,6 +361,79 @@ pub fn process_stake_deposit(
...
+        accounts.lido.key,
+        &lido,
+        &SplitStakeAccounts {
+            source_stake_account: accounts.source_stake_account,
+            destination_stake_account: accounts.destination_stake_account,
+            authority: accounts.stake_authority,
+            system_program: accounts.system_program,
+            stake_program: accounts.stake_program,
+        },
+        amount,
```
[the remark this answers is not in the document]
enriquefynn: Did a test that leaves the stake at the `MINIMUM_STAKE_ACCOUNT_BALANCE`, less and also more than the stake account has.

program/src/logic.rs
ruuda: Okay.

program/src/logic.rs
```diff
@@ -382,6 +383,130 @@ pub fn distribute_fees<'a, 'b>(
     Ok(())
 }
 
+/// Checks if the stake accounts on `accounts` correspond to the ones generated
+/// by the validator's seeds. Returns the destination bump seed.
+pub fn check_unstake_accounts(
```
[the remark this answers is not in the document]
enriquefynn: We need different things for each of them... Would make it more complicated to do it like that imo.

program/src/stake_account.rs
```diff
@@ -202,13 +202,13 @@ impl StakeAccount {
...
+        // Deactivating will be counted in the active lamports
+        let inactive_lamports = account_lamports
+            .0
             .checked_sub(active_lamports)
             .expect("Active stake cannot be larger than stake account balance.")
             .checked_sub(activating_lamports)
-            .expect("Activating stake cannot be larger than stake account balance - active.")
-            .checked_sub(deactivating_lamports)
-            .expect("Deactivating stake cannot be larger than stake account balance - active - activating.");
+            .expect("Activating stake cannot be larger than stake account balance - active.");
```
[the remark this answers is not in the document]
enriquefynn: Did some tests in #378

ruuda:
Conclusion from off-channel discussion:

* We will track both the total of the stake + unstake accounts, and the total of the unstake accounts. This way the exchange rate computations can remain the same. The reason we need to know the unstake amount, is because `Withdraw` only allows withdrawing from the heaviest validator.
* In `Withdraw`, when we check that you withdraw from the heaviest validator, we need to look at `total - unstake` to determine who is the heaviest validator, not at `total`.
* `WithdrawInactiveStake` should inspect unstake accounts too, and close them (withdraw to the reserve and bump the `begin` seed).
* For simplicity, we will **not** support merging of unstake accounts: unstake is a last resort for rebalancing. The maintainer bot should wait until close to the end of the epoch, and then issue a single unstake, if needed. Then the stake will immediately be inactive next epoch (in theory it can happen that it only partially deactivates, but then it just takes a bit longer to rebalance, no big deal), so it can do a new unstake next epoch. We can limit the amount of stake accounts by enforcing that `end - begin <= max_unstake_accounts` in the instruction.
* For `StakeDeposit`, to limit what rogue maintainers can do, and also to prevent wrong balances when two maintainers issue the same `StakeDeposit`, we should check that it stakes with the lightest validator. @ruuda will make a PR for this.

program/src/processor.rs
```diff
@@ -351,6 +355,82 @@ pub fn process_stake_deposit(
...
+            &accounts.lido.key.to_bytes(),
+            STAKE_AUTHORITY,
+            &[lido.stake_authority_bump_seed],
+        ]],
+    )?;
+
+    let validator = lido
+        .validators
+        .get_mut(accounts.validator_vote_account.key)?;
+    validator.entry.stake_accounts_balance = (validator.entry.stake_accounts_balance - amount)?;
```
ruuda, line 420: As discussed, we will track this separately.

program/tests/tests/unstake.rs
```diff
@@ -0,0 +1,123 @@
...
+use solana_program_test::tokio;
+use solana_sdk::signer::Signer;
+const STAKE_AMOUNT: Lamports = Lamports(10_000_000_000);
+
+struct UnstakeContext {
+    context: Context,
+}
+
+impl UnstakeContext {
+    async fn new(stake_amount: Lamports) -> UnstakeContext {
```
ruuda, line 20: `UnstakeContext` has a single field and only a constructor, we can simplify this to just a function

```rust
async fn new_unstake_context(stake_amount: Lamports) -> Context
```

program/src/processor.rs
```diff
@@ -351,6 +355,82 @@ pub fn process_stake_deposit(
...
+        &SplitStakeAccounts {
+            source_stake_account: accounts.source_stake_account,
+            destination_stake_account: accounts.destination_stake_account,
+            authority: accounts.stake_authority,
+            system_program: accounts.system_program,
+            stake_program: accounts.stake_program,
+        },
+        amount,
+        &[&seeds],
+    )?;
```
ruuda, line 392: Ok so now we test that you need to leave at last `MINIMUM_STAKE_ACCOUNT_BALANCE`, but with that we can’t do validator removal.

We can add a special case for unstaking exactly the current balance. That should be allowed, and also bump the `begin` seed, as it closes the source stake account (we should confirm that in a test too).

Alternatively we can remove the check that we leave the `MINIMUM_STAKE_ACCOUNT_BALANCE` in the account, but then we can end up with stake accounts that have less than the minimum, which would be very confusing, so I think special-casing unstaking 100% is fine. Also because in that case we need to bump the `begin` seed and in other cases we don’t.

ruuda (commented):
Starting to look good!

program/src/processor.rs
```diff
@@ -310,14 +313,15 @@ pub fn process_stake_deposit(
             accounts.lido.key,
             validator,
             // Does not underflow, because end > begin >= 0.
-            validator.entry.stake_accounts_seed_end - 1,
+            validator.entry.stake_seeds.end - 1,
             accounts.stake_account_merge_into,
+            VALIDATOR_STAKE_ACCOUNT,
```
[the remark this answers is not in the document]
enriquefynn: But we don't have two functions. Like that we can use a single function.

program/src/processor.rs
enriquefynn: I think we should allow unstaking > `MINIMUM_STAKE_ACCOUNT_BALANCE` if the validator is being deactivated. That should be fine

ruuda: > I think we should allow unstaking > MINIMUM_STAKE_ACCOUNT_BALANCE if the validator is being deactivated. That should be fine

Ah, we only allow it for inactive validators? That makes sense. We could go a step further and require that if you unstake from an inactive validator, you unstake the full account.

enriquefynn: Implemented it like that.

program/src/processor.rs
```diff
@@ -351,6 +355,104 @@ pub fn process_stake_deposit(
...
+        ]],
+    )?;
+
+    let validator = lido
+        .validators
+        .get_mut(accounts.validator_vote_account.key)?;
+    // Increase the `unstake_accounts_balance` by `amount`.
+    validator.entry.unstake_accounts_balance = (validator.entry.unstake_accounts_balance + amount)?;
+    msg!(
+        "VALIDATOR ACTIVE: {}, STAKE ACCOUNT:{}",
```
ruuda, line 423: Was this a temporary debugging print, or do you want to keep it? If you want to keep it, I would add a little more detailed descriptions of the values.

program/src/processor.rs
```diff
@@ -351,6 +355,104 @@ pub fn process_stake_deposit(
...
+            validator.entry.stake_accounts_balance,
+            MINIMUM_STAKE_ACCOUNT_BALANCE
+        );
+            return Err(LidoError::InvalidAmount.into());
+        }
+    } else {
+        let full_amount = (validator.entry.stake_accounts_balance
+            - validator.entry.unstake_accounts_balance)
+            .expect("Unstake accounts balance is not greater than Stake accounts balance.");
+        if full_amount != amount {
```
ruuda, line 441: This check implies that `full_amount` is in a single stake account. I think that is fair, it simplifies things. If there is another stake account with activating stake, then we need to wait for it to become active to unstake that part too, so we might as well wait and unstake everything at once, as soon as there is only a single stake account left. I would add a comment about this, because even if you try to withdraw `full_amount`, it may fail the transaction if the `full_amount` is spread over multiple stake accounts.

program/src/processor.rs
```diff
@@ -351,6 +355,104 @@ pub fn process_stake_deposit(
...
+            msg!(
+                "An inactive validator must have all its stake withdrawn. Tried\\
+                to withdraw {}, Should withdraw {} instead.",
+                amount,
+                full_amount,
+            );
+            return Err(LidoError::InvalidAmount.into());
+        }
```
ruuda, line 449: In this case, we also need to bump `stake_accounts.begin`. Otherwise the range `begin..end` will contain non-existing accounts. And maybe we can check then that `stake_accounts.begin == stake_accounts.end`, i.e. we removed the last stake account that the validator had.

program/src/processor.rs
```diff
@@ -463,17 +565,13 @@ pub fn process_withdraw_inactive_stake(
...
 
     // Visit the stake accounts one by one, and check how much SOL is in there.
     for seed in begin..end {
-        let (stake_account_address, _bump_seed) = Validator::find_stake_account_address(
-            program_id,
-            accounts.lido.key,
-            &validator.pubkey,
-            seed,
-        );
+        let (stake_account_address, _bump_seed) =
```
ruuda, line 573: We can write the loop above as `for seed in &validator.entry.stake_seeds` now.

program/src/processor.rs
```diff
@@ -463,17 +565,13 @@ pub fn process_withdraw_inactive_stake(
     let mut observed_total = Lamports(0);
     let mut excess_removed = Lamports(0);
     let mut stake_accounts = accounts.stake_accounts.iter();
-    let begin = validator.entry.stake_accounts_seed_begin;
-    let end = validator.entry.stake_accounts_seed_end;
+    let begin = validator.entry.stake_seeds.begin;
+    let end = validator.entry.stake_seeds.end;
 
     // Visit the stake accounts one by one, and check how much SOL is in there.
     for seed in begin..end {
```
ruuda, line 572: We now need to walk over all unstake accounts as well here. Maybe we can extract the loop body into a function so we can reuse it for stake accounts and unstake accounts. One thing that we need to change is that for unstake accounts, it can happen that they are 100% inactive, and if we withdraw 100%, then we need to bump the seed for that account. (It doesn’t need to be a special case for unstake, it will just not happen for regular stake accounts).

program/src/stake_account.rs
```diff
@@ -198,10 +198,12 @@ impl StakeAccount {
         // to true. See also https://github.com/ChorusOne/solido/issues/184#issuecomment-861653316.
         let fix_stake_deactivate = true;
 
-        let (active_lamports, activating_lamports, deactivating_lamports) = stake
+        let (mut active_lamports, activating_lamports, deactivating_lamports) = stake
             .delegation
             .stake_activating_and_deactivating(target_epoch, history, fix_stake_deactivate);
 
+        // Deactivating lamports is counted on the active lamports.
```
ruuda, line 205: To clarify, I would phrase this as:

> `stake_activating_and_deactivating` counts deactivating stake both as part of the active lamports, and as part of the deactivating lamports, but we want to split the lamports into mutually exclusive categories, so for us, active should not include deactivating lamports. There cannot be more lamports deactivating than there are active lamports, so this does not underflow.

program/src/state.rs
```diff
@@ -601,7 +608,28 @@ pub struct Validator {
     #[serde(serialize_with = "serialize_b58")]
     pub fee_address: Pubkey,
 
-    /// Start (inclusive) of the seed range for currently active stake accounts.
+    /// Seeds for active stake accounts.
+    pub stake_seeds: SeedRange,
+    /// Seeds for inactive stake accounts.
+    pub unstake_seeds: SeedRange,
+
+    /// Sum of the balances of the stake accounts.
```
ruuda, line 616: s/of the stake accounts/of the stake accounts and unstake accounts/, to clarify that it counts both.

program/src/state.rs
```diff
@@ -614,21 +642,30 @@ pub struct Validator {
...
 
-    /// End (exclusive) of the seed range for currently active stake accounts.
-    pub stake_accounts_seed_end: u64,
+    /// End (exclusive) of the seed range for stake accounts.
+    pub end: u64,
+}
 
-    /// Sum of the balances of the stake accounts.
-    pub stake_accounts_balance: Lamports,
+impl IntoIterator for &SeedRange {
```
ruuda, line 659: Nice :sunglasses:

program/tests/tests/unstake.rs
```diff
@@ -0,0 +1,97 @@
...
+    let rent = context.get_rent().await;
+    let stake_rent = rent.minimum_balance(std::mem::size_of::<StakeState>());
+    let result = context.try_unstake(Lamports(stake_rent - 1)).await;
+    assert!(result.is_err());
+
+    // If we unstake so that the validator still has the minimum, it should work.
+    context
+        .unstake((STAKE_AMOUNT - MINIMUM_STAKE_ACCOUNT_BALANCE).unwrap())
+        .await;
+}
```
ruuda, line 81: We should also add a test that in the case of an inactive validator, we can unstake the full amount.

ruuda (commented):
Found a few more things, but I think after this it should be good to go!

program/src/processor.rs
enriquefynn: Removed, that was a forgotten temporary debugging print.

PR #374, opened 2021-08-19
Log what we did from deposit and withdraw

ruuda:
Brian mentioned this morning that he did a deposit on devnet, and it worked, but transaction on explorer.solana.com is difficult to
understand. Part of this is just inherent, we need to reference all of those accounts. The way to make this more comprehensible is for block explorers to add special Solido parsers. But there is one thing we can do on our side: log what we did.

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -168,6 +168,15 @@ pub fn process_deposit(
         st_sol_amount,
     )?;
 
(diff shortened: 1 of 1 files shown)


enriquefynn (approved):
LGTM!
Just `cargo fmt` is complaining about a long line.

Issue #365, opened 2021-08-16
Make `RemoveValidator` world-callable, integrate it with the maintainer bot


ruuda:
Parent issue: #347.

After #351, validators now have an active flag. It is true initially, and when it is false, new deposits to this validator are blocked, and its stake balance target becomes zero. When a validator has `active` set to false, and it has no stake accounts, and no unclaimed fee credits, then the validator can safely be removed. This does not need to be gated either, anybody should be able to call this.

* Update `RemoveValidator` to only allow removal of validators where `inactive` is true, and where there are no stake accounts. (The check for fee credit is already there.)
* Update the maintainer bot to call this whenever possible. It could also trigger claiming the fees, if that is the only thing blocking removal.
* Add tests. Even though we cannot yet test unstake, we can test removing a validator that never had any stake in the first place.

ruuda:
This has been implemented in the pull request referenced above, and #399 will call it from the bot.

PR #377, opened 2021-08-20
365 make removevalidator world callable

glottologist:
I ran out of time to finish #365 off entirely but this PR is for the small headway I made into the ticket.


ruuda:
I changed the base branch to `unstake` so the changes from #373 don’t show up in the diff here. GitHub will change it to `next` automatically when #373 is merged into `next`.

program/src/process_management.rs
```diff
@@ -84,9 +84,11 @@ pub fn process_remove_validator(
...
-    let removed_validator = lido
-        .validators
-        .remove(accounts.validator_vote_account_to_remove.key)?;
+    let removed_validator = lido.validators.remove_on(
+        accounts.validator_vote_account_to_remove.key,
+        |x| x.can_be_removed(),
+        LidoError::ValidatorIsStillActive,
+    )?;
 
     if removed_validator.fee_credit != StLamports(0) {
```
ruuda, line 93: Maybe for defensive programming, we can assert that `validator.stake_accounts_balance == Lamports(0)` before we remove the validator. If it has no stake accounts, then its stake accounts balance *should* be 0, but if for whatever reason that is not the case, it is probably better to fail the transaction and keep the validator so we can investigate, than to allow it to succeed and possibly lose track of some SOL.

program/tests/tests/add_remove_validator.rs
```diff
@@ -43,11 +48,57 @@ async fn test_remove_validator_without_unclaimed_credits() {
...
 
-    context
-        .try_remove_validator(vote_account)
-        .await
-        .expect("Failed to remove validator.");
+    let result = context.try_remove_validator(vote_account).await;
 
-    let solido = context.get_solido().await;
-    assert_eq!(solido.validators.len(), 0);
+    assert_solido_error!(result, LidoError::ValidatorIsStillActive);
```
ruuda, line 53: The test name is misleading now, removing a an _inactive_ validator without unclaimed credits should succeed ... if it fails on being active, then the fact that it has no unclaimed credits isn’t really relevant any more. Also, there is no test now that tests successful removal of the validator. I think we can extend this test to

* After asserting that it has no unclaimed fees, deactivate the validator.
* Then remove it, and assert that that succeeds.

And then we can add a new test to confirm that trying to remove an active validator fails.

program/tests/tests/add_remove_validator.rs
```diff
@@ -43,11 +48,57 @@ async fn test_remove_validator_without_unclaimed_credits() {
...
+    );
+
+    // This was also the first deposit, so that should have created one stake account.
+    assert_eq!(validator_after.stake_seeds.stake_accounts_seed_begin, 0);
+    assert_eq!(validator_after.stake_seeds.stake_accounts_seed_end, 1);
+
+    let result = context.try_remove_validator(validator.vote_account).await;
+
+    // The validator should not be able to be removed if it is still active
+    //  (i.e. the active flag is set toe true OR it has stake accounts)
```
ruuda, line 102: s/toe/to/

program/tests/tests/add_remove_validator.rs
```diff
@@ -43,11 +48,57 @@ async fn test_remove_validator_without_unclaimed_credits() {
...
+    assert_solido_error!(result, LidoError::ValidatorIsStillActive);
```
ruuda, line 103: If removal fails on the fact that the validator is still active, then all of this business above of giving it stake was not relevant. We should also deactivate the validator (which does not deactivate the stake), and then we can test that it’s really the stake that prevents removing. The function to deactivate is here: https://github.com/chorusone/solido/blob/295a263525fb20c0e17c01bae994c76dd74c4d12/program/tests/context.rs#L707

ruuda (commented):
I have a few comments about how to finish this, but I suppose you will not get to it @glottologist?

enriquefynn:
I just rebased on top of `unstake`, @glottologist can work on the issues @ruuda raised.

program/src/state.rs
```diff
@@ -673,6 +673,32 @@ impl Default for Validator {
...
+    pub fn has_no_stake_accounts(&self) -> bool {
+        self.stake_seeds.begin == self.stake_seeds.end
+    }
+
+    pub fn check_can_be_removed(&self) -> Result<(), LidoError> {
+        if self.active {
+            msg!("Refusing to remove validator because it is still active, deactivate it first.");
+            return Err(LidoError::ValidatorIsStillActive);
+        }
+        if !self.has_no_stake_accounts() {
```
ruuda, line 686: We can simplify the double negation by having a method `has_stake_accounts` that checks `begin != end`.

program/tests/tests/add_remove_validator.rs
```diff
@@ -34,24 +38,30 @@ async fn test_successful_add_validator() {
...
+async fn test_removing_validator_with_stake_accounts_should_fail() {
+    let (mut context, _) = Context::new_with_two_stake_accounts().await;
+    let validator = &context.get_solido().await.validators.entries[0];
+    let result = context.try_remove_validator(validator.pubkey).await;
+
+    // The validator should not be able to be removed if it is still active
+    //  (i.e. the active flag is set toe true OR it has stake accounts)
+    assert_solido_error!(result, LidoError::ValidatorIsStillActive);
+}
+
```
ruuda, line 64: We should also add a third test where the validator has no stake accounts, but it does have unclaimed credits, and test that we refuse to remove it in that case too.

ruuda (commented):
One test is still missing, and I have one optional suggestion. LGTM aside from those!

program/tests/tests/add_remove_validator.rs
```diff
@@ -51,6 +51,51 @@ async fn test_successful_remove_validator() {
...
+    let solido = context.get_solido().await;
+    let vote_account = solido.validators.entries[0].pubkey;
+    assert_eq!(
+        solido.validators.entries[0].entry.fee_credit,
+        StLamports(62_301_530_769)
+    );
+
+    context.deactivate_validator(vote_account).await;
+    // let solido = context.get_solido().await;
+    // let vote_account = solido.validators.entries[0].pubkey;
```
ruuda, line 94: Can this be removed?

program/src/process_management.rs
```diff
@@ -84,9 +84,11 @@ pub fn process_remove_validator(
     let mut lido = deserialize_lido(program_id, accounts.lido)?;
     lido.check_manager(accounts.manager)?;
```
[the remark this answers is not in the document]
ruuda: > This needs a change in instruction.rs to remove the manager from the input accounts as well.

We forgot this part ... Discovered by Neodyme, will fix.

PR #378, opened 2021-08-23
Test `StakeAccount` and bugfix.

enriquefynn:
Previously, `StakeAccount` was wrong, it displayed the number of active Lamports even when some Lamports were deactivating. This adds some tests to safeguard changes in Solana.

### DIFF
--- a/program/src/stake_account.rs
+++ b/program/src/stake_account.rs
@@ -23,7 +23,7 @@ use solana_program::{instruction::Instruction, stake_history::StakeHistory};
 /// The sum of the four fields is equal to the SOL balance of the stake account.
 /// Note that a stake account can have a portion in `inactive` and a portion in
 /// `active`, with zero being activating or deactivating.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub struct StakeBalance {
     pub inactive: Lamports,
     pub activating: Lamports,
@@ -31,7 +31,7 @@ pub struct StakeBalance {
     pub deactivating: Lamports,
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 
 pub struct StakeAccount {
     pub balance: StakeBalance,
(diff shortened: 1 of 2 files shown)


program/src/stake_account.rs
```diff
@@ -198,17 +198,19 @@ impl StakeAccount {
...
-        let (active_lamports, activating_lamports, deactivating_lamports) = stake
+        let (mut active_lamports, activating_lamports, deactivating_lamports) = stake
             .delegation
             .stake_activating_and_deactivating(target_epoch, history, fix_stake_deactivate);
 
-        // Deactivating will be counted in the active lamports
-        let inactive_lamports = account_lamports
-            .0
+        // Deactivating lamports is counted on the active lamports.
+        active_lamports -= deactivating_lamports;
```
ruuda, line 206: Aaah, now I understand what you meant ... so Solana’s `stake_activating_deactivating` counts the same lamports twice, once as part of `active_lamports`, and once as part of `deactivating_lamports`?

That’s unexpected to me, but then again, I don’t know what I was expecting, the function is not documented at all :weary: 

Great idea to add a test to confirm the behavior!

program/src/stake_account.rs
```diff
@@ -23,15 +23,15 @@ use solana_program::{instruction::Instruction, stake_history::StakeHistory};
 /// The sum of the four fields is equal to the SOL balance of the stake account.
 /// Note that a stake account can have a portion in `inactive` and a portion in
 /// `active`, with zero being activating or deactivating.
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug, PartialEq)]
```
ruuda, line 26: We can also implement `Eq` then.

program/tests/tests/solana_assumptions.rs
```diff
@@ -157,3 +161,115 @@ async fn test_deactivating_stake_earns_rewards() {
...
+            activating: Lamports(0),
+            active: Lamports(0),
+            deactivating: (amount - stake_rent).unwrap(),
+        }
+    );
+
+    let warp_to_slot = context.get_clock().await.slot + slots_per_epoch;
+    context.context.warp_to_slot(warp_to_slot).unwrap();
+
+    // Stake is now deactivating.
```
ruuda, line 254: s/deactivating/inactive/

program/tests/tests/solana_assumptions.rs
```diff
@@ -157,3 +161,115 @@ async fn test_deactivating_stake_earns_rewards() {
...
+    assert_eq!(
+        deactivated_stake_account.balance,
+        StakeBalance {
+            inactive: amount,
+            activating: Lamports(0),
+            active: Lamports(0),
+            deactivating: Lamports(0),
+        }
+    );
+}
```
ruuda, line 275: :+1: Thanks for adding a test!

ruuda (approved):
LGTM, thanks for confirming how this works and fixing it!

I found one typo.

Issue #219, opened 2021-06-21
Make current validator balance part of StakeDeposit


ruuda:
When two instances of the maintenance bot both submit the same `StakeDeposit` transaction, they may both get executed, and the validator will get more SOL staked with it than intended. This is not a very big problem, in the long run it evens out because the bot tries to stake with the furthest-below-target validator first. But it can achieve perfect balance if maintenance bots don’t interfere.

The bot already uses a random polling delay to avoid multiple bots running in sync, but they could still submit the same transaction at roughly the same time, especially because submitting a transaction is fairly slow, and we may need `num_validators` transactions. While one bot is updating, another one could come along and start "helping".

We can sidestep this issue by making the current validator balance an argument to `StakeDeposit`. If it no longer matches the current value when the instruction executes, the on-chain program will bail out and fail the transaction. This ensures that the effect of the transaction is what the bot who submitted it intended.

ruuda:
This has been superseded by #381.

PR #381, opened 2021-08-24
Guard stake deposit

ruuda:
Add an additional check in `StakeDeposit` to confirm that no other validator with less stake exists.

* This reduces the power that maintainers have to mess with the stake distribution, which is good because it reduces the amount of trust in them. It is still possible for rogue maintainers to disturb the balance, but not by as much.
* When two maintainers race and produce the same `StakeDeposit` instruction, previously this could lead to one validator having too much stake, but now it will cause one of the two transactions to fail.

This is an alternative to #219, which will be obsolete once this is merged.

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -197,9 +197,31 @@ pub fn process_stake_deposit(
     lido.check_stake_authority(program_id, accounts.lido.key, accounts.stake_authority)?;
     lido.check_can_stake_amount(accounts.reserve, accounts.sysvar_rent, amount)?;
 
+    // Confirm that there is no other active validator with a lower balance that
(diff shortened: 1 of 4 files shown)


program/src/processor.rs
```diff
@@ -435,9 +467,7 @@ pub fn process_unstake(
             return Err(LidoError::InvalidAmount.into());
         }
     } else {
-        let full_amount = (validator.entry.stake_accounts_balance
-            - validator.entry.unstake_accounts_balance)
-            .expect("Unstake accounts balance is not greater than Stake accounts balance.");
+        let full_amount = validator.entry.effective_stake_balance();
```
enriquefynn, line 470: 👍  That's much better, thanks!

enriquefynn (approved):
LGTM

PR #382, opened 2021-08-25
Write changelog for v0.5.0

ruuda:
The Neodyme audit is starting today, so I thought we should tag what we have at this point, but before we do that, let’s include an up to date changelog.


enriquefynn (approved):
LGTM

Issue #386, opened 2021-08-26
Retry transaction to unstake from inactive validator


enriquefynn:
If for some reason Solana doesn't let us unstake the full amount a validator has, the transaction should be repeated with a lower value, maybe exponentially diminish the unstake amount until it passes.

ruuda:
Why wouldn’t Solana let us unstake the full amount? It might take more than one epoch to deactivate, but the transaction shouldn’t fail, should it?

enriquefynn:
Yeah, I think it shouldn't. I understood the mechanism wrong 😅

PR #387, opened 2021-08-27
Unstake inactive client

enriquefynn:
Tries to unstake from inactive validator.
If the validator's `inactive` flag is `true`, the maintainer bot will try to unstake all it's stake.
If, for some reason, Solana doesn't let the maintainer unstake the transaction will fail. We can implement this at a later point (#386)

This PR also prepares the remove validator client cmd. We can discuss here if the maintainer should call this instruction if the validator fits all the requirements for removal, or the function can be called manually.

### DIFF
--- a/cli/src/daemon.rs
+++ b/cli/src/daemon.rs
@@ -54,7 +54,7 @@ struct MaintenanceMetrics {
     // so we know how much SOL it costs to operate.
     // spent_lamports_total: u64
     /// Number of times we performed `UnstakeFromInactiveValidator`.
-    unstake_from_inactive: u64,
+    unstake_from_inactive_validator: u64,
 }
 
 impl MaintenanceMetrics {
@@ -95,6 +95,8 @@ impl MaintenanceMetrics {
                         .with_label("operation", "MergeStake".to_string()),
                     Metric::new(self.transactions_claim_validator_fee)
                         .with_label("operation", "ClaimValidatorFee".to_string()),
+                    Metric::new(self.unstake_from_inactive_validator)
+                        .with_label("operation", "UnstakeFromInactiveValidator".to_string()),
                 ],
             },
         )?;
@@ -137,7 +139,7 @@ fn run_main_loop(
         transactions_collect_validator_fee: 0,
         transactions_merge_stake: 0,
         transactions_claim_validator_fee: 0,
-        unstake_from_inactive: 0,
+        unstake_from_inactive_validator: 0,
     };
     let mut rng = rand::thread_rng();
 
@@ -176,7 +178,7 @@ fn run_main_loop(
                             metrics.transactions_claim_validator_fee += 1
                         }
                         MaintenanceOutput::UnstakeFromInactiveValidator { .. } => {
-                            metrics.unstake_from_inactive += 1
+                            metrics.unstake_from_inactive_validator += 1
                         }
                     }
                 }
--- a/cli/src/daemon.rs
+++ b/cli/src/daemon.rs
@@ -53,7 +53,7 @@ struct MaintenanceMetrics {
     // TODO(#96#issuecomment-859388866): Track how much the daemon spends on transaction fees,
     // so we know how much SOL it costs to operate.
     // spent_lamports_total: u64
-    /// Number of times we performed `UnstakeFromInactive`.
+    /// Number of times we performed `UnstakeFromInactiveValidator`.
     unstake_from_inactive: u64,
 }
 
@@ -175,7 +175,7 @@ fn run_main_loop(
                         MaintenanceOutput::ClaimValidatorFee { .. } => {
                             metrics.transactions_claim_validator_fee += 1
                         }
-                        MaintenanceOutput::UnstakeFromInactive { .. } => {
+                        MaintenanceOutput::UnstakeFromInactiveValidator { .. } => {
                             metrics.unstake_from_inactive += 1
                         }
                     }
--- a/cli/src/daemon.rs
+++ b/cli/src/daemon.rs
@@ -53,6 +53,8 @@ struct MaintenanceMetrics {
     // TODO(#96#issuecomment-859388866): Track how much the daemon spends on transaction fees,
     // so we know how much SOL it costs to operate.
     // spent_lamports_total: u64
+    /// Number of times we performed `UnstakeFromInactive`.
+    unstake_from_inactive: u64,
 }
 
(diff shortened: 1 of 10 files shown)


ruuda:
> We can discuss here if the maintainer should call this instruction if the validator fits all the requirements for removal, or the function can be called manually.

I would just do it automatically. There’s no going back either way after deactivating a validator, and anybody can finish the removal, so we might as well make the bot do it.

cli/src/maintenance.rs
```diff
@@ -403,6 +433,53 @@ impl SolidoState {
         Some((instruction, task))
     }
 
+    /// If there is a validator being deactivated, try to unstake its funds.
+    // TODO(#386): Solana could fail the transaction to unstake, in that case,
+    // it should be retried.
+    pub fn try_unstake_from_inactive(&self) -> Option<(Instruction, MaintenanceOutput)> {
+        for validator in &self.solido.validators.entries {
+            // Validator is inactive, try to unstake from it.
+            if !validator.entry.active {
```
ruuda, line 442: This is a bit a matter of preference, but we can write

```rust
// We are only interested in unstaking from inactive validators.
if validator.entry.active {
    continue
}
```
and that avoids some rightward drift because the remainder of the function doesn’t have to be inside the `if`.

cli/src/maintenance.rs
```diff
@@ -403,6 +433,53 @@ impl SolidoState {
...
+
+                return Some((
+                    lido::instruction::unstake(
+                        &self.solido_program_id,
+                        &lido::instruction::UnstakeAccountsMeta {
+                            lido: self.solido_address,
+                            maintainer: self.maintainer_address,
+                            validator_vote_account: validator.pubkey,
+                            source_stake_account: validator_stake_account,
+                            destination_stake_account: validator_unstake_account,
```
ruuda, line 471: Oh, should we name this field `destination_unstake_account` in `UnstakeAccountsMeta`?

cli/src/maintenance.rs
```diff
@@ -793,6 +870,7 @@ pub fn try_perform_maintenance(
         // ensure that the balance update needs to reference as few accounts
         // as possible.
         .or_else(|| state.try_merge_on_all_stakes())
+        .or_else(|| state.try_unstake_from_inactive())
```
ruuda, line 873: I would put this after updating the exchange rate; we should update the exchange rate as soon as we can, but removing the validator is not high-prio; it can’t receive new stake anyway.

tests/test_solido.py
```diff
@@ -575,3 +575,37 @@ def deposit(lamports: int, expect_created_token_account: bool = False) -> None:
...
+    'show-solido',
+    '--solido-program-id',
+    solido_program_id,
+    '--solido-address',
+    solido_address,
+)
+# Should have bumped the validator's `stake_seeds` and `unstake_seeds`.
+val = solido_instance['solido']['validators']['entries'][0]['entry']
+assert val['stake_seeds'] == {'begin': 1, 'end': 1}
+assert val['unstake_seeds'] == {'begin': 0, 'end': 1}
```
ruuda, line 611: Thanks for adding a test too!

If we call `RemoveValidator` automatically, then we can test that here as well.

cli/src/maintenance.rs
```diff
@@ -88,6 +88,18 @@ pub enum MaintenanceOutput {
         from_stake_seed: u64,
         to_stake_seed: u64,
     },
+
+    UnstakeFromInactive {
```
ruuda, line 92: On the one hand there is only one thing that could be inactive here, but still, I would write `UnstakeFromInactiveValidator` to make it more descriptive.

ruuda (commented):
Looks nice, I have just a few superficial comments.

About the subcommand, I think we should remove it, and just make the bot call `RemoveValidator`.

cli/src/daemon.rs
```diff
@@ -53,6 +53,8 @@ struct MaintenanceMetrics {
     // TODO(#96#issuecomment-859388866): Track how much the daemon spends on transaction fees,
     // so we know how much SOL it costs to operate.
     // spent_lamports_total: u64
+    /// Number of times we performed `UnstakeFromInactive`.
+    unstake_from_inactive: u64,
```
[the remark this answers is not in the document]
ruuda: Oh also, we need to actually expose this in the metrics here: https://github.com/chorusone/solido/blob/86726cda5249b5dcd5d6ccb04adcf79c15f7fa74/cli/src/daemon.rs#L83-L96

cli/src/maintenance.rs
enriquefynn: Yeah, I think it's better, thanks

cli/src/daemon.rs
```diff
@@ -53,6 +53,8 @@ struct MaintenanceMetrics {
     // TODO(#96#issuecomment-859388866): Track how much the daemon spends on transaction fees,
     // so we know how much SOL it costs to operate.
     // spent_lamports_total: u64
+    /// Number of times we performed `UnstakeFromInactiveValidator`.
+    unstake_from_inactive_validator: u64,
```
ruuda, line 57: This still doesn’t have the `transactions_` prefix unlike the other countres.

cli/src/maintenance.rs
```diff
@@ -403,6 +437,52 @@ impl SolidoState {
...
     }
 
+    /// If there is a validator being deactivated, try to unstake its funds.
+    pub fn try_unstake_from_inactive_validator(&self) -> Option<(Instruction, MaintenanceOutput)> {
+        for validator in &self.solido.validators.entries {
+            // We are only interested in unstaking from inactive validators.
+            if validator.entry.active {
+                continue;
+            }
+            let try_unstake_balance = validator.entry.effective_stake_balance();
```
ruuda, line 447: We should unstake the amount in the first stake account, not the full balance; it might be different if there are multiple stake accounts.

Also, we need to check if the account is fully inactive, otherwise the transaction will fail. Fortunately we should have the balances already.

ruuda (commented):
I found one issue still.

cli/src/maintenance.rs
enriquefynn: Well spotted, I changed it in #409.
I think we should unstake all the stake account's balance.

cli/src/daemon.rs
enriquefynn: :+1 Done in #409

PR #388, opened 2021-08-30
Do not require manager to sign validator removal

ruuda:
This was reported by Neodyme.

`RemoveValidator` still required a signature from the manager. We forgot to remove this account when we made `RemoveValidator` world-callable in https://github.com/ChorusOne/solido/pull/377. (I [mentioned](https://github.com/ChorusOne/solido/pull/377#discussion_r693012245) it in my first review but then I missed it after edits.)

Also remove program accounts that are not needed, I think these are a leftover from when we still used the SPL stake pool.

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -651,16 +651,10 @@ accounts_struct! {
             is_signer: false,
             is_writable: true,
         },
-        pub manager {
-            is_signer: true,
-            is_writable: false,
-        },
         pub validator_vote_account_to_remove {
             is_signer: false,
             is_writable: false,
         },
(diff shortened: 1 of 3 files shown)


ruuda:
I would like to add a regression test, but due to the way our test context works, it would be a pointless test; we already test validator removal, and we used to make the manager sign, but now we don’t any more ... it’s hard to test for “the manager should not need to sign” because making the manager sign is some additional step that you have to do. I think it’s fine like this with the existing test.

ruuda:
Also, I wonder if we should re-organize our code and put the accounts struct definition near the function that handles it. I think that would make it harder to miss.

enriquefynn (approved):

PR #391, opened 2021-08-30
Fix withdraw checks

ruuda:
This fixes an issue discovered by Neodyme, where `process_withdraw` would not consider the unstake account balance when checking that we withdraw from the validator that has the most stake.

Also add a test to catch this. To make the test easier to write, I modified the test context a bit.

### DIFF
--- a/program/tests/context.rs
+++ b/program/tests/context.rs
@@ -779,16 +779,18 @@ impl Context {
             .expect("Failed to call Deposit on Solido instance.")
     }
 
-    /// Withdraw from the validator at `self.validator`.
+    /// Withdraw from the given validator and stake account.
     pub async fn try_withdraw(
         &mut self,
         user: &Keypair,
         st_sol_account: Pubkey,
         amount: StLamports,
+        validator_vote_account: Pubkey,
         source_stake_account: Pubkey,
     ) -> transport::Result<Pubkey> {
         // Where the new stake will live.
         let new_stake = self.deterministic_keypair.new_keypair();
+
         send_transaction(
             &mut self.context,
             &mut self.nonce,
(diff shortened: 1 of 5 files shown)


program/tests/tests/withdrawals.rs
```diff
@@ -223,27 +230,117 @@ async fn test_withdrawal_result() {
...
-async fn test_withdrawal_from_different_validator() {
-    let mut context = Context::new_with_maintainer_and_validator().await;
-    let amount = Lamports(100_000_000_000);
-    let validator = context.validator.take().unwrap();
-    let other_validator = context.add_validator().await;
-
-    let (user, token_addr) = context.deposit(amount).await;
-    let stake_account = context
-        .stake_deposit(validator.vote_account, StakeDeposit::Append, amount)
+async fn test_withdraw_fails_if_validator_with_more_stake_exists() {
```
enriquefynn, line 233: Thanks for the test 😄

program/tests/context.rs
```diff
@@ -898,14 +907,15 @@ impl Context {
             .expect("Failed to call StakeDeposit on Solido instance.")
     }
 
-    /// Unstake from the validator at `self.validator`.
-    pub async fn try_unstake(&mut self, amount: Lamports) -> transport::Result<()> {
+    /// Try to unstake from the validator.
+    pub async fn try_unstake(
+        &mut self,
+        validator_vote_account: Pubkey,
```
enriquefynn, line 913: Better like that 👍.

program/tests/context.rs
```diff
@@ -940,9 +950,9 @@ impl Context {
         Ok(())
     }
 
-    /// Withdraw from the validator at `self.validator`.
-    pub async fn unstake(&mut self, amount: Lamports) {
-        self.try_unstake(amount)
+    /// Unstake from the validator.
```
enriquefynn, line 953: 👍

enriquefynn (approved):

PR #393, opened 2021-08-30
Withdraw from unstake account

enriquefynn:
Modifies the `WithdrawInactiveStake` to also withdraw stake from `unstake` accounts.


program/src/processor.rs
```diff
@@ -519,52 +521,82 @@ pub fn process_update_exchange_rate(
...
+        }
+    }
+}
+
+pub struct WithdrawExcessOpts<'a, 'b> {
+    accounts: &'a WithdrawInactiveStakeInfo<'a, 'b>,
+    clock: &'a Clock,
+    rent: &'a Rent,
+    stake_history: &'a StakeHistory,
+    stake_account: &'a AccountInfo<'b>,
```
ruuda, line 543: What is the difference between this account, and the other `accounts`?

program/src/processor.rs
```diff
@@ -519,52 +521,82 @@ pub fn process_update_exchange_rate(
...
+            );
+            Lamports(
+                stake_info
+                    .balance
+                    .inactive
+                    .0
+                    .saturating_sub(stake_account_rent.0),
+            )
+        }
+        StakeType::Unstake => Lamports(stake_info.balance.inactive.0),
```
ruuda, line 579: If we go with the approach of always fully withdrawing a stake account, or withdrawing nothing at all, should we reflect that here, and only make the excess positive, when 100% of the balance is inactive?

program/src/processor.rs
```diff
@@ -519,52 +521,82 @@ pub fn process_update_exchange_rate(
...
         stake_account_seed,
     );
-    let excess_balance = Lamports(
-        stake_info
-            .balance
-            .inactive
-            .0
-            .saturating_sub(stake_account_rent.0),
-    );
+    let excess_balance = match stake_type {
```
ruuda, line 564: This `StakeType` argument and then the match inside make this function pretty subtle (kudos for using an enum and not a bool by the way). I wonder if it would be simpler if we extract the part below inside the `if excess_balance > Lamports(0) {` into a function, but then have two functions that call it, one `withdraw_excess_inactive_sol` like the original, and one `withdraw_full_stake_account_if_possible`. I suspect it is simpler, because `withdraw_full_stake_account_if_possible` doesn’t need to do much besides checking that `.inactive == account.lamports()`.

program/src/processor.rs
```diff
@@ -579,8 +611,42 @@ pub fn withdraw_excess_inactive_sol<'a, 'b>(
     Ok(excess_balance)
 }
 
-/// Recover any inactive balance from the validator's stake accounts to the
-/// reserve account.
+pub fn compare_stake_account<'a, 'b>(
```
ruuda, line 614: I am confused by this function signature; how can an `AccountInfo` be the outcome of a comparison? I would expect a `compare` function to return [`Ordering`](https://doc.rust-lang.org/std/cmp/enum.Ordering.html).

program/src/processor.rs
```diff
@@ -579,8 +611,42 @@ pub fn withdraw_excess_inactive_sol<'a, 'b>(
...
+    account_iter: &'a mut Iter<AccountInfo<'b>>,
+    n_unstake_accounts: usize,
+    n_stake_accounts: usize,
+    seed: u64,
```
ruuda, line 618: I would call this `stake_account_seed` to distinguish it from other kinds of seeds.

program/src/processor.rs
```diff
@@ -579,8 +611,42 @@ pub fn withdraw_excess_inactive_sol<'a, 'b>(
...
+    stake_account_address: &Pubkey,
+    accounts: &'a WithdrawInactiveStakeInfo<'a, 'b>,
+    stake_type: StakeType,
+) -> Result<&'a AccountInfo<'b>, LidoError> {
+    match account_iter.next() {
+        None => {
+            msg!(
+                "Not enough {} accounts provided, got {} but expected {}.",
+                stake_type,
+                accounts.stake_accounts.len() - n_unstake_accounts as usize,
```
ruuda, line 628: This can overflow, for example, if you expect 1 stake account and 1 unstake account, but provide zero accounts. Then here it will do 0 - 1. I’m not sure if that causes a crash or a wraparound when compiled to Solana BPF.

We might move this check into `process_withdraw_inactive_stake`, something like
```rust
if accounts.stake_accounts.len() != n_stake_accounts + n_unstake_accounts {
    msg!("Wrong number of stake accounts provided, expected {} stake accounts and {} unstake accounts, \
        but got {} accounts.", n_stake_accounts, n_unstake_accounts, accounts.stake_accounts.len(),
    );
    return Err(LidoError::InvalidStakeAccount);
}

// Does not panic, because len = n_stake_accounts + n_unstake_accounts >= n_stake_accounts.
let (stake_accounts, unstake_accounts) = accounts.stake_accounts.split_at(n_stake_accounts);
```

Then we can zip the seeds with the accounts and do
```rust
for (seed, stake_accounts) in validator.entry.stake_seeds.zip(&stake_accounts) {
    let (stake_account_address, _bump_seed) = validator.find_stake_account_address(program_id, accounts.lido.key, seed);
    if account.key != stake_account_address {
        msg!("Wrong stake account for seed {}: expected {} but got {}.", seed, stake_account_address, account.key);
        return Err(LidoError::InvalidStakeAccount);
     }
}
// After this it is safe to use `account`.
```

It’s about the same size as the call to `compare_stake_account` is now. It’s a bit more duplication, but I think it actually makes things clearer because we don’t have to handle the two slightly different cases in one function.

program/src/processor.rs
```diff
@@ -640,44 +695,108 @@ pub fn process_withdraw_inactive_stake(
...
     msg!("{} in donations observed.", donation);
 
     // Store the new total. If we withdrew any inactive stake back to the
     // reserve, that is now no longer part of the stake accounts, so subtract
     // that.
-    validator.entry.stake_accounts_balance =
-        (observed_total - excess_removed).expect("Does not underflow, because excess <= total.");
+    validator.entry.stake_accounts_balance = (validator.entry.stake_accounts_balance
+        - excess_removed)
+        .expect("Does not underflow, because excess <= total.");
```
ruuda, line 739: This needs to move to the end of the function; we don’t know the new total yet before looking at the unstake accounts.

program/src/processor.rs
```diff
@@ -640,44 +695,108 @@ pub fn process_withdraw_inactive_stake(
...
+        let account_balance = Lamports(**unstake_account.lamports.borrow());
+        msg!(
+            "Unstake account at seed {} ({}) contains {}.",
+            seed,
+            unstake_account.key,
+            account_balance
+        );
+
+        // If validator's stake is at the beginning, try to withdraw.
+        if validator.entry.unstake_seeds.begin == seed {
```
ruuda, line 766: We should also check that the stake is 100% inactive, otherwise the withdraw will fail.

program/src/processor.rs
```diff
@@ -640,44 +695,108 @@ pub fn process_withdraw_inactive_stake(
...
+                    rent: &rent,
+                    stake_history: &stake_history,
+                    stake_account: unstake_account,
+                },
+                seed,
+                lido.stake_authority_bump_seed,
+                StakeType::Unstake,
+            )?;
+            // If withdrew everything, bump the seed
+            if account_balance == unstake_removed_here {
```
ruuda, line 780: And then, when we always withdraw 100% or nothing, this can become an assertion instead of a check.

program/src/state.rs
```diff
@@ -705,6 +705,9 @@ impl Validator {
     pub fn has_stake_accounts(&self) -> bool {
         self.stake_seeds.begin != self.stake_seeds.end
     }
+    pub fn has_unstake_accounts(&self) -> bool {
+        self.unstake_seeds.begin != self.unstake_seeds.end
+    }
```
ruuda, line 710: We need to update `check_can_be_removed` below to also refuse removal if the validator has unstake accounts!

ruuda (commented):
I reviewed `processor.rs` and `state.rs`, the maintainer changes we can get right later (maybe you can split those off into a separate PR?).

I think the approach is correct, but I have a few suggestions that I hope will make things simpler.

program/src/processor.rs
enriquefynn: I think it suffices to return the `balance.total()` from the stake, will try to withdraw everything.

program/src/error.rs
```diff
@@ -163,9 +163,13 @@ pub enum LidoError {
...
     ValidatorShouldHaveNoStakeAccounts = 41,
 
+    /// Tried to remove a validator when it when it was active or had stake accounts.
+    #[error("ValidatorShouldHaveNoUnStakeAccounts")]
+    ValidatorShouldHaveNoUnStakeAccounts = 42,
+
     /// There is a validator that has less stake than the selected one, stake to that one instead.
     #[error("ValidatorWithLessStakeExists")]
-    ValidatorWithLessStakeExists = 42,
+    ValidatorWithLessStakeExists = 43,
```
ruuda, line 172: Hey, we shouldn’t renumber these error codes once we’ve defined them :sweat_smile: 

For now nothing bad will happen because this is not yet deployed, but we should make a habit of keeping them stable.

program/src/state.rs
```diff
@@ -721,6 +724,10 @@ impl Validator {
             msg!("Refusing to remove validator because it still has stake accounts, unstake them first.");
             return Err(LidoError::ValidatorShouldHaveNoStakeAccounts);
         }
+        if self.has_unstake_accounts() {
+            msg!("Refusing to remove validator because it still has unstake accounts, withdraw them first.");
+            return Err(LidoError::ValidatorShouldHaveNoUnStakeAccounts);
```
ruuda, line 729: Nit: We call the instruction `Unstake`, but here it’s written `UnStake` with capital S.

program/src/processor.rs
ruuda: Yes, that’s the new total balance, and the new unstake balance is the `observed_unstake_total - unstake_removed`.

program/src/processor.rs
ruuda: Ooh never mind, I was confused. The accounts are just the original accounts for the call.

program/src/processor.rs
```diff
@@ -579,8 +609,35 @@ pub fn withdraw_excess_inactive_sol<'a, 'b>(
     Ok(excess_balance)
 }
 
-/// Recover any inactive balance from the validator's stake accounts to the
-/// reserve account.
+pub fn get_stake_account_balance(
+    stake_account_address: &Pubkey,
```
ruuda, line 613: Maybe we should make this a method on `Validator` instead of taking the address; right now you can just pass `accountinfo.key, accountinfo` as the arguments. (Even if it is not called like that right now, there is a risk that somebody will in the future.) If we instead take only the seed, and derive the expected address, then it’s much harder to use incorrectly.

Or maybe just call it `derived_stake_account_address` to at least clarify that we shouldn’t pass in just any address.

program/src/processor.rs
```diff
@@ -579,8 +609,35 @@ pub fn withdraw_excess_inactive_sol<'a, 'b>(
     Ok(excess_balance)
 }
 
-/// Recover any inactive balance from the validator's stake accounts to the
-/// reserve account.
+pub fn get_stake_account_balance(
```
ruuda, line 612: This does more than just getting the balance, it also check the address. Maybe call it `check_address_and_get_balance`?

program/src/processor.rs
```diff
@@ -601,83 +658,127 @@ pub fn process_withdraw_inactive_stake(
...
-                    account.key,
-                );
-                return Err(LidoError::InvalidStakeAccount.into());
-            }
-            Some(account) => account,
-        };
-        let account_balance = Lamports(**stake_account.lamports.borrow());
-        msg!(
-            "Stake account at seed {} ({}) contains {}.",
+        let stake_account = &stake_accounts[i];
```
ruuda, line 680: If we zip with `stake_accounts` instead, that will allow the bounds check to be elided. Not sure if it makes things better or worse because I’m not sure if LLVM can see that those two iterators have the same length.

program/src/processor.rs
ruuda: Good idea to make this function though :+1: `withdraw_inactive_stake` is much easier to read now.

program/src/processor.rs
ruuda: But that will make the attempt to withdraw fail if it is not 100% inactive, and that will fail the transaction, which means we can’t call `withdraw_inactive_stake` on any validator that has deactivating stake. Or am I missing something?

program/src/processor.rs
```diff
@@ -601,83 +658,127 @@ pub fn process_withdraw_inactive_stake(
...
+        return Err(LidoError::ValidatorBalanceDecreased.into());
+    }
+
     // Store the new total. If we withdrew any inactive stake back to the
     // reserve, that is now no longer part of the stake accounts, so subtract
-    // that.
-    validator.entry.stake_accounts_balance =
-        (observed_total - excess_removed).expect("Does not underflow, because excess <= total.");
+    // that + the total unstake removed.
+    validator.entry.stake_accounts_balance = ((validator.entry.stake_accounts_balance
```
ruuda, line 773: The second `validator.entry.stake_accounts_balance` should be `stake_observed_total + unstake_observed_total`; at this point we haven’t updated `stake_accounts_balance` yet.

```
stake_accounts_balance = stake_observed_total - excess_removed + unstake_observed_total - unstake_removed
```

program/src/processor.rs
```diff
@@ -601,83 +658,127 @@ pub fn process_withdraw_inactive_stake(
...
+        - excess_removed)
+        .expect("Does not underflow, because excess <= total.")
+        - unstake_removed)
```
ruuda, line 776: It’s a superficial thing, but with a chain of arithmetic and interspersed `expect`s like this, I find it nicer to use the method versions, because they `.`-chain nicely:

```rust
validator.entry.stake_accounts_balance = stake_observed_total
    .add(unstake_observed_total)
    .expect("If Solido has enough SOL to make this overflow, something has gone very wrong.")
    .sub(excess_removed)
    .expect(...)
    .sub(unstake_removed)
    .expect(...);
```

program/src/error.rs
enriquefynn: 😬 Reverted it.

program/src/processor.rs
enriquefynn: Let's called it `derived_stake_account_address`, we can change it later

program/src/processor.rs
enriquefynn: I like it better with zip indeed :)

program/src/processor.rs
enriquefynn: humm... correct, we can reuse the computation from before.

program/src/processor.rs
enriquefynn: Yes, I got confused. I changed it now.

program/src/processor.rs
```diff
@@ -519,68 +521,115 @@ pub fn process_update_exchange_rate(
...
+    accounts: &'a WithdrawInactiveStakeInfo<'a, 'b>,
+    clock: &'a Clock,
+    stake_history: &'a StakeHistory,
+    stake_account: &'a AccountInfo<'b>,
+}
+
+/// Withdraw `amount` from `withdraw_excess_opts.stake_account`.
+pub fn withdraw_inactive_sol(
+    withdraw_excess_opts: &WithdrawExcessOpts,
     stake_account_seed: u64,
```
ruuda, line 549: It looks like in all places where we use `WithdrawExcessOpts`, we also have a `stake_account_seed` argument. We could put the `stake_account_seed` in the `WithdrawExcessOpts`.

program/src/processor.rs
```diff
@@ -601,83 +650,138 @@ pub fn process_withdraw_inactive_stake(
...
+
+        let withdraw_opts = WithdrawExcessOpts {
+            accounts: &accounts,
+            clock: &clock,
+            stake_history: &stake_history,
+            stake_account: unstake_account,
+        };
+        let stake_account = get_stake_account(&withdraw_opts, seed)?;
+
+        // If validator's stake is at the beginning, try to withdraw.
```
ruuda, line 747: It would be good to also explain why we don’t withdraw if it’s not at the beginning.

ruuda (commented):
I reviewed everything except for the changes in `cli/src/maintenance.rs` now, this looks good to me!

We do need more tests for this though. I’ll write some.

PR #394, opened 2021-08-31
Limit unstake accounts, fix amount computations

ruuda:
I wanted to put a limit on the number of unstake accounts, to make sure that we don't have so many that `WithdrawInactiveStake` would fail, because we don't merge the unstake accounts.

But while going over this, I noticed some funny things in the amount computations. We should enforce the minimum stake account balance per stake account, not for the validator's overall balance.

Also, when unstaking the full amount from an inactive validator, we should unstake the full stake account, not the full amount, which might be spread over multiple stake accounts. After a merge they should be in a single one and then the difference is not relevant, but still, this was a bit fishy, so fix that.

To do:

* [x] Add tests

### DIFF
--- a/program/tests/context.rs
+++ b/program/tests/context.rs
@@ -20,7 +20,6 @@ use solana_program_test::{processor, ProgramTest, ProgramTestBanksClientExt, Pro
 use solana_sdk::account::ReadableAccount;
 use solana_sdk::account::{from_account, Account};
 use solana_sdk::account_info::AccountInfo;
-use solana_sdk::clock::Epoch;
 use solana_sdk::pubkey::Pubkey;
 use solana_sdk::signature::{Keypair, Signer};
 use solana_sdk::transaction::Transaction;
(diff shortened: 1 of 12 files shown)


ruuda:
I added tests. I broke up the huge unstake test that was testing everything into multiple smaller tests, and I made some tweaks to the context setup to make that less verbose.

I also added a new method to the context to warp to an epoch, which simplifies many tests.

program/tests/context.rs
```diff
@@ -371,6 +371,19 @@ impl Context {
             .expect("Failed to send memo transaction.")
     }
 
+    /// Warp to the given epoch after the first normal slot.
+    ///
+    /// Note, the epoch number here is not the same as the epoch number of the
+    /// clock sysvar; we start counting from the first "normal" slot.
+    pub fn warp_to_normal_epoch(&mut self, epoch: u64) {
```
enriquefynn, line 378: We can't warp to an epoch in the past, right? Maybe call it `warp_to_future_epoch`?

enriquefynn: Great function 👍, thanks!

program/tests/tests/unstake.rs
```diff
@@ -57,65 +76,129 @@ async fn test_successful_unstake() {
 }
 
 #[tokio::test]
-async fn test_unstake_balance_combinations() {
```
enriquefynn, line 60: Good idea to split it.

enriquefynn (commented):
Great simplification and tests :)

enriquefynn (approved):

program/tests/context.rs
ruuda: I’m not sure if it is possible, but it makes no sense. How about `advance_to_normal_epoch`? I put `normal` in there because it starts counting at `genesis_config.first_normal_slot`, so the first “normal” epoch is not actually the first epoch.

ruuda:
I rebased and enabled the final test, please take another look @enriquefynn.

program/tests/context.rs
enriquefynn: Yeah, that's fine 👍

PR #396, opened 2021-08-31
Modifies the client to withdraw inactive stake.

enriquefynn:
Adequate the client to the new program that withdraws if there are unstake accounts.


cli/src/maintenance.rs
```diff
@@ -571,32 +614,53 @@ impl SolidoState {
...
-                let expected_difference = (current_balance
-                    - validator.entry.stake_accounts_balance)
-                    .expect("Does not overflow because current > entry.balance.");
-                // If the expected difference is less than some defined amount
-                // of Lamports, we don't bother withdrawing. We try to do this
-                // so we don't pay more for fees than the amount that we'll
-                // withdraw.
-                if expected_difference < SolidoState::MINIMUM_WITHDRAW_AMOUNT {
-                    continue;
+            let expected_difference_stake =
```
ruuda, line 628: This looks pretty complicated, and now that I’m looking at the original code, I’m not sure if it was doing the right thing ... it checked for donations but withdrawing inactive stake can still be worthwhile even if the total balance isn’t the same ...

There are two cases in which we want to call `WithdrawInactiveStake`:

* If the stake accounts balance stored doesn’t match the balance observed, then `WithdrawInactiveStake` will fix it.
* If there is inactive stake that we can withdraw, then we should.

So how about this:

* Loop over all stake accounts and unstake accounts, and compute the total balances, and the amount of withdrawable inactive stake (`inactive - rent` for stake accounts, `inactive` for unstake accounts).
* If the computed total does not match, or if the withdrawable amount is greater than the minimum withdraw amount, continue to call the intruction.

cli/src/maintenance.rs
```diff
@@ -214,6 +230,9 @@ pub struct SolidoState {
     /// the stake balance of the derived stake accounts from the begin seed until
     /// end seed.
     pub validator_stake_accounts: Vec<Vec<(Pubkey, StakeAccount)>>,
+    /// Similar to the stake accounts, holds the unstake balance of the derived
```
glottologist, line 233: Nit:  "the unstake balance" => "the unstaked balance"

cli/src/maintenance.rs
```diff
@@ -241,11 +260,22 @@ fn get_validator_stake_accounts(
...
+        StakeType::Unstake => &validator.entry.unstake_seeds,
+    };
+    for seed in seeds {
+        let (addr, _bump_seed) = match stake_type {
+            StakeType::Stake => {
+                validator.find_stake_account_address(solido_program_id, solido_address, seed)
+            }
+            StakeType::Unstake => {
+                validator.find_unstake_account_address(solido_program_id, solido_address, seed)
+            }
```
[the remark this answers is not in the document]
enriquefynn: It creates some more matches, but I think it's better to have just one function.

cli/src/maintenance.rs
```diff
@@ -440,8 +482,9 @@ impl SolidoState {
     /// If there is a validator being deactivated, try to unstake its funds.
     pub fn try_unstake_from_inactive_validator(&self) -> Option<(Instruction, MaintenanceOutput)> {
         for validator in &self.solido.validators.entries {
-            // We are only interested in unstaking from inactive validators.
-            if validator.entry.active {
+            // We are only interested in unstaking from inactive validators that
+            // have stake accounts.
+            if validator.entry.active || !validator.entry.has_stake_accounts() {
```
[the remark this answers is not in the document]
enriquefynn: If we have activating Lamports we can still unstake it, do you think it's a problem?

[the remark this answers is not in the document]
ruuda: Ah if it works with activating stake, then it’s fine. Please add a test to confirm that it works with activating stake too then.

cli/src/maintenance.rs
```diff
@@ -60,8 +63,11 @@ pub enum MaintenanceOutput {
         /// This is only an expected value, because a different transaction might
         /// execute between us observing the state and concluding that there is
         /// a difference, and our `WithdrawInactiveStake` instruction executing.
-        #[serde(rename = "expected_difference_lamports")]
-        expected_difference: Lamports,
+        #[serde(rename = "expected_difference_stake_lamports")]
+        expected_difference_stake: Lamports,
+
+        #[serde(rename = "unstaked_amount_lamports")]
+        unstaked_amount: Lamports,
```
ruuda, line 70: Huh, why does `unstaked_amount` go in the `WithdrawInactiveStake` output? `WithdrawInactiveStake` does not itself unstake anything? Is this the amount that we expect to withdraw from unstake accounts to the reserve?

Maybe call it `unstake_withdrawn_to_reserve` or something like that?

cli/src/maintenance.rs
```diff
@@ -241,11 +260,20 @@ fn get_validator_stake_accounts(
...
+        StakeType::Stake => &validator.entry.stake_seeds,
+        StakeType::Unstake => &validator.entry.unstake_seeds,
+    };
+    for seed in seeds {
+        let (addr, _bump_seed) = validator.find_stake_account_address(
+            solido_program_id,
+            solido_address,
+            seed,
+            stake_type,
+        );
```
ruuda, line 276: :+1: Nice, this function is quite simple now with the `StakeType`.

program/src/state.rs
```diff
@@ -756,27 +757,13 @@ impl PubkeyAndEntry<Validator> {
         program_id: &Pubkey,
         solido_account: &Pubkey,
         seed: u64,
+        stake_type: &StakeType,
```
ruuda, line 760: We can take it by value and derive `Copy` for `StakeType`, it should not take more than a byte, and it makes the call sites a little less cluttered because we can avoid the `&`.

program/tests/tests/unstake.rs
```diff
@@ -247,3 +253,42 @@ async fn test_unstake_allows_at_most_three_unstake_accounts() {
...
+    let stake_account_after = context.get_stake_account_from_seed(&validator, 0).await;
+    assert_eq!(
+        (stake_account_before.balance.total() - stake_account_after.balance.total()).unwrap(),
+        unstake_lamports
+    );
+    let unstake_account = context.get_unstake_account_from_seed(&validator, 0).await;
+
+    // Unstaking activating Sol will become inactive right away.
+    assert_eq!(unstake_account.balance.inactive, unstake_lamports);
+}
```
ruuda, line 294: :+1: Cool! Thanks for testing this as well.

cli/src/maintenance.rs
```diff
@@ -576,32 +624,49 @@ impl SolidoState {
...
-                    - validator.entry.stake_accounts_balance)
-                    .expect("Does not overflow because current > entry.balance.");
-                // If the expected difference is less than some defined amount
-                // of Lamports, we don't bother withdrawing. We try to do this
-                // so we don't pay more for fees than the amount that we'll
-                // withdraw.
-                if expected_difference < SolidoState::MINIMUM_WITHDRAW_AMOUNT {
-                    continue;
+            let expected_difference_stake =
+                if current_stake_balance > validator.entry.stake_accounts_balance {
```
ruuda, line 639: As `current_stake_balance` only sums over stake accounts, not unstake accounts, this should compare to `validator.entry.effective_stake_balance()`, not to `stake_accounts_balance`.

(I’m more and more wondering if we should’t have stored only the sum of the stake accounts in `stake_accounts_balance` because this appears to be a bit subtle ... but let’s not change that so close before the launch.)

cli/src/maintenance.rs
```diff
@@ -576,32 +624,49 @@ impl SolidoState {
...
+                || removed_unstake > Lamports(0)
+            {
                 // The balance of this validator is not up to date, try to update it.
-                let stake_account_addrs = stake_accounts.iter().map(|(addr, _)| *addr).collect();
+                let mut stake_account_addrs: Vec<Pubkey> =
+                    stake_accounts.iter().map(|(addr, _)| *addr).collect();
+                // Try to also withdraw from unstake accounts
+                let mut unstake_account_addrs: Vec<Pubkey> =
+                    unstake_accounts.iter().map(|(addr, _)| *addr).collect();
+                stake_account_addrs.append(&mut unstake_account_addrs);
```
ruuda, line 669: It’s possible to collect directly into the vec without an intermediate vec:

```rust
let mut stake_account_addrs = Vec::new();
stake_account_addrs.extend(stake_accounts.iter().map(|(addr, _)| *addr));
stake_account_addrs.extend(unstake_accounts.iter().map(|(addr, _)| *addr));
```

ruuda (commented):
Looks almost good! I have a few more comments.

cli/src/maintenance.rs
enriquefynn: Yeah, that was copy-pasta bug, fixed it

cli/src/maintenance.rs
enriquefynn: Yeah, it's a bit confusing. We can change it later

ruuda: > We can change it later

Not easily; it’s stored in the on-chain state, changing that after v1 will not be worth the trouble, and changing it now before v1 is not worth the risk in my opinion, so we are stuck with it. But it’s not the end of the world; probably if we change it we find other places that become subtle.

cli/src/maintenance.rs
enriquefynn: Much better, thanks!

ruuda (approved):
:shipit:

PR #398, opened 2021-08-31
Deactivate the stake on `Withdraw` instruction.

enriquefynn:
Deactivates the stake on withdrawal, giving the client an already deactivated stake.

### DIFF
--- a/program/src/logic.rs
+++ b/program/src/logic.rs
@@ -316,6 +316,34 @@ pub fn transfer_stake_authority(
     )
 }
 
+// Deactivates the stake.
+pub fn deactivate_stake<'a>(
+    stake_account: &AccountInfo<'a>,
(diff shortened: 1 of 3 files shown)


ruuda:
But why should we do this in the on-chain program? I think it’s better to keep that as minimal as possible. The frontend can include an unstake instruction in the transaction instead; it will have the same effect and keep the on-chain program smaller.

enriquefynn:
There was a problem with wallets not being able to deactivate the stake, maybe @hritique can comment on that.

ruuda:
I discussed with @hritique, there was some confusion about what we need to do, but I think it’s clear now, and we should be able to add a deactivate instruction to the transaction from within the frontend, so we can keep te on-chain program simpler.

PR #399, opened 2021-08-31
Remove validator on client

enriquefynn:
Not tested in the integration tests, but it worked on my local setup to withdraw from the `unstake` accounts and remove a validator.


ruuda:
> Not tested in the integration tests

This shouldn’t be too hard, we can call `deactivate-validator` and then a few rounds of `perform-maintenance`, and that should eventually remove it. Can you add a test?

cli/src/daemon.rs
```diff
@@ -55,6 +55,9 @@ struct MaintenanceMetrics {
     // spent_lamports_total: u64
     /// Number of times we performed `UnstakeFromInactiveValidator`.
     unstake_from_inactive_validator: u64,
+
+    /// Number of times we performed `RemoveValidator`.
+    remove_validator: u64,
```
ruuda, line 60: Please also prefix this counter with `transactions_` like the others.

cli/src/maintenance.rs
```diff
@@ -945,6 +979,9 @@ pub fn try_perform_maintenance(
         .or_else(|| state.try_collect_validator_fee())
         // Same for updating the validator balance.
         .or_else(|| state.try_withdraw_inactive_stake())
+        // Try to remove validator should happen before withdrawing inactive
+        // stake and collecting validator's fees.
```
ruuda, line 983: Huh, no, it should happen after ... we can’t remove the validator if it has stake or unclaimed fees.

It doesn’t matter so much because we check if it can be removed, so it will still happen one iteration later, but I would still put this one last, because it is the least-priority thing we have to do; if we can stake a deposit, we prefer to do that sooner, rather than removing the validator.

program/src/state.rs
```diff
@@ -711,27 +711,52 @@ impl Validator {
...
         }
         if self.has_unstake_accounts() {
-            msg!("Refusing to remove validator because it still has unstake accounts, withdraw them first.");
             return Err(LidoError::ValidatorShouldHaveNoUnstakeAccounts);
         }
         // If not, this is a bug.
         assert_eq!(self.stake_accounts_balance, Lamports(0));
         Ok(())
     }
+    pub fn show_removed_error_msg(error: Result<(), LidoError>) -> Result<(), LidoError> {
```
ruuda, line 729: It would be a bit simpler if we don’t return the same error again, we can just borrow it.

ruuda (commented):
I have some suggestions.

program/src/state.rs
```diff
@@ -712,27 +712,46 @@ impl Validator {
...
         }
         if self.has_unstake_accounts() {
-            msg!("Refusing to remove validator because it still has unstake accounts, withdraw them first.");
             return Err(LidoError::ValidatorShouldHaveNoUnstakeAccounts);
         }
         // If not, this is a bug.
         assert_eq!(self.stake_accounts_balance, Lamports(0));
         Ok(())
     }
+    pub fn show_removed_error_msg(error: &Result<(), LidoError>) {
```
ruuda, line 730: Nit: please leave a blank line between two methods.

ruuda (approved):
LGTM with one nit! Thanks for the test as well!

program/src/state.rs
enriquefynn: Thanks, will take care 😬

PR #404, opened 2021-09-01
Rename misleading account creation method

ruuda:
Neodyme pointed out:

> `create_account_overwrite_if_exists` comment is slightly misleading. While it DOES allow creating an account on an address with >0 lamports balance, it does NOT allow overwriting an account which "really" exists (ie has owner or data set). You only use/need it for the first case though, so the code is entirely sane. (except for the options.sign_seeds passed to the transfer instruction, which are unecessary but also don't break anything).

This fixes that.

### DIFF
--- a/program/src/logic.rs
+++ b/program/src/logic.rs
@@ -106,11 +106,11 @@ pub struct CreateAccountOptions<'a, 'b> {
 /// Create a new account and fund it from the reserve.
 ///
 /// Unlike `system_instruction::create_account`, this will not fail if the account
-/// already exists. This is important, because if account creation would fail for
+/// is already funded. This is important, because if account creation would fail for
 /// stake accounts, then someone could transfer a small amount to the next stake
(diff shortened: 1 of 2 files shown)


program/src/logic.rs
```diff
@@ -105,11 +105,11 @@ pub struct CreateAccountOptions<'a, 'b> {
 /// Create a new account and fund it from the reserve.
 ///
 /// Unlike `system_instruction::create_account`, this will not fail if the account
-/// already exists. This is important, because if account creation would fail for
+/// is already funded. This is important, because if account creation would fail for
```
glottologist, line 108: Nit: "if account creation would fail for" might be better as "if account creations fails for"

enriquefynn (approved):

program/src/logic.rs
ruuda: Thanks, applied.

PR #405, opened 2021-09-01
Simplify target balance computations, take unstake accounts into account

ruuda:
As reported by Neodyme:

> `get_target_balance` has a weird API design. Why would you reuse an old target_balance array instead of creating a new one? This is an issue if the input is not all-zero and some validators are disabled, since their will not get overwritten. But because of your sanity-check for total_lamports_distributed this will simply cause an assertion failure. Currently, the only place that uses this function uses an all-zero input

I’m not sure why I went with the mutable argument, returning a vec is nicer indeed, this pull request implements that.

Also, as reported by Neodyme:

> `get_validator_furthest_below_target` can return an inactive validator, even if the inactive validator has a target of 0, if it is the first validator and all other validators have reached their target (I think this cannot actually happen right now. But it might be acheivable after you restructure your staking logic.)

This pull request fixes that as well. If there is no active validator, `get_validator_furthest_below_target` will panic, but that’s fine, because `get_target_balance` returns an `Err` when there are no active validators, so we should never get tot a point where `get_validator_furthest_below_target` is called with no active validators.

### DIFF
--- a/cli/src/maintenance.rs
+++ b/cli/src/maintenance.rs
@@ -359,10 +359,9 @@ impl SolidoState {
 
     /// If there is a deposit that can be staked, return the instructions to do so.
     pub fn try_stake_deposit(&self) -> Option<(Instruction, MaintenanceOutput)> {
-        // We can only stake if there is an active validator.
-        if self.solido.validators.iter_active().next().is_none() {
-            return None;
-        }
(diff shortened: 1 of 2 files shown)


enriquefynn (approved):
LGTM

Issue #401, opened 2021-09-01
Confirm owners of program-owned accounts


ruuda:
As reported by Neodyme:

> Your account-validation routines do not check the owners of the accounts. This has no real security impact though, but is is something to consider. Take for example `PartialVoteState::deserialize`. You run it once in `process_add_validator`. It checks if the given `validator_vote_account` is set up correctly. But only the data! If a validator with such a faked account would be added, all checks will trivially pass. The validator can then immediately after he is added use the current owner program to delete the account, and then recreate a correct vote-account at the same address, but with different parameters. This check is thus "useless" against a malicious attacker, but does protect against accidental misconfiguration issues. The same applies for `check_is_st_sol_account` (used when setting dev/treasury/reward accounts).

Let’s consider the two accounts separately

### Vote accounts

We do confirm the vote account’s authorized withdrawer and commission when we add the vote account, but this is more of a sanity check than a security check; the multisig should already have confirmed that the validator is set up correctly. Also, Solido sees a snapshot of the vote account when we add the validator, but the vote account can change. The authorized withdrawer shouldn’t change, but the commission percentage can be changed.

For now, I don’t think there is a need to enforce this in the on-chain program, but we should monitor; I’ll open a separate issue for this. Then if a validator misbehaves, the multisig can remove this validator. If we grow to hundreds of validators, maybe we should automate more of the enforcement, but for now, I think it is fine to leave this to the multisig.

Still, it doesn’t hurt for us to verify the account owner, let’s do this.

### SPL token accounts

For token accounts that we accept directly in an instruction (e.g. the one in `Deposit`), the check that it’s an stSOL account is just a sanity check so we can print somewhat helpful things to the log; the SPL token program should refuse a transfer/mint if the account’s mint is wrong. It should also check the account owner, if it doesn’t that’s a problem in the SPL token program, not in Solido. Still, it doesn’t hurt for us to verify the account owner, let’s add this.

For accounts that we store (the treasury and developer accounts, and the validator fee accounts), I don’t think checking the owner changes anything — the account can still be closed, and if it was created at an address on the curve, a user can put something else there. For the treasury and developer, we assume they are benevolent so this is not a problem. A validator could do this, but that then only blocks payment of their fees, which then makes this a case of #400.

[...]

PR #407, opened 2021-09-01
Add comments from Jon

ruuda:
This pull request is not intended to be merged. It adds the comments from @joncinque. (Thanks for the review Jon! <3) I’m opening it as a pull request so I can add comments after an issue is resolved, so we can keep track of what still needs to be addressed.

### DIFF
--- a/program/src/logic.rs
+++ b/program/src/logic.rs
@@ -67,6 +67,11 @@ pub(crate) fn check_mint(
         msg!("Mint should have an authority.");
         return Err(LidoError::InvalidMint.into());
     }
+    // LOW
+    // You may also want to check if there's an existing freeze authority to
+    // avoid any nasty surprises
+    // Additionally, check that `mint` is owned by the expected token program,
+    // probably `spl_token::id()`
     Ok(())
 }
 
@@ -78,6 +83,13 @@ pub fn get_reserve_available_balance(
     rent: &Rent,
     reserve_account: &AccountInfo,
 ) -> Result<Lamports, LidoError> {
+    // LOW
+    // There's a strange attack that can happen here -- if someone has instantiated
+    // a Lido with a reserve that has data allocated, its rent-exempt reserve will
+    // actually be bigger than `rent.minimum_balance(0)`, which means that this
+    // calculation can be wrong.  You can do `rent.minimum_balance(reserve_account.data_len())`
+    // to be totally safe.  Either way, you don't need to worry about a rent
+    // exemption for an account with 0 size.
     let minimum_balance = Lamports(rent.minimum_balance(0));
     match Lamports(reserve_account.lamports()) - minimum_balance {
         Ok(balance) => Ok(balance),
@@ -200,6 +212,9 @@ pub fn mint_st_sol_to<'a>(
     // use those.
     let mint_to_signers = [];
 
+    // nit: The spl_token instruction creator saves you here by checking
+    // `spl_token_program.key == spl_token::id()`, but to be safe, be sure to
+    // check for yourself that `spl_token_program.key` is the expected key.
     let instruction = spl_token::instruction::mint_to(
         spl_token_program.key,
         st_sol_mint.key,
@@ -237,6 +252,14 @@ pub fn burn_st_sol<'a, 'b>(
         spl_token::state::Account::unpack_from_slice(&accounts.st_sol_account.data.borrow())?;
 
     // Check if the user is the account owner.
+    // LOW
+    // This is a bit of an anti-pattern with SPL token. Typically, to
+    // be safe, we encourage people to approve a certain amount of their SPL
+    // tokens to a delegate, and then sign with that.  That way, if you're
+    // interacting with a program, you know that the program can only take the
+    // amount delegated, and nothing more.  When you sign with the owner, the
+    // program could take all of your tokens. So instead of forcing a signature from
+    // the owner, you want to allow a delegate to sign.
     if &st_sol_account.owner != accounts.st_sol_account_owner.key {
         msg!(
             "Token is owned by {}, but provided owner is {}.",
@@ -249,6 +272,8 @@ pub fn burn_st_sol<'a, 'b>(
     // The SPL token program supports multisig-managed mints, but we do not
     // use those.
     let burn_signers = [];
+    // nit: The spl_token instruction creator saves you again here, but in general
+    // be sure to check that it's the expected token program!
     let instruction = spl_token::instruction::burn(
         accounts.spl_token.key,
         accounts.st_sol_account.key,
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -64,6 +64,18 @@ pub fn process_initialize(
     let accounts = InitializeAccountsInfo::try_from_slice(accounts_raw)?;
     let rent = &Rent::from_account_info(accounts.sysvar_rent)?;
     check_rent_exempt(rent, accounts.lido, "Solido account")?;
+    // MEDIUM
+    // There's no check on the `reserve_account` to make sure it's
+    // the address corresponding to `find_program_address([lido.key, RESERVE_ACCOUNT])`
+    // This means someone could instantiate a lido, have people deposit, and
+    // steal all the funds from the reserve immediately because they have the
+    // key to the reserve.
+    // As an alternative, you could have the reserve be an undelegated stake
+    // account. You can `system_instruction::transfer` into those no problem,
+    // and there are some transparent stake / withdraw authorities for moving
+    // funds out.
+    // As another note, are you planning on storing data on the reserve account
+    // at some point?
     check_rent_exempt(rent, accounts.reserve_account, "Reserve account")?;
 
     let is_uninitialized = accounts.lido.data.borrow()[..LIDO_CONSTANT_SIZE]
@@ -119,6 +131,11 @@ pub fn process_initialize(
         stake_authority_bump_seed: deposit_bump_seed,
         rewards_withdraw_authority_bump_seed,
         reward_distribution,
+        // LOW
+        // reward distribution is just accepted as it is. is there any risk with
+        // setting `st_sol_appreciation` from the outside before init?
+        // also, you could run a risk of overflowing, but that's more of a
+        // manager-error than anything else
         fee_recipients: FeeRecipients {
             treasury_account: *accounts.treasury_account.key,
             developer_account: *accounts.developer_account.key,
@@ -180,6 +197,8 @@ pub fn process_deposit(
         st_sol_amount
     );
 
+    // Is there a reason to not update the amount of `st_sol_supply` and `sol_balance`
+    // in the exchange rate?
     lido.metrics.deposit_amount.observe(amount)?;
     lido.save(accounts.lido)
 }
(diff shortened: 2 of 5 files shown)


program/src/processor.rs
```diff
@@ -711,6 +760,12 @@ pub fn process_collect_validator_fee(
         accounts.rewards_withdraw_authority,
     )?;
 
+    // LOW
+    // Be sure to check that `validator_vote_account` is actually part of
+    // Lido! I don't think there's anything exploitable here, but if someone
+    // created a vote account and gave withdraw authority to Lido, and called
+    // this with their vote account, it would succeed.  If there's a change to
+    // the accounting, perhaps this could get exploited later.
```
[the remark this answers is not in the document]
ruuda: https://github.com/ChorusOne/solido/pull/408

program/src/processor.rs
```diff
@@ -793,6 +848,10 @@ pub fn process_withdraw(
         return Err(LidoError::ValidatorWithMoreStakeExists.into());
     }
 
+    // Nit: this also creates the same kind of confusion as the unstake instruction,
+    // where it only uses `stake_seeds.begin`, so if there are multiple stake
+    // accounts, we have to merge before performing a withdraw, which may
+    // be confusing.
```
ruuda, line 854: We only allow withdrawing from the first stake account, this is on purpose. The maintainer bot merges as many stake accounts as it can, so in the best case, we have one that’s fully active, and one that’s activating, that we can’t withdraw from anyway.

There was a bug checking which validator to withdraw from, which we fixed in https://github.com/ChorusOne/solido/pull/391.

program/src/stake_account.rs
```diff
@@ -23,6 +23,12 @@ use solana_program::{instruction::Instruction, stake_history::StakeHistory};
 /// The sum of the four fields is equal to the SOL balance of the stake account.
 /// Note that a stake account can have a portion in `inactive` and a portion in
 /// `active`, with zero being activating or deactivating.
+///
+// Note: it is actually possible for a stake account to be in a "mixed" state due to
+// cluster-wide movement of stake. This doesn't happen too often
+// anymore, but it is technically possible for a stake to have :
+// * inactive, activating, and active balance
+// * inactive, active, and deactivating balance
```
ruuda, line 31: Yes, we are aware of this.

program/src/stake_account.rs
```diff
@@ -121,6 +127,7 @@ pub fn deserialize_stake_account(account_data: &[u8]) -> Result<Stake, ProgramEr
 
     // We read 196 bytes from the data, so check that up front, so that bounds
     // checks can be optimized away below.
+    // I thought this was 200 the whole time, thanks for calculating this!
```
ruuda, line 130: I’m not sure if there is any data after it, it’s just that we only access the first 196 bytes.

program/src/stake_account.rs
```diff
@@ -196,6 +203,8 @@ impl StakeAccount {
 
         // This toggle is a historical quirk in Solana and should always be set
         // to true. See also https://github.com/ChorusOne/solido/issues/184#issuecomment-861653316.
+        // Note: this has all finally been removed in 1.7, so you can get rid of it
+        // whenever all of the code is updated to the newest SDK.
```
ruuda, line 207: I’t still there in 1.7.10, but it’s gone in 1.7.11 ... do you do breaking changes in patch releases? :fearful: 

We are building against 1.7.8 and stuck with it for now because the multisig program uses Anchor which pins Solana to 1.7.8. But good to know, and glad to hear it’s gone! I guess will find out automatically when we upgrade.

program/src/stake_account.rs
```diff
@@ -265,6 +274,7 @@ impl StakeAccount {
             return true;
         }
         // The voter pubkey and credits observed must match. Voter must be the same by assumption.
+        // Note: You can get rid of this soon, hooray!
```
ruuda, line 277: Dancing Banana

program/src/state.rs
```diff
@@ -27,6 +27,11 @@ use crate::{
 };
 use crate::{REWARDS_WITHDRAW_AUTHORITY, VALIDATOR_STAKE_ACCOUNT, VALIDATOR_UNSTAKE_ACCOUNT};
 
+// Nit:
+// Since an account is initialized with 0'ed out data, you can't tell if an
+// account has been initialized or not if you're looking for a 0.  That's fine
+// because you look through all the bytes on init, but for simplicity, you could
+// make this `1` and only check for that on init.
```
ruuda, line 34: We’ll go with the 0 for now because other projects already started integrating.

program/src/state.rs
```diff
@@ -157,6 +162,11 @@ impl ExchangeRate {
         // The result is in Lamports, because the type system considers Rational
         // dimensionless, but in this case `rate` has dimensions stSOL/SOL, so
         // we need to re-wrap the result in the right type.
+        //
+        // LOW
+        // This is a user error, but immediately after earning any rewards, 1 lamport
+        // would result in 0 stSOL, and the program will happily take the user's
+        // lamport without giving anything in return.
```
ruuda, line 169: Let’s look at when this can return 0. It returns a product of two rationals (if we upgrade the left-hand-side to a rational), which is zero if at least one of the factors is zero:

* If `amount` is zero then you should get `0` stSOL out. This is working as intended.
* The other case is where `rate` is 0, and the only way for that to happen is when the numerator `st_sol_supply` is 0. But we checked above, and in case `st_sol_supply == StLamports(0)`, we return the input amount nominally converted to stSOL. So here the user does get a nonzero amount of stSOL (unless `amount` was 0, but then it’s again working as intended).

program/src/state.rs
```diff
@@ -259,6 +269,9 @@ impl Lido {
 
     /// Confirm that the given account is an SPL token account with our stSOL mint as mint.
     pub fn check_is_st_sol_account(&self, token_account_info: &AccountInfo) -> ProgramResult {
+        // LOW
+        // You can add an owner check on `token_account_info` to make sure it's
+        // actually owned by the token program.
```
ruuda, line 274: Yes, we should. https://github.com/ChorusOne/solido/issues/401

ruuda (commented):
@joncinque Thanks again for the review! Fortunately it looks like we already addressed all serious issues. I opened pull requests for the other things you found; the only thing that still needs to be addressed is to check the owners of accounts that should be program-owned, I opened an issue for that.

program/src/logic.rs
```diff
@@ -78,6 +83,13 @@ pub fn get_reserve_available_balance(
     rent: &Rent,
     reserve_account: &AccountInfo,
 ) -> Result<Lamports, LidoError> {
+    // LOW
+    // There's a strange attack that can happen here -- if someone has instantiated
+    // a Lido with a reserve that has data allocated, its rent-exempt reserve will
+    // actually be bigger than `rent.minimum_balance(0)`, which means that this
+    // calculation can be wrong.  You can do `rent.minimum_balance(reserve_account.data_len())`
+    // to be totally safe.  Either way, you don't need to worry about a rent
+    // exemption for an account with 0 size.
```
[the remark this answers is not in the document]
joncinque: Nope, accounts with 0 size are just plain SOL accounts.

program/src/logic.rs
```diff
@@ -237,6 +252,14 @@ pub fn burn_st_sol<'a, 'b>(
...
 
     // Check if the user is the account owner.
+    // LOW
+    // This is a bit of an anti-pattern with SPL token. Typically, to
+    // be safe, we encourage people to approve a certain amount of their SPL
+    // tokens to a delegate, and then sign with that.  That way, if you're
+    // interacting with a program, you know that the program can only take the
+    // amount delegated, and nothing more.  When you sign with the owner, the
+    // program could take all of your tokens. So instead of forcing a signature from
+    // the owner, you want to allow a delegate to sign.
```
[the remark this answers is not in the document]
joncinque: Ok, that's fine.  As an alternative, it would be easy just to have two accounts: owner and signer, where the signer is either a delegate or the owner.  Your choice!

program/src/processor.rs
```diff
@@ -119,6 +131,11 @@ pub fn process_initialize(
         stake_authority_bump_seed: deposit_bump_seed,
         rewards_withdraw_authority_bump_seed,
         reward_distribution,
+        // LOW
+        // reward distribution is just accepted as it is. is there any risk with
+        // setting `st_sol_appreciation` from the outside before init?
```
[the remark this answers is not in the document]
joncinque: You can disregard this, I didn't understand what `st_sol_appreciation` was supposed to be, and sorry, I must be so used to seeing `as u64` that I totally missed them in `sum` :sweat:

program/src/processor.rs
```diff
@@ -119,6 +131,11 @@ pub fn process_initialize(
...
+        // also, you could run a risk of overflowing, but that's more of a
```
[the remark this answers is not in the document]
joncinque: As above, you are correct, and can disregard this

program/src/processor.rs
```diff
@@ -180,6 +197,8 @@ pub fn process_deposit(
         st_sol_amount
     );
 
+    // Is there a reason to not update the amount of `st_sol_supply` and `sol_balance`
+    // in the exchange rate?
```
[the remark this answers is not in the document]
joncinque: Ah, thanks for the reading, makes sense!

program/src/processor.rs
```diff
@@ -492,6 +524,15 @@ pub fn process_unstake(
...
 
+// nit: I'm probably missing some context, but it seems like you would be better off
+// simply updating these fields as they get changed and getting rid of this and
+// privilege `WithdrawInactiveStake` to do the accounting.
+// If this is called before all of the `WithdrawInactiveStake` instructions,
+// for that whole epoch, withdrawers will not get enough SOL back, and depositors
+// will get too many stSOL tokens.
+// For example, the stake pool program has a similar instruction, but in there,
+// it's actually checking all of the stake account balances to find out how much
+// SOL it has under management.
```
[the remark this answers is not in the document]
joncinque: Interesting, that's a smart way of keeping the system accessible even as you pay out all of these rewards

program/src/stake_account.rs
joncinque: I think I went through the whole struct afterwards just to check if you were right, and I think it actually is `196`

program/src/stake_account.rs
joncinque: > do you do breaking changes in patch releases?

Sometimes... sorry. It'll be an easy fix at least!

program/src/state.rs
joncinque: Fair enough, it's not actually a problem, just something to keep in mind for next time

program/src/state.rs
joncinque: The implementation for `Mul` between a token type and `Rational` is:
```
            fn mul(self, other: Rational) -> Result<$TokenLamports> {
                // This multiplication cannot overflow, because we expand the
                // u64s into u128, and u64::MAX * u64::MAX < u128::MAX.
                let result_u128 = ((self.0 as u128) * (other.numerator as u128))
                    .checked_div(other.denominator as u128)
                    .ok_or(ArithmeticError)?;
                u64::try_from(result_u128)
                    .map($TokenLamports)
                    .map_err(|_| ArithmeticError)
            }
```
If you have `sol_balance = 101` and `st_sol_supply = 100`, then for 1 lamport you'll have `1 * 100 / 101`, which gives 0.  Is that correct?

Either way, it's not a big deal, since you'll spend way more on transaction fees by depositing 1 lamport.

ruuda: Ah, this is a loss due to rounding errors, that can happen indeed. If `sol_balance` ≈ `st_sol_supply` then it is not a problem. But if `st_sol_supply` is much larger than `sol_balance`, then it can become an issue, because in a sense it causes us to lose precision, and users will lose some funds to rounding errors. This is a known problem, but it can be prevented by ensuring that we deposit enough SOL initially at the 1:1 exchange rate.

ruuda:
Closing this now that #401 has been addressed.

PR #408, opened 2021-09-01
Refuse to collect fees from foreign vote accounts

ruuda:
This was an issue caught by @joncinque: in `CollectValidatorFee`, we never checked that the validator was part of Solido. While I don’t believe this would have any negative consequences, we should still check this, especially in case we make a change to `CollectValidatorFee` in the future and we don’t realize that we did not check this.

I also added a regression test, and I wrote the test first to confirm that it would have caught the bug.


glottologist (approved):
LTGM!

PR #409, opened 2021-09-02
Fix issue in when unstaking from inactive validator.

enriquefynn:


cli/src/maintenance.rs
```diff
@@ -439,29 +439,38 @@ impl SolidoState {
...
+            .iter()
+            .zip(self.validator_stake_accounts.iter())
+        {
+            // We are only interested in unstaking from inactive validators that
+            // have stake accounts.
+            if validator.entry.active || stake_accounts.is_empty() {
+                continue;
+            }
+            // Validator's stake account is not fully inactive.
+            if stake_accounts[0].1.balance.inactive != stake_accounts[0].1.balance.total() {
```
glottologist: Small suggestion:  Using the tuple accessors everywhere makes this a little difficult to read (reason about) in my opinion.  An improvement would be to map the stake accounts into a struct with field names.  Or better yet, change:

```Rust
    pub validator_stake_accounts: Vec<Vec<(Pubkey, StakeAccount)>>,
```
to use a struct rather than a ```(Pubkey, StakeAccount)``` but that will have a little more impact.

cli/src/maintenance.rs
```diff
@@ -439,29 +439,34 @@ impl SolidoState {
...
+            .iter()
+            .zip(self.validator_stake_accounts.iter())
+        {
+            // We are only interested in unstaking from inactive validators that
+            // have stake accounts.
+            if validator.entry.active || stake_accounts.is_empty() {
+                continue;
+            }
+            // Validator already has 3 unstake accounts.
+            if validator.entry.unstake_seeds.end - validator.entry.unstake_seeds.begin >= 3 {
```
ruuda, line 455: We should make this 3 a constant (maybe in `program/src/lib.rs`), instead of repeating it in multiple places.

cli/src/maintenance.rs
```diff
@@ -471,11 +476,11 @@ impl SolidoState {
                         lido: self.solido_address,
                         maintainer: self.maintainer_address,
                         validator_vote_account: validator.pubkey,
-                        source_stake_account: validator_stake_account,
+                        source_stake_account: stake_accounts[0].0,
                         destination_unstake_account: validator_unstake_account,
                         stake_authority: self.get_stake_authority(),
                     },
-                    try_unstake_balance,
+                    stake_accounts[0].1.balance.total(),
```
ruuda, line 483: We also still need to confirm somewhere that the stake account is fully active, right? Or does it work even when it is activating?

cli/src/maintenance.rs
ruuda: Can also give it a name temporarily:

```rust
let (stake_account_address, stake_account_balance) = &stake_accounts[0];

cli/src/maintenance.rs
enriquefynn: It works when it's activating, I added some tests in #396

ruuda (approved):
LGTM

PR #411, opened 2021-09-07
Verify owner accounts

enriquefynn:
Verifies owner accounts for the validator's vote account and the SPL token accounts that call `check_is_st_sol_account`.


program/src/error.rs
```diff
@@ -159,6 +159,12 @@ pub enum LidoError {
     /// We can't unstake more in this epoch, wait for stake to deactivate, close
     /// the unstake accounts with `WithdrawInactiveStake`, and retry next epoch.
     MaxUnstakeAccountsReached = 44,
+
+    /// The validator's vote account is not owned by the vote program.
+    ValidatorVoteAccountHasDifferentOwner = 45,
+
+    /// We expected the StSol account to be owned by the SPL program.
```
ruuda, line 166: s/SPL program/SPL token program/

ruuda (approved):
LGTM! Thanks for the test as well.

PR #413, opened 2021-09-07
Add subcommand to approve a batch of multisig transactions

ruuda:
We are going to onboard 15 validators tomorrow, which will require 15 multisig transactions, plus some to add the maintainers.

Approving those one by one is going to be tedious and error-prone, therefore:

* Add a helper command `multisig approve-batch`, that takes the path to a file that contains transaction addresses, one per line.
* For every transaction, the command:
  * Prints it, like `multisig show-transaction` would.
  * Asks the user to confirm to approve.
  * If the threshold has been reached, asks the user to confirm to execute.

I quickly tested this manually, but it needs better testing with an automated test. Before I do that though, I’m going to add a script for us to also automatically propose these multisig transactions from the validator onboarding form.

This is still very basic. For example, if anything fails, you have to start over, it can’t retry. It’s not polished at all, but I think it’s better to have than having to manually approve transactions one by one.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -580,7 +580,6 @@ cli_opt_struct! {
 
 cli_opt_struct! {
     ApproveBatchOpts {
-        // TODO: Can be omitted, we can obtain it from the transaction account.
         /// The multisig account whose owners should vote for this proposal.
         #[clap(long, value_name = "address")]
         multisig_address: Pubkey,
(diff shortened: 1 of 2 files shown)


ruuda:
Idea: we can allow trailing content after the address in the file with transaction addresses, to describe what the transaction should be doing. If we generate all multisig transactions with a script, we can have it include the name of the validator that it's adding.

enriquefynn (approved):
Let's merge it now and work on the improvements after.

enriquefynn (approved):
Let's merge it now and work on the improvements after.

ruuda:
I added a18014ba to support extra content after the transaction address, to make it compatible with the output of #414.

PR #414, opened 2021-09-07
Add script to propose multisig transactions for validators, and to verify these transactions

ruuda:
## Summary

We need to onboard 15 validators tomorrow. We have their details in a Google Form that we can export as tsv. We have a script to verify the responses (#410). Now there are two more things we need:

* A way for us to easily crease multisig transactions to add these validators.
* A way for multisig members to easily verify that they are approving the right thing.

This pull request adds two scripts that do just that.

## Details

I added a "parser" for the form response tsv in #410, extract it into a module so it can be reused.

* `propose_add_validators.py` iterates the validators and creates multisig transactions. It writes the transaction addresses to a file.
* `verify_multisig_transactions.py` takes the the file with transaction addresses, and also the tsv file of form responses, and checks that the vote account and fee account in the transaction match those in the form response.

After doing the verification, multisig members should be able to use `approve-batch` (#413) to approve the transactions one by one. I need to modify #413 to allow for the extra content after the transaction address.

CC @mkaczanowski we can use `propose_add_validators.py` tomorrow to create the multisig transactions, maybe you can have a look as well?

### DIFF
--- a/scripts/verify_validator_submissions.py
+++ b/scripts/verify_validator_submissions.py
@@ -6,24 +6,7 @@
 """
 This script verifies details provided by validators that are onboarding.
 
-Takes tab-separated values on stdin with the following columns:
-
- * Timestamp
- * Email address
- * Validator name
- * Keybase account name
- * Vote account
- * (Withdraw authority check, constant)
- * (Commission percentage check, constant)
- * (Vote promise check, constant)
- * stSOL account
- * (stSOL account mint check, constant)
- * (Added public key to keybase check)
- * Identity name as published on identity account
- * (unused column, optional)
- * Maintainer address (if applicable; optional)
-
-This script then verifies:
+Takes tab-separated form responses on stdin. This script then verifies:
 
  * The vote account withdraw authority.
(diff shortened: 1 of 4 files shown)


scripts/verify_validator_submissions.py
```diff
@@ -153,183 +134,128 @@ class VoteAccount(NamedTuple):
...
-
-        name = vote_accounts.setdefault(self.vote_account_address, self.validator_name)
-        if name != self.validator_name:
-            print_error(f'Vote account is already in use by {name}.')
-        else:
-            print_ok('Vote account address is unique among responses seen so far.')
-
-        name = identity_accounts.setdefault(
-            vote_account.validator_identity_address, self.validator_name
+def get_vote_account(self: ValidatorResponse) -> Optional[VoteAccount]:
```
ruuda, line 137: This diff looks scarier than it is by default. I recommend clicking the cog icon at the top of the review window, and turning on “hide whitespace changes”.

glottologist (approved):
LGTM!

PR #415, opened 2021-09-08
Write changelog for v1.0.0

ruuda:
I think this is the only thing left blocking the release.

It would be good to get #414 in as well, for the sole reason that we can tell multisig participants to check out `v1.0.0` and the verification scripts will be there.


glottologist (approved):
LGTM v1.0.0 !!!

PR #417, opened 2021-09-08
Test commission change.

### LINKED ISSUE #412: Watch commission of enrolled validators, or confirm that it cannot be changed
Continuation of https://github.com/ChorusOne/solido/pull/410#discussion_r703042549.

We should confirm that only the authorized withdrawer can change the commission percentage of a vote account, ideally by adding a test in `tests/solana_assumptions.rs`.

If it turns out that it is possible for somebody other than the withdraw authority to change the commission percentage, then we should add monitoring so we can detect if any of the validator commission percentages is not the expected 100%. I think the best way to expose that would be to add a metric for it, and then we can set Prometheus alerts on it. But before we go there, let’s verify whether it is necessary.

enriquefynn:
Test that only the withdraw authority can change the commission.
Closes #412.

### DIFF
--- a/program/tests/tests/solana_assumptions.rs
+++ b/program/tests/tests/solana_assumptions.rs
@@ -8,9 +8,7 @@
 //! In some places the Solana documentation is absent or incomplete, so we test
 //! the implementation to see how Solana actually works.
 
-use solana_program::{
-    pubkey::Pubkey, stake::state::StakeState, system_instruction, system_program,
-};
+use solana_program::{pubkey::Pubkey, stake::state::StakeState};
 use solana_program_test::tokio;
 use solana_sdk::signature::{Keypair, Signer};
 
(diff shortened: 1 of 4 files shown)


program/tests/tests/solana_assumptions.rs
```diff
@@ -258,3 +264,122 @@ async fn test_stake_accounts() {
...
+) -> solana_sdk::transport::Result<()> {
+    send_transaction(
+        &mut context.context,
+        &mut context.nonce,
+        &[vote_instruction::update_commission(
+            &vote_account,
+            &withdraw_authority.pubkey(),
+            new_commission,
+        )],
+        vec![&withdraw_authority],
```
ruuda, line 282: In this case we pass the withdraw authority both to the `update_commission` instruction, and we use it as the signer. But what if we pass the right withdraw authority, but sign with something else, like the private key of the vote account? Is changing the commission still rejected then?

ruuda (commented):
Thanks!

program/tests/tests/solana_assumptions.rs
enriquefynn: I hope so. Added some changes 👍

enriquefynn: The problem is that signing panics with `KeypairPubkeyMismatch`. I don't think we need to test this assumption :P

ruuda: Hmm, ok, good to know.

ruuda (approved):
LGTM!

Issue #427, opened 2021-09-13
Active rebalancing


ruuda:
## Description

Currently we only support **passive rebalancing**, i.e. when new deposits come in, we stake them in such a way that we get closer to the target distribution. For withdrawals, we enforce that they withdraw from the validator with most stake, which also brings us closer to the target distribution, although it can overshoot and increase the imbalance. To mitigate that, the amount per withdrawal is limited.

With only passive rebalancing, if we onboard a new validator, they will not get any stake if no new deposits come in. Also, after a big withdrawal that leaves the pool unbalanced, we cannot rebalance without new deposits coming in.

The solution to this is **active rebalancing**: actively unstaking from validators so we can later stake that elsewhere.

## Thoughts and challenges

(As discussed off-channel with @enriquefynn)

* To prevent simultaneous unstake, we need maintainer coordination. (#421)
* Because unstake is a last resort that is only necessary when new deposits can’t restore the balance, 
* Can also move staking to be only at the end, no need to do it after every deposit.
* Start staking and unstaking when 1h is left on the epoch.
* Fynn: We should also have a threshold for unstake, because unstake makes us miss out on rewards, so we should only do it when we really need to. Maybe when the validator is more than 10% above target?
* But that by itself may lead to us not unstaking when we add a new validator, because 9 validators are just a little above their target, but the 10th is far below. So if any validator is far below, we need to allow unstake on the others.
* Conclusion: need to unstake if one of the two conditions above holds.

## Concerns

* If we move the operations to the end of the epoch, if something goes wrong, there is little time to fix it.
* However, with epochs of 2.5 days, they are still relatively frequent, so this shouldn’t be a big issue.

enriquefynn:
I stumbled on some things while implementing this feature.
It would be better to unstake several stake accounts in a single transaction. Would be great to do them all in one go, for instance, if we have validators with effective balance [10, 30, 21] Lamports, the balanced target would be [21, 20, 20]. Meaning we need to unstake [0, 10, 1] Lamports from the stake accounts. If we unstake 10 Sol from the second validator, the next iteration would not unstake 1 Sol because doesn't pass the threshold. That's not so bad in this case, but I believe if a new validator is joining and we need to unstake from the rest to fund this new validator, there might not be enough Sol to pass the threshold for each unstake if some conditions are met.

[...]

PR #430, opened 2021-09-15
Unstake active validators

enriquefynn:
Adds a function in the maintainer to unstake Lamports from validators if there's the need of a rebalance.

There's a threshold (in %) for when to unstake from validators, the conditions for the unstake are:
- If the amount that should be unstaked is above the threshold.
- If there is some validator which needs to be staked and the amount is above the threshold.
- 
As discussed in #427 with @ruuda this PR implements and tests the unstaking logic.

### DIFF
--- a/cli/src/daemon.rs
+++ b/cli/src/daemon.rs
@@ -61,8 +61,8 @@ struct MaintenanceMetrics {
     /// Number of times we performed `RemoveValidator`.
     transactions_remove_validator: u64,
 
-    /// Number of times we performed `Unstake` on a validator for balancing purposes.
-    transactions_unstake_from_validator: u64,
+    /// Number of times we performed `Unstake` on an active validator for balancing purposes.
+    transactions_unstake_from_active_validator: u64,
 }
 
 impl MaintenanceMetrics {
@@ -136,8 +136,8 @@ impl MaintenanceMetrics {
                 self.transactions_unstake_from_inactive_validator += 1
             }
             MaintenanceOutput::RemoveValidator { .. } => self.transactions_remove_validator += 1,
-            MaintenanceOutput::UnstakeFromValidator { .. } => {
-                self.transactions_unstake_from_validator += 1
+            MaintenanceOutput::UnstakeFromActiveValidator { .. } => {
+                self.transactions_unstake_from_active_validator += 1
             }
         }
     }
@@ -291,7 +291,7 @@ impl<'a, 'b> Daemon<'a, 'b> {
             transactions_claim_validator_fee: 0,
             transactions_unstake_from_inactive_validator: 0,
             transactions_remove_validator: 0,
-            transactions_unstake_from_validator: 0,
+            transactions_unstake_from_active_validator: 0,
         };
         Daemon {
             config,
--- a/cli/src/daemon.rs
+++ b/cli/src/daemon.rs
@@ -107,6 +107,8 @@ impl MaintenanceMetrics {
                         .with_label("operation", "UnstakeFromInactiveValidator".to_string()),
                     Metric::new(self.transactions_remove_validator)
                         .with_label("operation", "RemoveValidator".to_string()),
+                    Metric::new(self.transactions_unstake_from_active_validator)
+                        .with_label("operation", "UnstakeFromActiveValidator".to_string()),
                 ],
             },
         )?;
--- a/cli/src/daemon.rs
+++ b/cli/src/daemon.rs
@@ -60,6 +60,9 @@ struct MaintenanceMetrics {
 
     /// Number of times we performed `RemoveValidator`.
     transactions_remove_validator: u64,
+
+    /// Number of times we performed `Unstake` on a validator for balancing purposes.
+    transactions_unstake_from_validator: u64,
 }
 
 impl MaintenanceMetrics {
@@ -133,6 +136,9 @@ impl MaintenanceMetrics {
                 self.transactions_unstake_from_inactive_validator += 1
             }
             MaintenanceOutput::RemoveValidator { .. } => self.transactions_remove_validator += 1,
+            MaintenanceOutput::UnstakeFromValidator { .. } => {
+                self.transactions_unstake_from_validator += 1
+            }
         }
     }
 }
@@ -285,6 +291,7 @@ impl<'a, 'b> Daemon<'a, 'b> {
             transactions_claim_validator_fee: 0,
             transactions_unstake_from_inactive_validator: 0,
             transactions_remove_validator: 0,
+            transactions_unstake_from_validator: 0,
         };
         Daemon {
             config,
--- a/cli/src/maintenance.rs
+++ b/cli/src/maintenance.rs
@@ -358,7 +358,7 @@ impl SolidoState {
     const MINIMUM_WITHDRAW_AMOUNT: Lamports = Lamports(DEFAULT_TARGET_LAMPORTS_PER_SIGNATURE * 100);
 
     // Threshold that will trigger unstake on validators.
-    // Expressed as a percentage between 0-1.
+    // Expressed as a percentage between 0-100.
     const UNBALANCE_THRESHOLD: u64 = 10;
     /// Read the state from the on-chain data.
     pub fn new(
--- a/cli/src/maintenance.rs
+++ b/cli/src/maintenance.rs
@@ -869,48 +869,33 @@ impl SolidoState {
 
     /// Unstake from active validators in order to rebalance validators.
     pub fn try_unstake_from_active_validators(&self) -> Option<(Instruction, MaintenanceOutput)> {
-        // Get the target for each validator. Undelegated stake can be balanced
-        // when staking with validators.
+        // Get the target for each validator. Undelegated Lamports can be
+        // balanced when staking with validators.
         let targets = lido::balance::get_target_balance(Lamports(0), &self.solido.validators)
             .expect("Failed to compute target balance.");
         let unstake_amounts =
             lido::balance::get_unstake_to_rebalance(&self.solido.validators, &targets);
 
-        let rebalance = self
-            .solido
-            .validators
-            .entries
-            .iter()
-            .zip(unstake_amounts.iter())
-            .any(|(validator, unstake_amount)| {
-                (unstake_amount.0 * 100) / validator.entry.effective_stake_balance().0
-                    > SolidoState::UNBALANCE_THRESHOLD
-            });
-
-        if rebalance {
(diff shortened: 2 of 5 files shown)


cli/src/maintenance.rs
```diff
@@ -798,6 +829,58 @@ impl SolidoState {
...
+
+        // Get the maximum that can be unstaked from the stake account.
+        let amount = unstake_amounts[validator_index].min(stake_account.1.balance.total());
+
+        let expected_balance = (validator.entry.effective_stake_balance() - amount)
+            .expect("amount is always <= balance");
+        let amount = if expected_balance > MINIMUM_STAKE_ACCOUNT_BALANCE {
+            Some(amount)
+        } else {
+            let min_amount = (validator.entry.effective_stake_balance()
```
ruuda, line 861: This should be the amount in the first stake account, not the total effective balance.

tests/test_solido.py
```diff
@@ -211,46 +213,66 @@ def approve_and_execute(
...
-    'tests/.keys/validator-token-account-key.json', st_sol_mint_account
-)
-print(f'> Validator stSol token account: {validator_fee_account}')
+class Validator(NamedTuple):
+    account: TestAccount
+    vote_account: TestAccount
+    fee_account: str
+
+
+def add_validator(keypath_account: str, keypath_vote: str) -> Tuple[Validator, Any]:
```
ruuda, line 235: A bit more precise than `Any` for json objects is `Dict[str, Any]`.

tests/test_solido.py
```diff
@@ -484,6 +506,45 @@ def deposit(lamports: int, expect_created_token_account: bool = False) -> None:
...
+)
+
+del result['UnstakeFromActiveValidator']['from_stake_account']
+del result['UnstakeFromActiveValidator']['to_unstake_account']
+expected_result = {
+    'UnstakeFromActiveValidator': {
+        'validator_vote_account': validator.vote_account.pubkey,
+        'from_stake_seed': 0,
+        'to_unstake_seed': 0,
+        'amount': 1500000000,
```
ruuda, line 544: You can use numeric underscores in Python to make this large number more readable, i.e. `1_500_000_000`.

tests/test_solido.py
```diff
@@ -523,16 +588,58 @@ def deposit(lamports: int, expect_created_token_account: bool = False) -> None:
...
 )
+
+del result['StakeDeposit']['stake_account']
 expected_result = {
     'StakeDeposit': {
-        'validator_vote_account': validator_vote_account.pubkey,
-        # We have 1.0 SOL from the true deposit, and 1.0 donated.
-        'amount_lamports': int(2.0e9),
+        'validator_vote_account': validator_1.vote_account.pubkey,
+        'amount_lamports': 2_050_250_000,
```
ruuda, line 596: How did we get to this number? It looks a bit arbitrary to me, can you add a comment about why we should get this particular amount? (E.g. deposited X and Y = Z and we divide that over n validators.)

ruuda (commented):
This looks mostly good to me! My main comment is that we should unstake from the validator furthest above its target, not just from the first one.

Also, we were planning to avoid any stake/unstake operations until the last hour of the epoch, will that be done in a follow-up PR? (Maybe we should also have a flag to disable this behavior, because it will make testing more difficult.)

cli/src/maintenance.rs
```diff
@@ -95,21 +95,45 @@ pub enum MaintenanceOutput {
...
+    #[serde(serialize_with = "serialize_b58")]
+    from_stake_account: Pubkey,
+    #[serde(serialize_with = "serialize_b58")]
+    to_unstake_account: Pubkey,
+    from_stake_seed: u64,
+    to_unstake_seed: u64,
+    amount: Lamports,
+}
+
+fn print_unstake(f: &mut fmt::Formatter, unstake: &Unstake) -> fmt::Result {
```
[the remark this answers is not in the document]
enriquefynn: Great, it's better. Similar to the function below it.

program/src/token.rs
```diff
@@ -22,12 +22,19 @@ use std::{
...
+#[derive(Copy, Clone, PartialEq, Debug)]
 pub struct Rational {
     pub numerator: u64,
     pub denominator: u64,
 }
 
+impl PartialOrd for Rational {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        (self.numerator as f64 / self.denominator as f64)
+            .partial_cmp(&(other.numerator as f64 / other.denominator as f64))
```
ruuda, line 34: This comparison is lossy, `f64` cannot represent every `u64` exactly. This can lead to incorrect results:

```rust
#[test]
fn test_rational_cmp() {
    // Construct x and y such that x < y.
    let x = Rational {
        numerator: 1 << 53,
        denominator: 1,
    };
    let y = Rational {
        numerator: x.numerator + 1,
        denominator: x.denominator,
    };
    assert_eq!(x.partial_cmp(&y), Some(std::cmp::Ordering::Less));
}
```
```
thread 'token::test_rational_cmp' panicked at 'assertion failed: `(left == right)`
  left: `Some(Equal)`,
 right: `Some(Less)`', program/src/token.rs:49:5
```

We can fix this by avoiding floats. Observe that
```
   a/b < x/y
=> a   < bx/y    if b > 0
=> ay  < bx      if y > 0
```
We only work with non-negative denominators so the comparison flipping is not an issue, we only have to check to avoid division by zero:

```rust
    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
        if self.denominator == 0 || other.denominator == 0 {
            None
        } else {
            let x = self.numerator as u128 * other.denominator as u128;
            let y = other.numerator as u128 * self.denominator as u128;
            Some(x.cmp(&y))
        }
    }
```

The multiplications don't overflow because we widen to `u128`.

Please also add tests for other cases (like division by zero).

cli/src/daemon.rs
```diff
@@ -133,6 +136,9 @@ impl MaintenanceMetrics {
                 self.transactions_unstake_from_inactive_validator += 1
```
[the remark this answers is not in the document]
enriquefynn: It is already there, below `RemoveValidator`.

cli/src/maintenance.rs
```diff
@@ -836,6 +870,53 @@ impl SolidoState {
...
+
+        let (validator_index, unstake_amount) = lido::balance::get_unstake_validator_index(
+            &self.solido.validators,
+            &targets,
+            SolidoState::UNBALANCE_THRESHOLD,
+        )?;
+        let validator = &self.solido.validators.entries[validator_index];
+        let stake_account = &self.validator_stake_accounts[validator_index][0];
+
+        let maximum_unstake = (validator.entry.effective_stake_balance()
```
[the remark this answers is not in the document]
enriquefynn: This is the maximum that can be unstaked from a validator, but afterwards we get the minimum from the stake account, this value and what should be unstaked.

cli/src/maintenance.rs
```diff
@@ -836,6 +870,53 @@ impl SolidoState {
...
+            .expect("Validator should always have the minimum amount.");
+        // Get the maximum that can be unstaked from the stake account.  The
+        // minimum amongst the value to be unstaked, how much is in the stake
+        // account and the maximum that can be unstaked from the validator.
+        let amount = unstake_amount
+            .min(stake_account.1.balance.total())
+            .min(maximum_unstake);
+        if amount == Lamports(0) {
+            return None;
+        }
```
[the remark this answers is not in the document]
enriquefynn: This would be always the case because of the comment above.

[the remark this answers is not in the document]
enriquefynn: Ohh... I though the `MINIMUM_STAKE_ACCOUNT_BALANCE` was per validator's balance, otherwise we should keep track of the minimum to be rent exempt.

cli/src/maintenance.rs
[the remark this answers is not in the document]
enriquefynn: You're right 👍, the balance is per stake account and not per validator's balance

program/src/token.rs
enriquefynn: Thanks for the tests and more correct implementation, I'll change it and do more tests 👍

cli/src/daemon.rs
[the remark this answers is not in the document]
ruuda: No it’s not there, I mean here: https://github.com/ChorusOne/solido/blob/64509d59f7687470ef00f1d42fb5a8d09b91bfe5/cli/src/daemon.rs#L90-L107

cli/src/maintenance.rs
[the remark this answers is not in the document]
ruuda: The reason we have the minimum was to work around the Solana bug, where it’s not possible to merge stake accounts if they contain to little SOL.

program/src/balance.rs
```diff
@@ -102,6 +102,63 @@ pub fn get_target_balance(
...
+                Rational {
+                    numerator: target_difference,
+                    denominator: target.0,
+                } >= threshold
+            });
+
+    let (idx, validator) = validators
+        .entries
+        .iter()
+        .enumerate()
```
ruuda, line 138: We could zip it with `target_balance` and avoid the indexing:

```rust
validators
    .entries
    .iter()
    .zip(&target_balance)
    .max_by_key(|(validator, target)| ...)
```

program/src/token.rs
```diff
@@ -257,4 +269,36 @@ pub mod test {
...
+    fn test_division_by_zero_cmp() {
+        let x = Rational {
+            numerator: 1,
+            denominator: 0,
+        };
+        let y = Rational {
+            numerator: x.numerator,
+            denominator: x.denominator + 1,
+        };
+        assert_eq!(x.partial_cmp(&y), None);
```
ruuda, line 297: Please also test the reverse case of `y.partial_cmp(&x)`, that’s one of those places where a subtle copy-paste bug might hide. In fact, let’s also add the reverse case in the `test_rational_cmp` test above.

ruuda (commented):
Looks almost good to me, I have a few small comments left.

program/src/balance.rs
enriquefynn: Yeah, but we still need to return the index if we want to later get the stake account associated with the validator. We could pass it on the function, but I think it's not necessary.

ruuda: Fair point, never mind then.

ruuda (approved):
LGTM :shipit: :rocket:

Issue #421, opened 2021-09-09
Reducing races between maintainer instances


ruuda:
When there is maintenance to perform, all instances of the maintainer bot will submit a transaction to do the operation, but only one of those will succeed, and the others are unlucky.

To prevent those races, we could assign every maintainer some slots in which they are the "leader", and if a maintainer is not a leader, it will not submit transactions. This can still lead to races at the boundaries, but it should greatly reduce the number of races we see.

We expect all maintainers to cooperate, so we can select the "leader" without coordination by mapping slots to maintainers. We can do something like

    leader_index = (slot_number // 100) % num_maintainers

where `//` is integer division. Increasing the constant (100 above) will lead to fewer races, but also longer delays until maintenance happens, when a maintainer is offline. Decreasing the constant will improve responsiveness at the cost of more races. I think ~100 would be a good value, with the current 650ms block times, that means the leader rotates about once per minute.

enriquefynn:
We also need to add some sort of failure detector. If another maintainer hasn't seen the leader performing a maintenance for 10 consecutive times, it should rotate the leader, maybe // 100 + 1

enriquefynn:
That should decrease the races ^^

ruuda:
Permissionless maintainers will make it worse actually, because then you no longer know who else is participating, and you can’t coordinate to avoid the races. But I guess that’s the price of permissionlessness.

ruuda:
I have a draft implementation for this in the `maintainer-duty` branch, and it will prevent maintainers from taking action when they are not “on duty”.

However, there is a bad interaction with the current polling setup: if the maintainer goes to sleep just before their duty, it might oversleep, and never perform their duty. I haven’t worked out the math for how bad this is and what the expected latency is; I think it’s better to adjust the sleeping logic so the maintainer will only sleep until it’s their turn; we can time it such that the maintainer wakes up at the right moment.

To make that a bit nicer to implement, I am refactoring the daemon loop a bit. While I’m at it I’ll tackle #420.

PR #431, opened 2021-09-16
Refactor maintainer

### LINKED ISSUE #420: Maintainer should expose metrics even if its transactions fail
The maintenance bot does this in a loop:

 1. Read the on-chain state
 2. See if there is any maintenance to do, and if so, do it
 3. Update the state snapshot that is served by the `/metrics` endpoint

If either step 1 or 2 fail, then the state snapshot used for metrics remains empty. If step 1 fails, that’s the best we can do. But if step 2 fails, then we should still be able to update the state snapshot and have working metrics.

ruuda:
This makes a few changes to how the maintenance daemon is set up:

 * Distinguish between failing to read the on-chain state (which is likely some connectivity problem), and successfully reading the on-chain state, but failing to perform maintenance (possibly because some other transaction raced us).

 * The latter error no longer blocks the on-chain state snapshot from being available in the metrics.

 * In the case of a connectivity problem, retry with exponential backoff with jitter, instead of retrying at the normal poll interval.

 * Don't randomize the poll interval in case of success. This was initially added to break ties when multiple maintainers are running, but we are going to solve that differently with [maintainer duties](https://github.com/ChorusOne/solido/issues/421), so this should no longer be necessary.

Fixes #420.

### DIFF
--- a/cli/src/daemon.rs
+++ b/cli/src/daemon.rs
@@ -13,15 +13,16 @@ use std::io;
 use std::sync::Arc;
 use std::sync::Mutex;
 use std::thread::JoinHandle;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
-use rand::Rng;
+use rand::{rngs::ThreadRng, Rng};
 use tiny_http::{Request, Response, Server};
 
 use crate::config::RunMaintainerOpts;
-use crate::error::AsPrettyError;
+use crate::error::{AsPrettyError, Error};
 use crate::maintenance::{try_perform_maintenance, MaintenanceOutput, SolidoState};
 use crate::prometheus::{write_metric, Metric, MetricFamily};
+use crate::snapshot::SnapshotError;
 use crate::SnapshotClientConfig;
 
 /// Metrics counters that track how many maintenance operations we performed.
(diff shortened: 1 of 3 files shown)


cli/src/daemon.rs
```diff
@@ -107,6 +108,32 @@ impl MaintenanceMetrics {
         )?;
         Ok(())
     }
+
+    /// Increment the counter for a maintenance operation.
+    pub fn observe_maintenance(&mut self, maintenance_output: &MaintenanceOutput) {
```
enriquefynn, line 113: Better like this, thanks! 👍

cli/src/daemon.rs
```diff
@@ -129,104 +192,126 @@ struct Snapshot {
...
+            .unwrap()
+            .replace(Arc::new(snapshot));
+    }
+
+    /// Sleep with exponential backoff and jitter.
+    fn sleep_after_error(&mut self) {
+        // For the sleep time we use exponential backoff with jitter [1]. By taking
+        // the time since the last success as the target sleep time, we get
+        // exponential backoff. We clamp this to ensure we don't wait indefinitely.
+        // 1: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
```
enriquefynn, line 258: The maintainer just got fancier now ^^, but where's the exponential backoff part of it?
Great!

ruuda: The sleep time is the time since the last successful call, see below. This makes the time between calls grow exponentially, as it roughly doubles every time: the time you sleep this time is the sum of all previous sleeps.

enriquefynn (approved):

PR #433, opened 2021-09-17
Include balances of all maintainers in metrics

ruuda:
This enables two things:

 * We no longer assume that the configured signer is a maintainer, so you can run the maintenance daemon even if your account does not exist on mainnet. It will never perform any maintenance (after we implement duties), but you can still extract metrics this way.

 * Maintainers (or anybody really) can monitor each other, not just themselves.


glottologist (approved):
Nice!

PR #435, opened 2021-09-22
Expose withdraw stats, Solana stake, and Solana epoch info in `/metrics`

ruuda:
Expose the following new Prometheus metrics.

* Withdrawal count and amount. We were already tracking these, but we forgot to expose them.
* Amount of stake on Solana. This allows us to compute the fraction of stake controlled by Lido. Unfortunately it doesn’t appear to be possible to get current amount of active stake, but we can get the amount from the previous epoch from the stake history sysvar.
* Solana epoch, epoch start slot, and slots per epoch. With the start slot, slots per epoch, and the block height that we already had, we can estimate the time left until the next epoch.

For withdrawal stats, also print them in `solido show-solido`, next to the deposit stats.

### DIFF
--- a/cli/src/helpers.rs
+++ b/cli/src/helpers.rs
@@ -446,6 +446,17 @@ impl fmt::Display for ShowSolidoOutput {
         )?;
         writeln!(
             f,
+            "  Total amount withdrawn:   {}, valued at {} when it was withdrawn",
+            self.solido.metrics.withdraw_amount.total_st_sol_amount,
+            self.solido.metrics.withdraw_amount.total_sol_amount,
+        )?;
+        writeln!(
+            f,
+            "  Number of withdrawals:    {}",
+            self.solido.metrics.withdraw_amount.count,
+        )?;
+        writeln!(
+            f,
             "  Total deposited:          {}",
             self.solido.metrics.deposit_amount.total
         )?;
--- a/cli/src/prometheus.rs
+++ b/cli/src/prometheus.rs
@@ -259,9 +259,7 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
             name: "solido_withdraw_count_total",
             help: "Total number of withdrawals done by users.",
             type_: "counter",
-            metrics: vec![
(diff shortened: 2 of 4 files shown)


cli/src/helpers.rs
```diff
@@ -444,6 +444,17 @@ impl fmt::Display for ShowSolidoOutput {
             "  Total stSOL appreciation: {}",
             self.solido.metrics.st_sol_appreciation_sol_total
         )?;
+        writeln!(
+            f,
+            "  Total amount withdrawn:   {}, valued at {} when it was withdrawn",
```
glottologist, line 449: Nit: Might be better to add units into message, i..e "  Total amount withdrawn:   {} stSOL, valued at {} SOL when it was withdrawn"

cli/src/prometheus.rs
```diff
@@ -252,6 +252,41 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
             metrics: solido_histogram_to_metrics(at, &metrics.deposit_amount),
         },
     )?;
+
+    write_metric(
+        out,
+        &MetricFamily {
+            name: "solido_withdraw_count_total",
+            help: "Total number of withdrawals done by users.",
```
glottologist, line 260: Nit: "Total number of withdrawals done by users." -> "Total number of withdrawals made by users."

glottologist (approved):
LGTM, with a couple of nits

cli/src/helpers.rs
ruuda: The units are included, they are part of the `Lamports` and `StLamports` display impl: https://github.com/ChorusOne/solido/blob/a601d3a8e3aabd4056ac7b14d8534b197706712e/program/src/token.rs#L61-L64

cli/src/prometheus.rs
ruuda: Fixed.

cli/src/helpers.rs
glottologist: Perfect!

PR #436, opened 2021-09-22
Expose validator info in `/metrics` and `solido show-solido`

ruuda:
## Summary

Identifying validators by their vote account address is a pain. Some of them grinded an address with some prefix, but trusting vanity addresses is dangerous, and not all validators have one anyway. So to be able to tell who is who, use Solana’s validator info mechanism to fetch the name and Keybase username, and include those in metrics, and display them in `solido show-solido`. Then we can use `{{validator_name}}` as a label for dashboards.

## Background: Solana validator info

The way validator info is stored on Solana is a bit convoluted. Here’s an overview of how some account types relate:

solana

We identify validators by their vote account everywhere. We can load and parse the vote account to get a `VoteState`, which contains a `node_pubkey` field. This holds the public key that is referred to as the “validator identity account” in Solana docs and CLI tools. However, the identity account does not store anything itself. So where *is* validator metadata stores?

There is a [Config Program](https://docs.rs/solana-config-program/1.7.12/solana_config_program/index.html) on Solana. The config program creates accounts that hold lists of `(Pubkey, bool)`. One of the pubkeys that may occur here is [`solana_account_decoder::validator_info::id()`](https://docs.rs/solana-account-decoder/1.7.12/solana_account_decoder/validator_info/fn.id.html). If that pubkey is present, then the config account is a "validator info account". Aside from this magic pubkey that makes it a validator info account, the list will then also contain the validator identity account pubkey at index 1.

A config account holds this list of `(Pubkey, bool)` serialized with bincode. But if the config account is a validator info account, then after the list there is one more thing: a string serialized with bincode. This string contains a json object, and the json object holds the validator metadata.

For example, the config account that holds the metadata for our Lido validator is `5TP7x4wtj1KcvpFNFhtgeAW5beF2zTwu2KVdtvvFNG3W`:
```console
$ solana account 5TP7x4wtj1KcvpFNFhtgeAW5beF2zTwu2KVdtvvFNG3W

Public Key: 5TP7x4wtj1KcvpFNFhtgeAW5beF2zTwu2KVdtvvFNG3W
Balance: 0.004796996 SOL
Owner: Config1111111111111111111111111111111111111
Executable: false
Rent Epoch: 227
Length: 643 (0x283) bytes
0000:   02 07 51 97  01 74 48 f2  ac 5d c2 3c  9e bc 7a c7   ..Q..tH..].<..z.
0010:   8c 0a 27 25  7a c6 14 45  8d e0 a4 f1  6f 80 00 00   ..'%z..E....o...
0020:   00 00 6c e2  73 42 1e 79  f6 71 fe f8  55 dd ab 63   ..l.sB.y.q..U..c
0030:   3e 96 aa 59  a0 4a 85 ec  a1 a7 e6 7b  74 61 0f 65   >..Y.J.....{ta.e
0040:   cc 7f 01 91  00 00 00 00  00 00 00 7b  22 64 65 74   ...........{"det
0050:   61 69 6c 73  22 3a 22 43  68 6f 72 75  73 20 4f 6e   ails":"Chorus On
0060:   65 20 76 61  6c 69 64 61  74 6f 72 20  66 6f 72 20   e validator for 
0070:   4c 69 64 6f  20 66 6f 72  20 53 6f 6c  61 6e 61 22   Lido for Solana"
0080:   2c 22 6b 65  79 62 61 73  65 55 73 65  72 6e 61 6d   ,"keybaseUsernam
0090:   65 22 3a 22  63 68 6f 72  75 73 6f 6e  65 69 6e 63   e":"chorusoneinc
00a0:   22 2c 22 6e  61 6d 65 22  3a 22 4c 69  64 6f 20 2f   ","name":"Lido /
00b0:   20 43 68 6f  72 75 73 20  4f 6e 65 22  2c 22 77 65    Chorus One","we
00c0:   62 73 69 74  65 22 3a 22  68 74 74 70  73 3a 2f 2f   bsite":"https://
00d0:   63 68 6f 72  75 73 2e 6f  6e 65 22 7d  00 00 00 00   chorus.one"}....
00e0:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
```

Now the big question is, **given a validator identity account address, how do we find the config account that contains its validator info**? Answer: we have to check them all. I’m not aware of any relation between the config account address and the validator that it contains the info for. What [`solana validator-info get` does](https://github.com/solana-labs/solana/blob/0b50bb2b2025b43c27bc755a95fbf114f2655d30/cli/src/validator_info.rs#L378-L387), is to retrieve _all_ accounts owned by the config program, deserialize them, and check if there is a `validator_info::id()` in there. If it is then we found a validator info config, and it knows for which validator it holds the info.

Some unanswered questions that I have regarding this:

 * What prevents multiple info accounts from existing for the same validator identity account? Does the config program prevent this? How could it, it can’t possibly scan all accounts it owns ...
 * What ensures that only the validator itself can change its config? I hope the config program requires a signature when it puts a public key in the `(Pubkey, bool)` list?
 * What is the meaning of the bool? It’s not documented anywhere. In fact, the entire purpose of this `(Pubkey, bool)` list is unclear to me apart from the two entries used for the validator info accounts.

Anyway, what we can do then is get all validator infos, and then we also have it for the validators that we are interested in.

## Complications

* We have this `snapshot` module to work around RPC limitations, so we can do atomic reads of multiple accounts without worrying too much about it. But to get all config accounts, we need the `getProgramAccounts` RPC call, and then we have torn reads again. Fortunately we don't expect these config accounts to change much, so we can:
  * Get all config accounts once at the start, and parse them, and build a map from validator identity address to config address.
  * Read all accounts in a single RPC call as usual.
  * When we need the validator info, we consult the map to find the config account, and then we read the config account (which should be part of the snapshot). Then validate that the config account is still the config account for the validator we were looking for.
* Our internal RPC node times out when I call `getProgramAccounts`, but on `api.mainnet-beta.solana.com` it works fine. I’m still investigating this.


cli/src/maintenance.rs
```diff
@@ -341,6 +346,26 @@ fn get_account_balance_except_rent(rent: &Rent, account: &Account) -> Lamports {
...
+fn sanitize_validator_name(name: &str) -> String {
+    // Lido policy is that validator names should start with "Lido / ", so that
+    // adds no information, strip it here to leave more space for graphs in
+    // dashboards, and not waste so much space on the redundant part of the name.
+    match name.strip_prefix("Lido / ") {
+        // I don't want distracting emojis in my Grafana dashboards, so remove
+        // code points in the Supplementary Multilingual Plane and beyond. This
+        // strips emojis and dingbats while leaving letters and punctuation of
+        // all contemporary languages.
+        Some(suffix) => suffix
```
glottologist, line 359: Nice!

cli/src/maintenance.rs
```diff
@@ -946,24 +970,57 @@ impl SolidoState {
...
+            // lifetime annotations that way, so we manually define the closure
+            // struct here.
+            struct MetricAnnotator {
+                produced_at: SystemTime,
+                vote_account: String,
+                name: String,
+                keybase_username: String,
+            }
+
+            impl MetricAnnotator {
```
glottologist, line 994: Very nice!

glottologist (approved):
Awesome!

ruuda:
The test failed because `solido show-solido` now expects all Lido validators to have validator info, but the ones we created in the tests did not. Fixed by setting validator info in tests.

PR #437, opened 2021-09-23
Be more explicit about using "confirmed" commitment level, wait for confirmation

ruuda:
**Note**: This description is outdated, see discussion below for the current state of this pull request.

-------

I was running the maintenance daemon to export metrics all day, and I observed the vote credits going *down*. That’s problematic when used with Promtheus's `rate`, because it treats it as a counter restart and suddenly you have double the rate. The vote credits going down should never be possible, the one thing I can think of that would cause this is, is that the RPC node that we queried was following one particular chain where the votes got credited, but then a different fork won out in the end, and so these vote credits were never awarded after all.

If we use commitment level "finalized", this should not happen.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -325,8 +325,10 @@ fn main() {
         get_signer_from_key(opts.keypair.unwrap())
     };
 
+    // We use finalized here to be sure that we never observe rollbacks. In particular,
+    // this ensures that the maintenance daemon never observes vote credits going down.
     let rpc_client =
-        RpcClient::new_with_commitment(opts.cluster.unwrap(), CommitmentConfig::confirmed());
+        RpcClient::new_with_commitment(opts.cluster.unwrap(), CommitmentConfig::finalized());
     let snapshot_client = SnapshotClient::new(rpc_client);
 
     let output_mode = opts.output_mode.unwrap();
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -325,10 +325,16 @@ fn main() {
         get_signer_from_key(opts.keypair.unwrap())
     };
 
-    // We use finalized here to be sure that we never observe rollbacks. In particular,
-    // this ensures that the maintenance daemon never observes vote credits going down.
+    // Use commitment level "confirmed" as a middle ground between getting a recent
+    // state, and not having to wait for a long time for transactions to be confirmed.
+    // This means that we can sometimes read states that do not get finalized, when
+    // there is a reorg. We considered using "finalized" commitment level to avoid
+    // this, but then we have to wait 32 slots after every transaction, and on top
+    // of that we base transactions on old states, which increases the likelihood
+    // of them failing when executed. So we go back to "confirmed" after all. See
+    // also https://github.com/ChorusOne/solido/pull/437.
     let rpc_client =
-        RpcClient::new_with_commitment(opts.cluster.unwrap(), CommitmentConfig::finalized());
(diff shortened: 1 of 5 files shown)


ruuda:
Hmm, this makes setting up the multisig in tests fail:

```
Failed to create multisig.
Solana RPC client returned an error:

  Request:    Some(SendTransaction)
  Kind:       RPC response error
  Error code: -32002
  Message:    Transaction simulation failed: Blockhash not found
  Reason:     Transaction preflight failure
  Error:     

    Raw:      BlockhashNotFound
    Display:  Blockhash not found
```

ruuda:
I fixed this by changing the preflight commitment level. My understanding is this:

* Nodes maintain different "banks clients" (views of the on-chain state) for the different commitment levels. The "confirmed" banks client will have the state at block `i` where `i` is the highest block that has been voted on by a supermajority. The "finalized" banks client will have the state at block `j` where `j + 32` is the highest block that has been voted on by a supermajority. <sub>(I am not sure if I understand this correctly ... yesterday I observed a state at "confirmed" commitment that was later rolled back, so "confirmed" is definitely not final. But if it is not, then what is the meaning of a supermajority having voted for the block? Are Solana votes like the prepare votes in traditional BFT?)</sub>
* So if we query at a stricter commitment level, we see back further in time.
* For preflight, it makes little sense to preflight a transaction against an old state, because it will not be executed against that old state, it will be executed against the latest state. In fact, I wonder if we should preflight against `CommitmentLevel::Processed` instead ... I can’t think of a reason to ever test against an older state if you have a newer one (even if it is uncertain whether this will become the new state).

After submitting the transaction, if we want to observe its effects in the next read at finalized commitment level, then we need to wait for the transaction to be finalized. One disadvantage of this PR compared to the previous approach of using "confirmed" commitment level, is that waiting for finalization takes 32 blocks, so everything becomes a lot slower. I don’t think this is a big problem because we don't need high throughput; I would rather read finalized chain states with a slight delay, than to read states that can still be reverted with less delay.

ruuda:
@enriquefynn I’m curious to hear your opinion on this as well.

ruuda:
A downside of reading at finalized commitment level is that we "read in the past", so there is more risk of transactions failing if e.g. there was a withdrawal, which changes which validator we should stake with. We base our actions on an outdated view of the state. Preflight would catch that, but it means that in situations of high activity, we rarely get to make a transaction. I don’t think that is a big deal because there isn't so much activity.

I am convinced now that preflight should happen at the least strict commitment level.

ruuda:
Hmm, waiting for transactions to be finalized makes things unbearably slow, I am now leaning towards reverting it back to `confirmed`, and changing the metrics to be gauges instead of counters.

I still think that waiting for confirmation, and using `processed` rather than `confirmed` for preflight, are a good idea ... just our commitment level should be `confirmed` then, not `finalized`.

enriquefynn:
I agree that's not good to observe states that might change once they're finalized, but this makes the (already slow tests) much slower. We can also move the commitment to the config and use the `processed` commitment in the tests.

ruuda:
> We can also move the commitment to the config and use the processed commitment in the tests.

No, I think that would be bad, we should test what we ship. The tests failing after I changed the commitment uncovered that it was a lot more subtle than just a single setting, so let’s keep the tests in sync with what we actually run.

ruuda:
I changed the commitment level back to `confirmed`, and updated the affected metrics to be gauges instead of counters. Please take another look.

enriquefynn (approved):
LGTM!

PR #439, opened 2021-10-01
Write preliminary changelog for v1.1.0

ruuda:
This is a summary of the changes since v1.0.1. I think we still want to wait for #430 before we make a release, although, maybe we can just release this as v1.1.0 and then release v1.2.0 with active rebalancing and also staking/rebalancing only at the end of the epoch. Then we can already test v1.1.0 now. Other maintainer operators could wait until v1.2.0 before upgrading to reduce churn. What do you think?


enriquefynn (commented):
Great! We can now also add the changes from #440 and #430 for the new release.
Unstake from active validators and only unstake/stake when close to the epoch's end, except when removing a validator.

PR #440, opened 2021-10-04
Stake/Unstake at the end of the epoch

enriquefynn:
Adds a flag `stake_unstake_any_time`, if set to true, the maintainer
will try to stake/unstake whenever it's possible, otherwise, stake and
unstake operations will happen only close to the epoch's end.
Defaults to `false`.

To decide if the epoch is finishing, there's a `const
END_OF_EPOCH_THRESHOLD` set to `1/20`, if the remaining ratio for the
epoch's completion is greater than that, the epoch is considered ending
and maintainers will try to stake/unstake.

To verify the conditions, the function
`should_perform_maintenance_at_end_of_epoch` is called whenever there's
a stake/unstake operation in the maintainer.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -454,6 +454,12 @@ cli_opt_struct! {
         /// claim fees.
         #[clap(long)]
         validator_vote_account : Pubkey => Pubkey::default(),
+
+        /// Try to do stake and unstake operations any time if set to true,
+        /// otherwise will try to stake/unstake only at the end of the epoch.
+        /// Defaults to false.
+        #[clap(long, value_name = "true/false")]
+        stake_unstake_any_time: bool => false,
     }
 }
 
@@ -640,6 +646,12 @@ cli_opt_struct! {
         /// Maximum time to wait in seconds after there was no maintenance to perform, before checking again. Defaults to 30s
         #[clap(long)]
         max_poll_interval_seconds: u64 => 30,
+
+        /// Try to do stake and unstake operations any time if set to true,
+        /// otherwise will try to stake/unstake only at the end of the epoch.
+        /// Defaults to false.
+        #[clap(long, value_name = "true/false")]
+        stake_unstake_any_time: bool => false,
     }
 }
 
--- a/cli/src/maintenance.rs
+++ b/cli/src/maintenance.rs
@@ -306,6 +306,11 @@ pub struct SolidoState {
     /// Public key of the maintainer executing the maintenance.
     /// Must be a member of `solido.maintainers`.
     pub maintainer_address: Pubkey,
+
+    /// If set to true, stake and unstake instructions are issued whenever
+    /// possible, otherwise the instructions are issued only close to the end of
+    /// epoch.
+    pub stake_unstake_any_time: bool,
 }
 
 fn get_validator_stake_accounts(
@@ -388,11 +393,21 @@ impl SolidoState {
         numerator: 1,
         denominator: 10,
     };
+
+    /// Threshold for when to consider the end of an epoch.
+    /// E.g. if set to 1/20, the end of epoch would be considered if the system
+    /// is past 95% of the epoch's time.
+    const END_OF_EPOCH_THRESHOLD: Rational = Rational {
+        numerator: 1,
+        denominator: 20,
+    };
+
     /// Read the state from the on-chain data.
     pub fn new(
         config: &mut SnapshotConfig,
         solido_program_id: &Pubkey,
         solido_address: &Pubkey,
+        stake_unstake_any_time: bool,
     ) -> Result<SolidoState> {
         let solido = config.client.get_solido(solido_address)?;
 
@@ -477,6 +492,7 @@ impl SolidoState {
             epoch_schedule,
             stake_history,
             maintainer_address,
+            stake_unstake_any_time,
         })
     }
 
@@ -492,6 +508,7 @@ impl SolidoState {
 
     /// If there is a deposit that can be staked, return the instructions to do so.
     pub fn try_stake_deposit(&self) -> Option<(Instruction, MaintenanceOutput)> {
+        self.should_perform_maintenance_at_end_of_epoch()?;
         // We can only stake if there is an active validator. If there is none,
         // this will short-circuit and return None.
         self.solido.validators.iter_active().next()?;
@@ -606,6 +623,7 @@ impl SolidoState {
 
     /// If there is a validator being deactivated, try to unstake its funds.
     pub fn try_unstake_from_inactive_validator(&self) -> Option<(Instruction, MaintenanceOutput)> {
+        self.should_perform_maintenance_at_end_of_epoch()?;
         for (validator, stake_accounts) in self
             .solido
             .validators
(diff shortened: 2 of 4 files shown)


cli/src/config.rs
```diff
@@ -454,6 +454,12 @@ cli_opt_struct! {
         /// claim fees.
         #[clap(long)]
         validator_vote_account : Pubkey => Pubkey::default(),
+
+        /// Try to do stake and unstake operations any time if set to true,
+        /// otherwise will try to stake/unstake only at the end of the epoch.
+        /// Defaults to false.
+        #[clap(long, value_name = "true/false")]
+        stake_unstake_any_time: bool => false,
```
ruuda, line 462: Optional suggestion: we could make an enum

```rust
enum StakeTime {
    Anytime,
    OnlyNearEpochEnd,
}
// Would also need FromStr impl
```

Then it is clear in all the places where the value is used, what the two cases are. With `stake_unstake_any_time=false`, you know that we can’t stake/unstake at just any time, but it doesn’t tell you at what times it *is* possible, you have to refer to the doc comment of the field. If you see `stake_time=StakeTime::OnlyNearEpochEnd` it is more self-descriptive.

But this is a minor thing really, I’m also fine with `stake_unstake_any_time`.

cli/src/maintenance.rs
```diff
@@ -606,6 +623,7 @@ impl SolidoState {
 
     /// If there is a validator being deactivated, try to unstake its funds.
     pub fn try_unstake_from_inactive_validator(&self) -> Option<(Instruction, MaintenanceOutput)> {
+        self.should_perform_maintenance_at_end_of_epoch()?;
```
ruuda, line 626: This one we can leave unguarded. With unstake from active validators, it is a last resort, but with unstake from a validator that needs to be removed, we can unstake immediately.

cli/src/maintenance.rs
```diff
@@ -1356,6 +1375,37 @@ impl SolidoState {
             Some(self_slice_start_slot)
         }
     }
+
+    /// Return None if we observe we moved past `1 -
+    /// SolidoState::END_OF_EPOCH_THRESHOLD`%. Return Some(()) if the above
+    /// condition fails or `self.stake_unstake_any_time` is set to
+    /// `true`.
+    pub fn should_perform_maintenance_at_end_of_epoch(&self) -> Option<()> {
```
ruuda, line 1383: The `_at_end_of_epoch` is a bit misleading then, because it can also return `Some(())` at the start of the epoch if the flag is set. How about something like `confirm_should_stake_unstake_in_current_slot`?

cli/src/maintenance.rs
```diff
@@ -1356,6 +1375,37 @@ impl SolidoState {
...
+        if self.stake_unstake_any_time {
+            return Some(());
+        }
+        // Get the slot that the current epoch started.
+        let slots_epoch_begin = self.epoch_schedule.slots_per_epoch * self.clock.epoch;
```
ruuda, line 1388: This is not the right computation, due to some special behavior near genesis. There is `EpochSchedule::get_first_slot_in_epoch` for this instead.

cli/src/maintenance.rs
```diff
@@ -1648,4 +1704,36 @@ mod test {
...
+    #[test]
+    fn test_above_equal_epoch_threshold() {
+        let mut state = new_empty_solido();
+        state.stake_unstake_any_time = false;
+        state.clock.slot = 9195;
+        state.epoch_schedule.slots_per_epoch = 100;
+        state.clock.epoch = 91;
+        assert_eq!(state.should_perform_maintenance_at_end_of_epoch(), None);
+        state.clock.slot = 9199;
+        assert_eq!(state.should_perform_maintenance_at_end_of_epoch(), None);
```
ruuda, line 1727: Huh, aren’t these two tests the wrong way around? In `test_below_epoch_threshold` it should return `None` and here both should return `Some(())`? The implementation of `state.should_perform_maintenance_at_end_of_epoch` looks correct to me though. Am I missing something?

tests/test_solido.py
```diff
@@ -421,15 +421,22 @@ def add_validator(
...
+    return solido(
+        'perform-maintenance',
+        '--solido-address',
+        solido_address,
+        '--solido-program-id',
+        solido_program_id,
+        '--stake-unstake-any-time',
+        'true',
+        keypair_path=maintainer.keypair_path,
+    )
```
ruuda, line 435: :heart: Nice deduplication

ruuda (commented):
Looks almost good to me! I found one thing regarding the end-of-epoch computation, aside from that I only have superficial comments.

cli/src/maintenance.rs
enriquefynn: Yeah, that's much better 👍

cli/src/maintenance.rs
enriquefynn: Yeah, depends on the interpretation. Could be above the `1 - threshold`, but it's correct to swap the names 👍

cli/src/config.rs
enriquefynn: That's a good idea, thanks 👍

cli/src/maintenance.rs
```diff
@@ -1356,6 +1376,43 @@ impl SolidoState {
...
+                    .epoch_schedule
+                    .get_first_slot_in_epoch(self.clock.epoch + 1);
+                let slots_per_epoch = next_epoch_begin
+                    .checked_sub(slots_epoch_begin)
+                    .expect("Next epoch's slot should be always greater than previous.");
+                let slot_past_epoch = self.clock.slot.checked_sub(slots_epoch_begin).expect(
+        "Current slot is less than the beginning of the epoch's slot. This shouldn't happen.",
+    );
+                let ratio = Rational {
+                    numerator: slots_per_epoch.checked_sub(slot_past_epoch).expect(
```
ruuda, line 1402: Aah, I now undertand where my confusion about the tests came from, this function is wrong after all ... We should stake late in the epoch, but due to the subtract here, that is when `ratio` is close to zero. So the comparison below should be `ratio < SolidoState::END_OF_EPOCH_THRESHOLD`, not `>`. That’s why all the expected results of the tests are also flipped; the function should return `None` when it’s not yet time to stake (so the `?` in calls to `confirm_should_stake_unstake_in_current_slot()?` makes the caller return immediately).

One way to fix would be to swap flip the `>` to a `<`, but I think the end result will be simpler if we change the definition of `END_OF_EPOCH_THRESHOLD` to 19/20, change the definition of `ratio` here to just `slot_past_epoch/slots_per_epoch`, and leave the comparison as `ratio > END_OF_EPOCH_THRESHOLD`.

ruuda (commented):
I think the condition is the opposite of what it should be now. Aside from that it looks good.

cli/src/maintenance.rs
enriquefynn: Yeah, that's right 😬. I changed it, should be good now.

ruuda (approved):
LGTM :shipit:

PR #441, opened 2021-10-06
Fix parsing of `ChangeMultisig` in `multisig show-transaction`

ruuda:
Bugfix:

* Identifying the `ChangeMultisig` transaction relies on a 8-byte magic number derived from the multisig program module name. When we updated the multisig program, the module name changed, and therefore the magic number changed, so we should have changed this. We did not catch this in tests, because we only tested that we could execute the transaction, not that we could parse its details.
* Add a test to test that we can parse the details, to prevent this in the future.
* Fix the magic number to make the test pass.

Feature:

* Instead of displaying only the new multisig owners and threshold, display a diff, to make it easier to understand what you are approving.

New output:

```console
$ solido --config mainnet.json multisig show-transaction --transaction-address C5VLK2EBThXVw6UXmypZePuKkBiMUm3vMXY2BpftmsTk
Multisig: 3cXyJbjoAUNLpQsFrFJTTTp8GD3uPeabYbsCVobkQpD1
Did execute: false

Signers:
  [ ] Cv6GM219kzMrdUUdgDGVJUPW6fGosvrhsFrvmEhz3Mc6
  [ ] ENH1xvwjinUWkwEgw1hKduyAg7CrJMiKvr9nAS7wLHrp
  [ ] 6CawqfAJDviZGfUpHFJgeauq6H9vhKuivMMZULZeGnPw
  [ ] F4VFp4tFTyrQWo9YVjCbPE5eQP27ice2zyGDp2tN2Rkm
  [ ] AnoVUukL1fMAwEp4y2rrZV45BNHLes8ZwWsCRgEwhGH4
  [x] 6S21QCmpAadEhHj3pY2RMbPMGwgYNvS4Pd7zUXoRDMdK
  [ ] 8Lep9addZWUWqBNj3igx4QoHe43GBfvLhGJy18jJgWQa

Instruction:
  Program to call: AAHT26ecV3FEeFmL2gDZW6FfEqjPkghHbAkNZGqwT8Ww
  Accounts:

    * 3cXyJbjoAUNLpQsFrFJTTTp8GD3uPeabYbsCVobkQpD1
      signer: false, writable: true

    * GQ3QPrB1RHPRr4Reen772WrMZkHcFM4DL5q44x1BBTFm
      signer: true, writable: false

  This is a serum_multisig::set_owners_and_change_threshold instruction.
    Threshold (unchanged): 4 of 7
    Owners (changed):
        6CawqfAJDviZGfUpHFJgeauq6H9vhKuivMMZULZeGnPw
        6S21QCmpAadEhHj3pY2RMbPMGwgYNvS4Pd7zUXoRDMdK
        Cv6GM219kzMrdUUdgDGVJUPW6fGosvrhsFrvmEhz3Mc6
        ENH1xvwjinUWkwEgw1hKduyAg7CrJMiKvr9nAS7wLHrp
        AnoVUukL1fMAwEp4y2rrZV45BNHLes8ZwWsCRgEwhGH4
        F4VFp4tFTyrQWo9YVjCbPE5eQP27ice2zyGDp2tN2Rkm
      - 8Lep9addZWUWqBNj3igx4QoHe43GBfvLhGJy18jJgWQa
      + DHLXnJdACTY83yKwnUkeoDjqi4QBbsYGa1v8tJL76ViX
```

### DIFF
--- a/cli/src/multisig.rs
+++ b/cli/src/multisig.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: GPL-3.0
 
 use std::fmt;
+use std::collections::HashSet;
 
 use anchor_lang::prelude::{AccountMeta, ToAccountMetas};
 use anchor_lang::{Discriminator, InstructionData};
@@ -368,10 +369,15 @@ enum ParsedInstruction {
         spill_address: Pubkey,
     },
     MultisigChange {
-        threshold: u64,
+        old_threshold: u64,
 
         #[serde(serialize_with = "serialize_b58_slice")]
-        owners: Vec<Pubkey>,
+        old_owners: Vec<Pubkey>,
+
+        new_threshold: u64,
+
+        #[serde(serialize_with = "serialize_b58_slice")]
+        new_owners: Vec<Pubkey>,
     },
     SolidoInstruction(SolidoInstruction),
     TokenInstruction(TokenInstruction),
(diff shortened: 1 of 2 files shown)


enriquefynn (approved):
LGTM! The diff looks great ^^

Issue #438, opened 2021-09-27
Anchor integration


ruuda:
## Background

Integration with [Anchor](https://docs.anchorprotocol.com/) is one of the most requested features. After discussion with the Anchor/Terra team, on a high level it should look like this:

* A program on the Solana side
* Where users can deposit stSOL and receive bSOL
* And users can withdraw stSOL by returning the bSOL
* bSOL should be pegged to SOL 1:1
* The rewards get sent to Anchor users externally. Unlike stSOL, bSOL does not go up in value.
* So sometimes we need to sell some stSOL to make sure that the SOL value of the stSOL managed by the program is nominally equal to the bSOL supply.

## Design

* Solana program that manages the bSOL mint, similar to how Solido manages the stSOL mint.
* `Deposit` and `Withdraw` instructions.
* A `ClaimRewards` instruction that reads the exchange rate from the Solido state to determine what the excess stSOL is, and then sells it, and sends the rewards to Terra. Proceeds should be in UST.

## Open questions

* How do we sell stSOL? Likely on some AMM that has an stSOL / UST pair. We can create this pair if it doesn’t exist yet. But we need to figure out which AMM is easiest to integrate with programmatically. Ideally it would be one instruction we can call and the swap happens as part of the same transaction.
  * Saber
  * Raydium
  * Orca
  * Mercurial
  * Serum (?)
 
* What do we do with the proceeds ... how do we send them to Terra?
  * Can we use Wormhole?
  * Can we have a single address that we always send to; is Wormhole like a channel that is always open? Or is more like you have to jump through some hoops to set up a transfer, and you need to do this every time?
  * We need to have a destination address, but Anchor can provide this, once the Anchor side of the integration is done?
 
* Who calls `ClaimRewards`? A bot would. Let’s not worry right now about who runs it. If needed we could add it to the `solido` maintainer, but would be nice if Anchor can do it.

* Are there alternatives to waiting for Wormhole? Is there a bridge between Solana and Terra right now?

## Next steps

 * Build Solana program that manages a mint and has a `Deposit` and `Withdraw`. Let’s put it in this repo so we can reuse a lot of stuff.
 * Evaluate the AMMs and see which ones we can easily use to do a swap from a program.
    * [ ] Saber @enriquefynn 
    * [ ] Raydium @enriquefynn 
    * [ ] Orca @ruuda 
    * [ ] Mercurial @ruuda 
    * [ ] Serum (?)
 * Figure out how Wormhole transfers work.

[...]

PR #443, opened 2021-10-06
WIP: anchor integration init

enriquefynn:
Do not merge this PR until it's more mature.
Steps towards #438 (more information there).

### DIFF
--- a/anchor/src/instruction.rs
+++ b/anchor/src/instruction.rs
@@ -1,3 +1,6 @@
+// TODO(#449): Remove this once Anker functions are all complete.
+#![allow(dead_code)]
+
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
 
 use solana_program::{
--- a/anchor/src/lib.rs
+++ b/anchor/src/lib.rs
@@ -1,13 +1,21 @@
 // SPDX-FileCopyrightText: 2021 Chorus One AG
 // SPDX-License-Identifier: GPL-3.0
 
+use solana_program::pubkey::Pubkey;
+
 #[cfg(not(feature = "no-entrypoint"))]
 pub mod entrypoint;
 mod instruction;
 mod logic;
-mod processor;
-mod state;
-mod token;
+pub mod processor;
+pub mod state;
+pub mod token;
 
-/// Mint authority, mints StSol.
+/// Mint authority, mints bSOL.
 pub const ANCHOR_MINT_AUTHORITY: &[u8] = b"mint_authority";
+
+/// Return the address at which the Anchor instance should live that belongs to
+/// the given Solido instance.
+pub fn find_instance_address(anker_program_id: &Pubkey, solido_instance: &Pubkey) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&[solido_instance.as_ref()], anker_program_id)
+}
--- a/anchor/src/logic.rs
+++ b/anchor/src/logic.rs
@@ -24,7 +24,8 @@ pub fn deserialize_anchor(
 }
 
 /// Mint the given amount of bSOL and put it in the recipient's account.
-pub fn mint_b_sol_to<'a>(
+// TODO(#449): Remove underscore once this method is used.
+pub fn _mint_b_sol_to<'a>(
     anchor: &Anchor,
     anchor_address: &Pubkey,
     spl_token_program: &AccountInfo<'a>,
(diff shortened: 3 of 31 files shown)


anchor/src/entrypoint.rs
```diff
@@ -0,0 +1,24 @@
...
+    account_info::AccountInfo, entrypoint, entrypoint::ProgramResult, pubkey::Pubkey,
+};
+
+entrypoint!(process_instruction);
+fn process_instruction(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    instruction_data: &[u8],
+) -> ProgramResult {
+    if let Err(error) = processor::process(program_id, accounts, instruction_data) {
```
ruuda, line 19: This `if let` seems useless, can’t we just call `processor::process`? (Or if the error type needs to be mapped, call it with `?`)

anchor/src/instruction.rs
```diff
@@ -0,0 +1,102 @@
...
+        amount: StLamports,
+    },
+
+    /// Withdraw a given amount of stSOL.
+    ///
+    /// Caller provides some `amount` of StLamports that are to be burned in
+    /// order to withdraw bSOL.
+    Withdraw {
+        #[allow(dead_code)] // but it's not
+        amount: StLamports,
```
ruuda, line 32: Withdraw amount is in `BLamports`, not `StLamports`, right?

anchor/src/instruction.rs
```diff
@@ -0,0 +1,102 @@
...
+        // therefore return an IoError. But when serializing to a vec, there
+        // is no IO, so for this particular writer, it should never fail.
+        self.try_to_vec()
+            .expect("Serializing an Instruction to Vec<u8> does not fail.")
+    }
+}
+
+accounts_struct! {
+    InitializeAccountsMeta, InitializeAccountsInfo {
+        pub anchor {
```
ruuda, line 52: For the Lido program we were too late, but for the Anchor integration program, we can use this trick that Neodyme pointed out: make the instance address a program-derived address. We can use the Lido instance address as seed for the derivation. That way every Lido instance will have one unique associated Anchor instance.

(Nothing needs to be changed on this line for that, I’m just writing down the idea.)

anchor/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[cfg(not(feature = "no-entrypoint"))]
+pub mod entrypoint;
+mod instruction;
+mod logic;
+mod processor;
+mod state;
+mod token;
+
+/// Mint authority, mints StSol.
```
ruuda, line 12: s/StSol/bSOL/

anchor/src/logic.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use crate::state::Anchor;
+
+pub fn deserialize_anchor(
+    program_id: &Pubkey,
+    anchor: &AccountInfo,
+) -> Result<Anchor, ProgramError> {
+    if anchor.owner != program_id {
+        msg!(
+            "Lido state is owned by {}, but should be owned by the Lido program ({}).",
```
ruuda, line 16: s/Lido/Anchor/

anchor/src/logic.rs
```diff
@@ -0,0 +1,66 @@
...
+            anchor.owner,
+            program_id
+        );
+        return Err(LidoError::InvalidOwner.into());
+    }
+    let lido = try_from_slice_unchecked::<Anchor>(&anchor.data.borrow())?;
```
ruuda, line 22: s/lido/anchor/

anchor/src/logic.rs
```diff
@@ -0,0 +1,66 @@
...
+    Ok(lido)
```
ruuda, line 23: If we go with the idea of letting the Anchor instance live at a program-derived address derived from the Lido instance address, then this would be the place to verify.

anchor/src/state.rs
```diff
@@ -0,0 +1,32 @@
...
+use lido::util::serialize_b58;
+use serde::Serialize;
+use solana_program::{account_info::AccountInfo, entrypoint::ProgramResult, pubkey::Pubkey};
+
+#[repr(C)]
+#[derive(
+    Clone, Debug, Default, BorshDeserialize, BorshSerialize, BorshSchema, Eq, PartialEq, Serialize,
+)]
+pub struct Anchor {
+    /// The SPL Token mint address for bSOL.
```
ruuda, line 11: We should add a version byte, like we do with Lido.

Oh and about the version byte ... if we deserialize the Lido instance in this program, we need to check the version byte, and bail out if it doesn’t match (which means we upgraded the Lido program but not yet this program).

ruuda (commented):
Looks mostly good, I found a few places that were still referring to Lido things that should now have different names.

Issue #444, opened 2021-10-12
[Anchor] Deposit/Withdraw


enriquefynn:
Main issue: #438
Following the [anchor-integration](https://github.com/ChorusOne/solido/tree/anchor-integration) branch.

## Deposit / Withdraw
- Users should be able to deposit stSOL and get bSOL in return.
- Keep user's shares of the pool based on the amount of bSOL they have. bSOL tokens can be abstracted in shares and calculated on-the-fly if they can be not pegged 1:1 with SOL.
- To withdraw, users burn bSOL and get stSOL in return.
- Keep exchange rate similar to lido's to calculate amount to be withdrawn

ruuda:
Both deposit (#446) and withdraw (#460) are now implemented.

PR #446, opened 2021-10-12
Deposit instruction

enriquefynn:
Implement the deposit instruction for the Anchor integration, where a user deposits stSOL and gets bSOL in return.

Steps towards #444

### DIFF
--- a/anchor/src/processor.rs
+++ b/anchor/src/processor.rs
@@ -106,7 +106,7 @@ fn process_deposit(
 
     let lido = Lido::deserialize_lido(accounts.lido_program.key, accounts.lido)?;
 
-    // Use Lido's exchange rate (`st_sol_supply / sol_balance`) to compute the
+    // Use Lido's exchange rate (`sol_balance / sol_supply`) to compute the
     // amount of BLamports to mint.
     let sol_value = lido.exchange_rate.exchange_st_sol(amount)?;
     let b_sol_amount = BLamports(sol_value.0);


anchor/src/instruction.rs
```diff
@@ -85,6 +89,19 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
+        pub from_account {
+            is_signer: false,
```
ruuda, line 93: Doesn’t the from account need to sign? Or at least its owner?

anchor/src/state.rs
```diff
@@ -1,7 +1,72 @@
...
+use solana_program::{
+    account_info::AccountInfo, clock::Epoch, entrypoint::ProgramResult, msg, pubkey::Pubkey,
+};
+
+// Copied from lido::state
+#[repr(C)]
+#[derive(
+    Clone, Debug, Default, BorshDeserialize, BorshSerialize, BorshSchema, Eq, PartialEq, Serialize,
+)]
+pub struct ExchangeRate {
```
ruuda, line 18: For this program, I think we do not need to store the exchange rate, we can compute it on the fly from the reserve balance and the the bSOL supply in the mint, or we can read it from the Lido instance.

For Lido this was more complex because we store the SOL in multiple places, and we have the multi-step reward collection. But for this program, the only thing we need to do is put the stSOL in the reserve and mint the bSOL. That's only two accounts to deal with, we can just always have these two accounts as inputs, and compute the stSOL/bSOL exchange rate on the fly.

There are two ways in which we can compute the stSOL/bSOL exchange rate:

 * Empirical, `st_sol_in_reserve / b_sol_supply`.
 * Ideal, assuming that 1 bSOL = 1 SOL. Then the exchange rate is `st_sol_supply / sol_balance` stored in the Lido program.

When the two rates are different, that is when `ClaimRewards` should sell some stSOL to restore the exchange rate.

Which one should we use for deposit/withdrawals? I’m not really sure. I think going with the ideal exchange rate would be a bit better, because it eliminates one degree of freedom where the two rates can drift, and I have a vague feeling that that eliminates room for weird incentives that could otherwise exploit the difference, but I haven’t thought very hard about it.

anchor/src/processor.rs
```diff
@@ -36,7 +48,38 @@ fn process_deposit(
     amount: StLamports,
 ) -> ProgramResult {
     let accounts = DepositAccountsInfo::try_from_slice(accounts_raw)?;
+    if amount == StLamports(0) {
+        msg!("Amount must be greater than zero");
+        return Err(ProgramError::InvalidArgument);
+    }
+
     let anchor = deserialize_anchor(program_id, accounts.anchor)?;
+    if &anchor.reserve_account != accounts.to_reserve_account.key {
```
ruuda, line 57: So the reserve account is owned by the reserve authority? We could make it an associated token account. Then we don't need to store the reserve account address in the instance, and that eliminates one opportunity for mismatches. (Of course, we still need to verify that the account passed to the instruction call is the associated token account.)

anchor/src/instruction.rs
enriquefynn: Yeah, it was misnamed as `reserve_authority`, I replaced it for `user_authority`.

anchor/src/processor.rs
```diff
@@ -33,12 +71,57 @@ fn process_initialize(program_id: &Pubkey, accounts_raw: &[AccountInfo]) -> Prog
...
+            accounts.user_authority.clone(),
+            accounts.spl_token.clone(),
+        ],
+    )?;
+
+    let lido = Lido::deserialize_lido(accounts.lido_program.key, accounts.lido)?;
+
+    // Use Lido's exchange rate (`st_sol_supply / sol_balance`) to compute the
+    // amount of BLamports to mint.
+    let amount = BLamports(lido.exchange_rate.exchange_sol(Lamports(amount.0))?.0);
```
ruuda, line 114: This is the wrong way around, when you deposit stSOL, we want to know how much SOL that is worth, and mint the same amount of bSOL. So it should be

```rust
let sol_value = lido.exchange_rate.exchange_st_sol(amount)?;
let b_sol_amount = BLamports(sol_value.0);
```

The types become simpler too!

program/src/process_management.rs
```diff
@@ -8,14 +8,15 @@ use solana_program::{account_info::AccountInfo, entrypoint::ProgramResult, msg,
 
 use crate::logic::check_rent_exempt;
 use crate::processor::StakeType;
+use crate::state::Lido;
```
ruuda, line 11: The changes here look innocent to me, but if possible I would avoid changing things in `program`, so we don’t accidentally introduce a difference between the currently deployed program and the source code here.

program/src/state.rs
```diff
@@ -232,6 +232,18 @@ pub struct Lido {
...
+            msg!(
+                "Lido state is owned by {}, but should be owned by the Lido program ({}).",
+                lido.owner,
+                program_id
+            );
+            return Err(LidoError::InvalidOwner.into());
+        }
+        let lido = try_from_slice_unchecked::<Lido>(&lido.data.borrow())?;
+        Ok(lido)
+    }
```
ruuda, line 246: Nit: Please leave a blank line between methods.

anchor/src/processor.rs
enriquefynn: Oh, that's correct. Changed it

program/src/process_management.rs
enriquefynn: Yeah, in this case  we weren't exporting the `logic` module, I thought was simpler to move the function to Lido's state

PR #447, opened 2021-10-13
Fix error code when mismatched `st_sol_mint`

enriquefynn:


enriquefynn:
Let's leave it as it is then.

PR #448, opened 2021-10-14
Generalize test context so it can be used for Anchor tests

ruuda:
* [x] Extract test context into a crate so it can be reused
* [x] Layer a context on top for testing the Anchor integration

### DIFF
--- a/anchor/src/instruction.rs
+++ b/anchor/src/instruction.rs
@@ -1,4 +1,4 @@
-// TODO: Remove this once Anker functions are all complete.
+// TODO(#449): Remove this once Anker functions are all complete.
 #![allow(dead_code)]
 
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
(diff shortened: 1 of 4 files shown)


anchor/src/lib.rs
```diff
@@ -1,13 +1,21 @@
...
+pub mod state;
+pub mod token;
 
-/// Mint authority, mints StSol.
+/// Mint authority, mints bSOL.
 pub const ANCHOR_MINT_AUTHORITY: &[u8] = b"mint_authority";
+
+/// Return the address at which the Anchor instance should live that belongs to
+/// the given Solido instance.
+pub fn find_instance_address(anker_program_id: &Pubkey, solido_instance: &Pubkey) -> (Pubkey, u8) {
```
enriquefynn, line 19: Does that mean that Solido has to create the Anker instance? I think this is not needed. They should be two independent programs.

PR #452, opened 2021-10-22
Rename "Anchor" references to "Anker"

ruuda:
To distinguish our [Anchor Protocol](https://anchorprotocol.com/) integration from the real Anchor program on Terra, and from [Anchor-the-eDSL](https://github.com/project-serum/anchor) on Solana, we decided to call our part of the integration "Anker", so we have an unambiguous name to refer to it.

This pull request renames all "anchor" references to "anker". The remaining occurrences of

    git grep --ignore-case anchor

are all related to Anchor-the-eDSL, or they explain Anker in terms of Anchor-the-protocol in comments and the readme.

This pull request should not contain any functional changes.


ruuda:
It’s failing in `split_stake_account` in `process_withdraw`:

    Program failed to complete: Could not create program address with signer seeds: Provided seeds do not result in a valid address


We do not pass any additional seeds in that call, but that is not something that changed in this PR :thinking:

ruuda:
So it turns out, if the program id does not lie on the curve, that error happens. :woman_shrugging: I replaced it with a program id on the curve and now it should pass.

ruuda:
The tests pass now, build failure is something in test coverage data merging going wrong that shouldn’t be related to this PR.

@enriquefynn can you please review?

enriquefynn (approved):
LGTM

PR #455, opened 2021-10-26
Anker claim rewards

### LINKED ISSUE #445: [Anchor] ClaimRewards + SendRewards
Main issue: #438
Following the [anchor-integration](https://github.com/ChorusOne/solido/tree/anchor-integration) branch.

## ClaimRewards
stSOL rewards should be claimed, this will sell stSOL to TerraUSD on Orca (as decided in #438). Once the rewards were sold, they have to be transferred to the Terra chain by using Wormhole.
We can restrict the function to be called only in the beginning of every epoch by anyone.

@RishiSid can update on this issue once done with the Wormhole tests.

enriquefynn:
Implement the `ClaimRewards` instruction.
Selling StSol on a market in exchange for UST that goes to a predefined address.

- Rewards are calculated based on `reserve_st_sol - solido_exchange(b_sol_supply)`.
- Only proceed if there were some accrued rewards.
- Rewards are immediately sold in a predefined AMM for UST tokens, they'll live in a predefined account stored in the state as `rewards_destination`.

Closes #445


anker/src/processor.rs
```diff
@@ -167,13 +178,40 @@ fn process_deposit(
...
+    let st_sol_reserve_state =
+        spl_token::state::Account::unpack_from_slice(&accounts.reserve_account.data.borrow())?;
+    let reserve_st_sol = StLamports(st_sol_reserve_state.amount);
+
+    // Get StLamports corresponding to the amount of b_sol minted.
+    let st_sol_amount = lido.exchange_rate.exchange_sol(Lamports(b_sol_supply))?;
+
+    // If `reserve_st_sol` < `st_sol_amount` something went wrong, and we abort the transaction.
+    let rewards = (reserve_st_sol - st_sol_amount)?;
+    swap_rewards(rewards, &anker, &accounts)
```
ruuda, line 205: :+1: Beautiful, very straightforward function

anker/src/token.rs
```diff
@@ -12,3 +12,4 @@ use std::{
 use lido::token::Result;
 
 impl_token!(BLamports, "bSOL");
+impl_token!(UstLamports, "UST");
```
ruuda, line 15: Beware! One of the UST tokens has 6 decimals, not 9. Using the term “lamport” would be misleading then, and the string formatter would be all wrong. But the other UST token does have 9 decimals. I don’t know which is the right one, I asked for clarification from the Terra people:

* https://solscan.io/token/9vMJfxuKxXBoEa7rM12mYLMwTacLMLDJqHozw96WQL8i (6 decimals)
* https://solscan.io/token/CXLBjMMcwkc17GfJtBos6rQCo1ypeH6eDbB82Kby4MRm (9 decimals)

ruuda (commented):
Looks mostly good! I have a few comments.

anker/src/token.rs
ruuda: Kevin clarified that the right one is 9vMJfxuKxXBoEa7rM12mYLMwTacLMLDJqHozw96WQL8i, because that’s the one listed for the Wormhole v2 docs at https://docs.wormholenetwork.com/wormhole/overview-liquid-markets.

So we have a token with 6 decimals now :confused: We could take the format string or maybe the number of decimals as an argument to the macro, it only matters for presentation, in the end we are just putting the number of the token’s smallest denomination in a `u64`.

enriquefynn: Made the `impl_token` receive a decimals arg. But it's still printing 9 digits in the literal formatter `"{}.{:0>9} {}"`. Is there a nice way to pass the string literal with the alignment as `write!` parameter?

ruuda: I think we will need to pass in the format string as an argument to the macro :confused: I don’t think `std::fmt` allows the width to be chosen at runtime. There is `std::fmt::Arguments`, but it can only be constructed from a macro.

program/src/token.rs
```diff
@@ -51,7 +51,7 @@ pub type Result<T> = std::result::Result<T, ArithmeticError>;
 /// only used for `Debug` and `Display` printing.
 #[macro_export]
 macro_rules! impl_token {
-    ($TokenLamports:ident, $symbol:expr) => {
+    ($TokenLamports:ident, $symbol:expr, $decimals:expr) => {
```
ruuda, line 54: To make the call site a bit more obvious, we might write
```rust
($TokenLamports:ident, $symbol:expr, decimals = $decimals:expr) => {
```
or even
```rust
(type = $TokenLamports:ident, symbol = $symbol:expr, decimals = $decimals:expr) => {
```
and then at the call site it would become
```
impl_token!(UstLamports, "UST", decimals = 6);
```
so the meaning of the magic number is immediately clear.

anker/src/token.rs
ruuda: Oh actually what we might do is just string-format the number into a `String` and then left-pad the string! Or we could format it to 9 digits and then slice the resulting &str from the right, cutting off any excess zeros on the left.

program/src/token.rs
ruuda, line 54: To make the call site a bit more obvious, we might write
```rust
($TokenLamports:ident, $symbol:expr, decimals = $decimals:expr) => {
```
or even
```rust
(type = $TokenLamports:ident, symbol = $symbol:expr, decimals = $decimals:expr) => {
```
and then at the call site it would become
```
impl_token!(UstLamports, "UST", decimals = 6);
```
so the meaning of the magic number is immediately clear.

enriquefynn: Great, let's go for the describing the `decimals` magic number. The rest is more intuitive.

anker/src/processor.rs
```diff
@@ -66,35 +74,73 @@ fn process_initialize(program_id: &Pubkey, accounts_raw: &[AccountInfo]) -> Prog
...
+        &ust_reserve_account_seeds,
+        accounts.ust_reserve_account,
+        accounts.ust_mint,
     )?;
-    initialize_reserve_account(&accounts, &reserve_account_seeds)?;
+
+    let (_, token_swap_bump_seed) =
+        Pubkey::find_program_address(&[&accounts.pool.key.to_bytes()], &spl_token_swap::id());
 
     let anker = Anker {
```
ruuda, line 121: Note to self: I wanted to ask, don’t we need to store the UST mint address somewhere? But no we don’t, we can always extract it from our UST reserve. And we trust the UST reserve, because it lives at a program-derived address that we verify.

anker/src/state.rs
```diff
@@ -32,6 +34,14 @@ pub struct Anker {
     #[serde(serialize_with = "serialize_b58")]
     pub b_sol_mint: Pubkey,
 
+    /// Token swap data. Used to swap stSOL for UST.
+    #[serde(serialize_with = "serialize_b58")]
+    pub pool: Pubkey,
```
ruuda, line 39: Nit: maybe call it `token_swap_pool` or something? “pool” is rather generic, it might be confused for the reserve.

anker/src/state.rs
```diff
@@ -107,22 +123,43 @@ impl Anker {
...
+    pub fn check_ust_reserve_address(
+        &self,
+        anker_program_id: &Pubkey,
+        anker_instance: &Pubkey,
+        ust_reserve_account_info: &AccountInfo,
+    ) -> ProgramResult {
+        self.check_derived_account_address(
+            "the UST reserve account",
+            ANKER_UST_RESERVE_ACCOUNT,
+            self.st_sol_reserve_account_bump_seed,
```
ruuda, line 159: This should be the `ust_reserve_account_bump_seed`, not `st_sol`.

anker/src/state.rs
```diff
@@ -235,6 +272,102 @@ impl Anker {
...
+        // Check token swap instance parameters.
+        if &self.pool != accounts.pool.key {
+            msg!(
+                "Invalid Token Swap instance, expected {}, found {}",
+                self.pool,
+                accounts.pool.key
+            );
+            return Err(AnkerError::WrongSplTokenSwap.into());
+        }
+        // We should ignore the 1st byte for the unpack.
```
ruuda, line 295: I think I commented on this before, did I miss the reply? Why do we need to skip the first byte, and shouldn’t we validate the value?

anker/src/state.rs
```diff
@@ -235,6 +272,102 @@ impl Anker {
...
+            accounts.pool_fee_account.key
+        );
+            return Err(AnkerError::WrongSplTokenSwapParameters.into());
+        }
+
+        // Check rewards destination.
+        // The reserve address is checked in `deserialize_anker`, this function
+        // should be called prior to this. We don't need to check the reserve
+        // authority, as the transaction will fail if a different one is provided.
+        if &self.rewards_destination != accounts.rewards_destination.key {
```
ruuda, line 360: Isn’t the rewards destination the `ust_reserve` now? Let’s call it like that to avoid confusion.

program/src/token.rs
```diff
@@ -72,8 +72,8 @@ macro_rules! impl_token {
                 write!(
                     f,
                     "{}.{:0>9} {}",
-                    self.0 / 1_000_000_000,
-                    self.0 % 1_000_000_000,
+                    self.0 / 10u64.pow($decimals),
+                    self.0 % 10u64.pow($decimals),
```
ruuda, line 76: Note to self: I wonder if LLVM turns this into a constant

testlib/src/anker_context.rs
```diff
@@ -213,6 +306,79 @@ impl Context {
...
+            &mut self.solido_context.context,
+            &mut self.solido_context.nonce,
+            &[swap_instruction],
+            vec![authority],
+        )
+        .await
+        .expect("Failed to swap StSol for UST tokens.");
+    }
+
+    pub async fn sell_rewards(&mut self) {
```
ruuda, line 347: We should also have a `try_sell_rewards`, so we can try some invalid calls (e.g. passing in the wrong address that receives the proceeds) and confirm that they fail. But we can do that in the test PR, this one is large enough already.

ruuda (commented):
I have a few more comments.

Also, I realized, if the instance learns the UST mint address at initialization (and also the bSOL mint address), and the instance lives at a program-derived address, then this opens up one possible attack:
* We upload the program, but don’t initialize the instance yet.
* Before we are able to initialize the instance, somebody else does it, passing in the wrong token address for UST.
* Now we have a poisoned instance, and we can’t create a new one because the address is fixed.

I don’t think this is a big problem because:

* We can sidestep the problem by uploading the program to a new address, it will have a new instance address as well. So in the worst case, somebody could make us waste the rent for the program. (And we could reclaim that.)
* To exploit, somebody would have to be monitoring the chain for uploaded programs and quickly call `initialize` in the right way. That would be a lot of work and very targeted, for just a small impact.
* It is only exploitable in a small time window between uploading the program and initializing the instance.

I’ll open an issue for this so we can record it as a known issue. I don’t thing we need to take further action.

anker/src/lib.rs
```diff
@@ -21,15 +21,18 @@ pub const ANKER_MINT_AUTHORITY: &[u8] = b"mint_authority";
 pub const ANKER_RESERVE_AUTHORITY: &[u8] = b"reserve_authority";
 
 /// Anker's reserve account. Holds StSOL.
-pub const ANKER_RESERVE_ACCOUNT: &[u8] = b"reserve_authority";
+pub const ANKER_STSOL_RESERVE_ACCOUNT: &[u8] = b"reserve_account";
```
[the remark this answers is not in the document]
enriquefynn: Yeah, let's do it like that ^^

anker/src/state.rs
enriquefynn: 👍  Good catch!

anker/src/state.rs
enriquefynn: Yeah, but the Terra rewards destination should also exist in the state. I'll rename them accordingly.

anker/src/state.rs
enriquefynn: Done! ^^ The value should be 1

anker/src/instruction.rs
```diff
@@ -89,11 +89,11 @@ accounts_struct! {
             is_writable: false,
         },
         // Instance of the token swap data used for trading StSOL for UST.
-        pub pool {
+        pub token_swap_pool {
             is_signer: false,
             is_writable: false,
         },
-        pub rewards_destination {
+        pub terra_rewards_destination {
```
ruuda, line 96: I think we will eventually need to put this in the payload and not in the accounts, because I think it is going to be a different format pubkey, right? But for now this is fine, we can change it when we get to the Wormhole part.

anker/src/instruction.rs
```diff
@@ -287,7 +287,7 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
-        pub rewards_destination {
+        pub ust_reserve {
```
ruuda, line 290: <del>Wait, but now we reference two UST accounts: `ust_token` and `ust_reserve`. Shouldn’t that be one account?</del><ins>Ah, of course, the two `_token` accounts are the ones owned by the pool, and two are the ones owned by us.</ins>

Maybe we can call the ones owned by the pool `st_sol_pool_account` and `ust_pool_account`, to distinguish them from our `st_sol_reserve_account` and `ust_reserve`? (Should we also call it `_account`? And let’s put them next to each other, it doesn’t really matter for us with named accounts, but it matters for what you see on block explorers.)

program/src/token.rs
```diff
@@ -69,11 +69,18 @@ macro_rules! impl_token {
...
-                    "{}.{:0>9} {}",
+                let str_number = format!(
+                    "{}.{:0>9}",
                     self.0 / 10u64.pow($decimals),
                     self.0 % 10u64.pow($decimals),
+                );
+                write!(
+                    f,
+                    "{} {}",
+                    &str_number[..str_number.len() - (9 - $decimals)],
```
ruuda, line 80: This does not look right to me. If we have fewer decimals, we should cut out the zeros after the decimal dot, not the zeros at the end.

Please add a unit test for this to verify that the output is correct. E.g.

```rust
#[test]
fn test_token_format() {
    assert_eq!(format!("{}", Lamports(1)), "0.000000001 SOL");
    assert_eq!(format!("{}", Lamports(1_000_000_002)), "1.000000002 SOL");
    assert_eq!(format!("{}", UstLamports(1)), "0.000001 UST");
    assert_eq!(format!("{}", UstLamports(1_000_000_002)), "1000.000002 UST");
}
```

Also, now that I am writing this, I realize that we need to replace the `let mut exponent = 9_i32` in the `FromStr` impl with `let mut exponent: i32 = $decimals;`. There is a test `test_lamports_from_str_roundtrip` already. We can extend this test to test `UstLamports` in addition to `Lamports`. (Also, maybe the term “lamports” is confusing here when it doesn’t mean 10<sup>-9</sup>, we could call it `MicroUst` or something.)

testlib/src/anker_context.rs
```diff
@@ -404,12 +404,12 @@ pub async fn initialize_token_pool(
 
     // When packing the SwapV1 structure, `SwapV1::pack(swap_info, &mut
     // dst[1..])` is called. But the program also wants the size of the data
-    // to be `spl_token_swap::state::SwapV1::LEN`. That is why we add the
-    // `+1` to the size 🤷 .
+    // to be `spl_token_swap::state::SwapV1::LEN`. `LATEST_LEN` is 1 +
+    // SwapV1::LEN 🤷.
```
ruuda, line 408: The 🤷 mystery is solved now, right? It is a version byte, isn’t it?

ruuda (commented):
I found some more things, but I am hopeful that after those this is ready to merge :grimacing:

anker/src/instruction.rs
enriquefynn: Yeah, I think so. I think it's better to store it in Anker's state so it would be slightly more difficult to change.

testlib/src/anker_context.rs
enriquefynn: That's it ^^

ruuda (approved):
LGTM :shipit:

PR #462, opened 2021-11-22
Anchor claim rewards tests

enriquefynn:
Tests for the claim rewards functionality.
- Tests a successful claim rewards.
- Tests a swap operation on the AMM.
- Download the real Orca program for CI tests.


anker/tests/tests/amm.rs
```diff
@@ -0,0 +1,34 @@
+use lido::token::Lamports;
+use solana_program::program_pack::Pack;
+use solana_program_test::tokio;
+use solana_sdk::signer::Signer;
+use testlib::anker_context::Context;
+
+#[tokio::test]
+async fn test_successful_token_swap() {
+    let mut context = Context::new_with_initialized_token_pool().await;
+    let (st_sol_kp, st_sol_token) = context
```
ruuda, line 10: Nit: please write `keypair` instead of abbreviating it to keep the codebase readable for newcomers who may not be intimately familiar with the terminology.

anker/tests/tests/amm.rs
```diff
@@ -0,0 +1,34 @@
...
+    let ust_address = context
+        .solido_context
+        .create_spl_token_account(
+            context.token_pool_context.ust_mint_address,
+            st_sol_kp.pubkey(),
+        )
+        .await;
+
+    context
+        .swap_st_sol_for_ust(&st_sol_token, &ust_address, &st_sol_kp, 1_000_000_000, 0)
```
ruuda, line 24: I only notice this now, but this function takes a naked `u64` for `amount_in` and `minimum_amount_out`. We can make it more self-documenting and harder to misuse by using the stronger types `StLamports` and `UstLamports`: that makes it impossible to accidentally confuse the two.

Also, right now I have to jump to the definition of `swap_st_sol_for_ust` to understand the meaning of the last two arguments. (I can guess that the 1_000_000_000 is the amount in, but I wasn’t sure about the 0 until I looked it up.) We can make it more self-documenting by giving them names:

```rust
let amount_in = StLamports(1_000_000_000);
let min_amount_out = UstLamports(0);
context.swap_st_sol_for_ust(..., amount_in, min_amount_out).await;
```

anker/tests/tests/amm.rs
```diff
@@ -0,0 +1,34 @@
...
+        .await;
+
+    let ust_account = context.solido_context.get_account(ust_address).await;
+    let ust_spl_account: spl_token::state::Account =
+        spl_token::state::Account::unpack_from_slice(ust_account.data.as_slice()).unwrap();
```
ruuda, line 29: For stSOL and bSOL we have a helper method on the context to get the balance of an account, would be nice to add one for UST as well; it removes some of the implementation detail noise from the test, and it can return a more strongly typed result.

anker/tests/tests/claim_rewards.rs
```diff
@@ -0,0 +1,38 @@
...
+        .await;
+    let ust_spl_account: spl_token::state::Account =
+        spl_token::state::Account::unpack_from_slice(ust_account.data.as_slice()).unwrap();
+
+    // Exchange rate is 12 stSol : 13 Sol
+    // We have 1 stSOL, our rewards were 1 - (1 * 12/13) = 0.076923077
+    // Initially there are 10 StSol and 10 UST in the AMM
+    // We should get 10 - (10*10 / 10.076923077) = 0.07633587793834806 UST
+    assert_eq!(ust_spl_account.amount, 76335877);
+}
```
ruuda, line 38: We can also test that claiming a second time, after already claiming once this epoch, fails the transaction (because there is 0 excess to claim). But thinking about that, should we make it fail? I think claiming 0 is fine, it’s just a waste of your transaction fee.

anker/tests/tests/claim_rewards.rs
```diff
@@ -0,0 +1,38 @@
...
+    context.solido_context.update_exchange_rate().await;
+
+    context.claim_rewards().await;
+
+    let ust_account = context
+        .solido_context
+        .get_account(context.ust_rewards_account)
+        .await;
+    let ust_spl_account: spl_token::state::Account =
+        spl_token::state::Account::unpack_from_slice(ust_account.data.as_slice()).unwrap();
```
ruuda, line 31: A helper method to get the UST balance would already pay off here, now that it’s used twice :)

anker/tests/tests/amm.rs
```diff
@@ -0,0 +1,34 @@
+use lido::token::Lamports;
```
ruuda, line 1: Thanks for testing this! It’s good to have a sanity check.

testlib/src/anker_context.rs
```diff
@@ -77,7 +79,12 @@ impl TokenPoolContext {
         // Transfer some UST and StSOL to the pool.
         self.mint_ust(solido_context, &self.ust_address, ust_amount)
             .await;
-        let (kp_stsol, token_st_sol) = solido_context.deposit(Lamports(10_000_000_000)).await;
+        let solido = solido_context.get_solido().await;
+        let sol_amount = solido
+            .exchange_rate
+            .exchange_st_sol(st_sol_amount)
+            .expect("Some StSol should have been minted at this point.");
+        let (kp_stsol, token_st_sol) = solido_context.deposit(sol_amount).await;
```
ruuda, line 87: Also here, I would call it `keypair`, `kp` may be a bit cryptic.

testlib/src/anker_context.rs
```diff
@@ -327,10 +394,12 @@ impl Context {
...
 ///
-/// Fund UST and StSOL with 10 * 1e9 Lamports each.
-pub async fn initialize_token_pool(
-    solido_context: &mut solido_context::Context,
-) -> TokenPoolContext {
+/// The stake pool is not initialized at the end of this function.  To
+/// initialize the token swap instance, it requires funded token pairs on the
+/// liquidity pool.
+/// To get a new Context with an initialized token pool, call
+/// `Context::new_with_liquidity_on_amm`.
```
ruuda, line 401: This function does not exist.

ruuda (commented):
Great! I have a few comments to make it a bit nicer :)

anker/tests/tests/claim_rewards.rs
enriquefynn: Did it ^^

anker/tests/tests/claim_rewards.rs
enriquefynn: Should fail, because the token swap program doesn't allow swaps with 0 tokens.

anker/tests/tests/amm.rs
```diff
@@ -0,0 +1,39 @@
...
+async fn test_successful_token_swap() {
+    let mut context = Context::new_with_initialized_token_pool().await;
+    let (st_sol_keypair, st_sol_token) = context
+        .solido_context
+        .deposit(Lamports(10_000_000_000))
+        .await;
+
+    let ust_address = context
+        .solido_context
+        .create_spl_token_account(
```
ruuda, line 17: We might add a `create_ust_token_account` shorthand to the `AnkerContext` to make this test a bit more high-level to read.

anker/tests/tests/sell_rewards.rs
```diff
@@ -0,0 +1,59 @@
...
+        )
+        .await;
+    // Update the exchange rate so we see some rewards.
+    context.solido_context.advance_to_normal_epoch(1);
+    context.solido_context.update_exchange_rate().await;
+
+    context.sell_rewards().await;
+
+    let ust_balance = context.get_ust_balance(context.ust_reserve).await;
+    // Exchange rate is 12 stSol : 13 Sol
```
ruuda, line 28: Where do these 12 and 13 come from? The pool gets initialized with 10 stSOL and 10 UST, right? And we add a 1 SOL deposit, and then 1 SOL of rewards ... shouldn’t we have 12 SOL and 11 stSOL? Where does the additional one come from?

anker/tests/tests/sell_rewards.rs
```diff
@@ -0,0 +1,59 @@
...
+    // We have 1 stSOL, our rewards were 1 - (1 * 12/13) = 0.076923077
+    // Initially there are 10 StSol and 10 UST in the AMM
+    // We should get 10 - (10*10 / 10.076923077) = 0.07633587793834806 UST
+    assert_eq!(ust_balance, MicroUst(76335877));
```
ruuda, line 32: This is not 0.07633587793834806 UST, it is 76.335877 UST, right? It’s easier to see with numeric underscores: `MicroUst(76_335_877)`, and UST has 6 decimals.

anker/tests/tests/sell_rewards.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    // Test claiming the reward again fails.
+    let result = context.try_sell_rewards().await;
+    assert_solido_error!(result, AnkerError::ZeroRewardsToClaim);
+}
+
+#[tokio::test]
+async fn test_rewards_fail_with_different_reserve() {
```
ruuda, line 40: :+1: Thanks for testing this.

testlib/src/anker_context.rs
```diff
@@ -118,12 +125,11 @@ impl Context {
...
         let payer = solido_context.context.payer.pubkey();
 
-        let token_pool_context = initialize_token_pool(&mut solido_context).await;
+        let token_pool_context = setup_token_pool(&mut solido_context).await;
 
         let rewards_owner = solido_context.deterministic_keypair.new_keypair();
-        let ust_rewards_account = solido_context
-            .create_spl_token_account(token_pool_context.ust_mint_address, rewards_owner.pubkey())
-            .await;
+        // TODO: Replace with a valid Terra address.
```
ruuda, line 131: Hmm, so I looked briefly into this, it looks like Terra addresses are variable-length? Found this: <https://docs.rs/cosmwasm-terra/0.7.2/cosmwasm_terra/types/struct.CanonicalAddr.html>. and also this: https://docs.cosmos.network/master/basics/accounts.html#keys-accounts-addresses-and-signatures.

So it looks like they support multiple signature schemes, and addresses are byte arrays? We will have to dive a bit deeper later.

testlib/src/anker_context.rs
```diff
@@ -182,6 +188,66 @@ impl Context {
...
+    // provider AMM and initialized token pool.
+    pub async fn new_with_initialized_token_pool() -> Context {
+        const LIQUIDITY_AMOUNT: u64 = 10_000_000_000;
+        let mut context = Context::new().await;
+        context
+            .token_pool_context
+            .provide_liquidity(
+                &mut context.solido_context,
+                StLamports(LIQUIDITY_AMOUNT),
+                MicroUst(LIQUIDITY_AMOUNT),
```
ruuda, line 201: Ah, that’s why the decimal point was off earlier. This is not 10 UST, it is 10,000 UST; UST only has 6 decimals. I think we should just inline the numbers
```rust
StLamports(10_000_000_000),
MicroUst(10_000_000),
```

ruuda (commented):
Nice! I left a few comments but some are just observations, I think the only thing that stood out is that the UST amount is wrong by a factor 1000 now, because UST has 6 decimals, not 9.

anker/tests/tests/sell_rewards.rs
enriquefynn: `Context::new()` also deposits 1 Sol.

anker/tests/tests/sell_rewards.rs
enriquefynn: Let's make `10_000` UST liquidity in the pool.

testlib/src/anker_context.rs
enriquefynn: 👀  We will, can you write an issue about it?

testlib/src/anker_context.rs
enriquefynn: Let's make 10_000 UST, and change the swapped amount in the tests above.

anker/tests/tests/amm.rs
```diff
@@ -0,0 +1,36 @@
...
+            &ust_address,
+            &st_sol_keypair,
+            amount_in,
+            min_amount_out,
+        )
+        .await;
+
+    let ust_balance = context.get_ust_balance(ust_address).await;
+    // For the constant product AMM:
+    // 10 - (10*10 / 11) = 0.9090909090909083
```
ruuda, line 34: This comment is still off-by-1000, that’s 909.090909 UST now.

ruuda (approved):
Found one outdated comment, LGTM aside from that!

PR #464, opened 2021-11-24
Add more tests for Anker deposits

ruuda:
This adds two tests:

 * Test that deposit fails if you pass in the wrong reserve. (The bug that we had in Solido v0.5 before the audit, and that we added a regression test for there. We should pro-actively test it here.)
 * Test that the program fails if the instance doesn’t live at the right address. For this I modified the `solana-program-test` framework to add the ability to arbitrarily modify accounts, a kind of “god mode” to make it easier to write tests. To use that, override the dependencies so we use our fork instead of the crates.io version.
 * Test that withdraw uses the right exchange rate when the stSOL value goes down.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,15 @@ panic = "abort"
 [profile.release]
 panic = "abort"
 
+# We take our fork of Solana 1.7.8, which has one patch applied:
+# https://github.com/solana-labs/solana/pull/21415.
+# This patch enables testing more situations using solana-program-test.
+# We need to override the other Solana packages as well because they are interdependent,
+# and if we don’t override them, Cargo will include two incompatible versions of those
+# packages (one from crates.io and one from the fork).
 [patch.crates-io]
-solana-program-test = { path = "../solana/program-test" }
-solana-sdk = { path = "../solana/sdk" }
-solana-program = { path = "../solana/sdk/program" }
-solana-frozen-abi-macro = { path = "../solana/frozen-abi/macro" }
-solana-frozen-abi = { path = "../solana/frozen-abi" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -14,3 +14,16 @@ panic = "abort"
...
 panic = "abort"
+
+# We take our fork of Solana 1.7.8, which has one patch applied:
+# https://github.com/solana-labs/solana/pull/21415.
+# This patch enables testing more situations using solana-program-test.
+# We need to override the other Solana packages as well because they are interdependent,
+# and if we don’t override them, Cargo will include two incompatible versions of those
+# packages (one from crates.io and one from the fork).
+[patch.crates-io]
+solana-frozen-abi = { git = "https://github.com/ChorusOne/solana", branch = "program-test-178" }
```
enriquefynn, line 25: That's merged, and it seems rebased for >= 1.7.8 👀.

ruuda: :eyes: https://github.com/solana-labs/solana/pull/21418 targeting 1.8 though. We can remove the overrides here once we are able to upgrade.

anker/tests/tests/withdraw.rs
```diff
@@ -126,3 +131,61 @@ async fn test_withdraw_wrong_token_mint() {
         .withdraw(&owner, b_sol_account, BLamports(250_000_000))
         .await;
 }
+
+#[tokio::test]
+async fn test_withdraw_after_st_sol_price_decrease() {
```
enriquefynn, line 136: Nice test! ^^

enriquefynn (approved):

PR #465, opened 2021-11-26
Add CLI command to create a new Anker instance

### LINKED ISSUE #456: [Anchor] Add command to create a new Anker instance
We have a command `solido create-solido` to create a new Solido instance. We should add a command `create-anker` to create a new Anker instance as well. Just like `create-solido`, it will need to accept the mint, so we can create the mint ahead of time.

ruuda:
Closes #456.

To do:

* [x] Add a test script that calls this to initialize the Anker instance.
* [ ] ~~Add a `show-anker` subcommand so we can test that initialization was successful.~~ (Will do this in a follow-up PR.)


ruuda:
@enriquefynn I’ll rebase this on top of #455 after that one is merged, as it affects initialization.

cli/src/commands_anker.rs
```diff
@@ -0,0 +1,119 @@
...
+    let (mint_authority, _bump_seed) =
+        anker::find_mint_authority(opts.anker_program_id(), &anker_address);
+    let (st_sol_reserve_account, _bump_seed) =
+        anker::find_st_sol_reserve_account(opts.anker_program_id(), &anker_address);
+    let (ust_reserve_account, _bump_seed) =
+        anker::find_ust_reserve_account(opts.anker_program_id(), &anker_address);
+    let (reserve_authority, _bump_seed) =
+        anker::find_reserve_authority(opts.anker_program_id(), &anker_address);
+
+    let b_sol_mint_address = {
```
enriquefynn, line 67: Nice to allow optionally making the mint address.

cli/src/commands_anker.rs
```diff
@@ -0,0 +1,119 @@
...
+            solido: *opts.solido_address(),
+            solido_program: *opts.solido_program_id(),
+            st_sol_mint: solido.st_sol_mint,
+            b_sol_mint: b_sol_mint_address,
+            st_sol_reserve_account,
+            ust_reserve_account,
+            reserve_authority,
+            // TODO(#462): Make this configurable, but then we need to figure out
+            // the address format first.
+            terra_rewards_destination: Pubkey::new_unique(),
```
enriquefynn, line 103: 👀

tests/test_anker.py
```diff
@@ -0,0 +1,156 @@
...
+    '1',
+    '--owners',
+    ','.join(t.pubkey for t in test_addrs),
+)
+multisig_instance = multisig_data['multisig_address']
+multisig_pda = multisig_data['multisig_program_derived_address']
+print(f'> Created instance at {multisig_instance}.')
+
+
+def approve_and_execute(
```
enriquefynn, line 79: Can we use `approve_and_execute` from `prepare_test_environment`?
This function seems to only execute the tx, not doing the approve part, but it seems not to be called anywhere.

enriquefynn (commented):
Looks great with just some minor questions

tests/test_anker.py
ruuda: It does not need to be approved in this case, because I set up the multisig with a threshold of 1 for these tests, and the proposer automatically approves. But yeah I think we can reuse `prepare_test_environment` instead :+1:

ruuda: Ah wait it’s a bit more subtle than this, but we can make it a higher-order function.

enriquefynn:
There are still some issues with the python tests.

PR #469, opened 2021-12-01
[Anker] Remove ordering restriction for swap tokens.

### LINKED ISSUE #468: [Anker] Token's swap pool order should not matter
Currently, token a and b order matters for the swap (performed by the `SellRewards`).
This shouldn't be the case, since we can inspect the token's minter and decide which one is the UST/stSOL.

enriquefynn:
The stake pool tokens obeyed a strict ordering,
i.e. `token_a` and `token_b` were tied to stSOL and UST. This order is not necessary and removed.

Test swapping tokens with a stake pool created with swapped `token_a` and `token_b`.

Closes #468

### DIFF
--- a/anker/src/state.rs
+++ b/anker/src/state.rs
@@ -313,14 +313,32 @@ impl Anker {
         )?;
 
         // Pool stSOL and UST token could be swapped.
-        let (pool_st_sol_account, pool_ust_account) =
+        let (pool_st_sol_account, pool_st_sol_mint, pool_ust_account, pool_ust_mint) =
             if &token_swap.token_a == accounts.pool_st_sol_account.key {
-                (token_swap.token_a, token_swap.token_b)
+                Ok((
+                    token_swap.token_a,
+                    token_swap.token_a_mint,
+                    token_swap.token_b,
+                    token_swap.token_b_mint,
+                ))
+            } else if &token_swap.token_a == accounts.pool_ust_account.key {
+                Ok((
+                    token_swap.token_b,
+                    token_swap.token_b_mint,
+                    token_swap.token_a,
+                    token_swap.token_a_mint,
+                ))
             } else {
-                (token_swap.token_b, token_swap.token_a)
-            };
+                msg!(
+                    "Could not find a match for token swap account {}, candidates were the StSol account {} or UST account {}",
+                    token_swap.token_a,
+                    accounts.pool_st_sol_account.key,
+                    accounts.pool_ust_account.key
+                );
+                Err(AnkerError::WrongSplTokenSwapParameters)
+            }?;
 
-        // check stSOL token.
+        // Check stSOL token.
         if &pool_st_sol_account != accounts.pool_st_sol_account.key {
             msg!(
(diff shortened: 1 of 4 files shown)


anker/src/state.rs
```diff
@@ -339,20 +347,29 @@ impl Anker {
         );
             return Err(AnkerError::WrongSplTokenSwapParameters.into());
         }
+
+        // Get mint from tokens, depending on the order.
+        let (pool_st_sol_mint, pool_ust_mint) =
+            if &token_swap.token_a_mint == accounts.st_sol_mint.key {
```
ruuda, line 353: We should merge this with the earlier check, and resolve which one is A and which one is B only once. Right now what you can do is pass in the UST mint for the `st_sol_mint` argument, and it would not cause a failure.

anker/tests/tests/sell_rewards.rs
```diff
@@ -2,13 +2,38 @@ use anker::{error::AnkerError, token::MicroUst};
...
-    let mut context = Context::new_with_initialized_token_pool().await;
+    let mut context = Context::new_with_token_pool_rewards(Lamports(DEPOSIT_AMOUNT)).await;
+    context.sell_rewards().await;
+
+    let ust_balance = context.get_ust_balance(context.ust_reserve).await;
+    // Exchange rate is 12 stSol : 13 Sol
+    // We have 1 stSOL, our rewards were 1 - (1 * 12/13) = 0.076923077
+    // Initially there are 10 StSol and 10 UST in the AMM
+    // We should get 10 - (10*10 / 10.076923077) = 0.07633587793834806 UST
+    assert_eq!(ust_balance, MicroUst(76335877));
```
ruuda, line 20: This also has the off-by-1000 problem like #462.

anker/tests/tests/sell_rewards.rs
```diff
@@ -2,13 +2,38 @@ use anker::{error::AnkerError, token::MicroUst};
...
+// they are stSOL and UST), the order shouldn't make a difference.
+#[tokio::test]
+async fn test_successful_sell_rewards_pool_a_b_token_swapped() {
+    let mut context = Context::new().await;
+    // Swap the tokens a and b on Token Swap creation.
+    mem::swap(
+        &mut context.token_pool_context.token_a,
+        &mut context.token_pool_context.token_b,
+    );
+    context.initialize_token_pool().await;
```
ruuda, line 36: Rather than duplicating the full test, we can extract it into a function, and take the initialized context as an argument.

ruuda (commented):
Looks almost good, I have a suggestion.

anker/tests/tests/sell_rewards.rs
enriquefynn: Sorry about that, I think I was testing and overlook the copy-paste. Modified it a bit so it's nicer.

anker/src/state.rs
enriquefynn: Thanks! Modified it.

ruuda (approved):
LGTM!

PR #471, opened 2021-12-02
Split off a `solido anker` subcommand, add `anker show`

ruuda:
* Create a new `solido` subcommand, `solido anker`, where we can put the `create` and `show`. In the future we can add an Anker-specific `deposit` and `withdraw` here for testing, and the management commands.
* Add `anker show` to that, to display the details of the Anker instance.

To do:

* [ ] ~~Use this in the test script.~~ Will open a follow-up PR for this.
* [x] Print Terra rewards address.


ruuda:
@enriquefynn I rebased this on #473 and added a `TerraAddress` type, so we can accept the usual bech32 Terra addresses on the command line.

Still to do is actually printing the rewards address.

ruuda:
It almost works now, I still need to call `anker show` from the tests. Also, at the moment it prints

> Cannot exchange stSOL for SOL, because no stSTOL has been minted.

to stdout because of an `msg!` in the function ... maybe we can guard that to only print when compiling for BPF, so it doesn’t make us print invalid json output.

enriquefynn:
> It almost works now, I still need to call `anker show` from the tests. Also, at the moment it prints
> 
> > Cannot exchange stSOL for SOL, because no stSTOL has been minted.
> 
> to stdout because of an `msg!` in the function ... maybe we can guard that to only print when compiling for BPF, so it doesn’t make us print invalid json output.

This is for a deposits test?
Also now it's failing on the multisig test, do you know why? In my local machine it works just fine.

ruuda:
(Merging as @enriquefynn LGTM'd offline, but feel free to still leave comments anyway, and I’ll address them in a follow-up PR.)

PR #472, opened 2021-12-02
[Anker] Manager functions

enriquefynn:
Manager functions to change the Token Swap pool instance and the Terra rewards destination.
Test both functions

### DIFF
--- a/anker/src/processor.rs
+++ b/anker/src/processor.rs
@@ -1,5 +1,5 @@
 use borsh::BorshDeserialize;
-use lido::token::Lamports;
+use lido::{error::LidoError, token::Lamports};
 use solana_program::{
     account_info::AccountInfo,
     entrypoint::ProgramResult,
@@ -20,10 +20,11 @@ use crate::{
     find_instance_address, find_mint_authority, find_reserve_authority,
     find_st_sol_reserve_account,
     instruction::{
-        AnkerInstruction, DepositAccountsInfo, InitializeAccountsInfo, SellRewardsAccountsInfo,
-        WithdrawAccountsInfo,
+        AnkerInstruction, ChangeTerraRewardsDestinationAccountsInfo,
+        ChangeTokenSwapPoolAccountsInfo, DepositAccountsInfo, InitializeAccountsInfo,
+        SellRewardsAccountsInfo, WithdrawAccountsInfo,
     },
-    logic::{burn_b_sol, deserialize_anker, mint_b_sol_to},
+    logic::{burn_b_sol, deserialize_anker, get_token_swap_instance, mint_b_sol_to},
     state::Anker,
     token::BLamports,
 };
@@ -174,12 +175,13 @@ fn process_deposit(
         return Err(ProgramError::InvalidArgument);
     }
 
-    let (solido, anker) = deserialize_anker(
+    let (solido, anker) = deserialize_anker(program_id, accounts.anker, accounts.solido)?;
+    anker.check_st_sol_reserve_address(
         program_id,
-        accounts.anker,
-        accounts.solido,
+        accounts.anker.key,
         accounts.to_reserve_account,
     )?;
+    anker.check_is_st_sol_account(&solido, accounts.to_reserve_account)?;
 
     // Transfer `amount` StLamports to the reserve.
     invoke(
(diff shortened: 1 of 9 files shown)


anker/src/processor.rs
```diff
@@ -115,11 +116,6 @@ fn process_initialize(program_id: &Pubkey, accounts_raw: &[AccountInfo]) -> Prog
         accounts.ust_mint,
     )?;
 
-    let (_, token_swap_bump_seed) = Pubkey::find_program_address(
```
enriquefynn, line 118: Sneaked that here. In fact we don't really need this as it's stored in the token swap instance.

anker/tests/tests/manager.rs
```diff
@@ -0,0 +1,125 @@
...
+            new_token_pool.get_authority(),
+        )
+        .await;
+
+    new_token_pool
+        .initialize_token_pool(&mut context.solido_context)
+        .await;
+    let new_token_pool_address = new_token_pool.swap_account.pubkey();
+
+    let manager = Keypair::from_bytes(&context.solido_context.manager.to_bytes()).unwrap();
```
ruuda, line 39: Why jump through these hoops, we can just pass `&context.solido_context.manager` to `try_change_token_swap_pool` below?

anker/tests/tests/manager.rs
```diff
@@ -0,0 +1,125 @@
...
+    let result = context
+        .try_change_token_swap_pool(&manager, new_token_pool_address)
+        .await;
+    assert_solido_error!(result, AnkerError::WrongSplTokenSwapParameters);
+}
```
ruuda, line 79: :+1: Thanks for the extensive tests.

anker/tests/tests/manager.rs
```diff
@@ -0,0 +1,125 @@
...
+    let result = context
+        .try_change_terra_rewards_destination(&wrong_manager, new_terra_rewards_address)
+        .await;
+    assert_solido_error!(result, LidoError::InvalidManager);
+    let new_anker = context.get_anker().await;
+    assert_eq!(
+        anker.terra_rewards_destination,
+        new_anker.terra_rewards_destination
+    );
+}
```
ruuda, line 125: :+1: Thanks for checking the error cases too.

ruuda (commented):
:+1: Very nice, I like how the processor functions are very small and obvious.

Looks almost good to me, I think in a few places a comment would be helpful.

anker/src/processor.rs
```diff
@@ -340,6 +334,35 @@ fn process_withdraw(
...
+/// Solido's manager needs to sign the transaction.
+fn process_change_terra_rewards_destination(
+    program_id: &Pubkey,
+    accounts_raw: &[AccountInfo],
+) -> ProgramResult {
+    let accounts = ChangeTerraRewardsDestinationAccountsInfo::try_from_slice(accounts_raw)?;
+    let (solido, mut anker) = deserialize_anker(program_id, accounts.anker, accounts.solido)?;
+    solido.check_manager(accounts.manager)?;
+
+    anker.terra_rewards_destination = *accounts.terra_rewards_destination.key;
```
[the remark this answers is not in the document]
enriquefynn: This is done in #473

anker/src/state.rs
```diff
@@ -284,26 +356,10 @@ impl Anker {
         anker_program_id: &Pubkey,
         accounts: &SellRewardsAccountsInfo,
     ) -> ProgramResult {
+        // Check if the token swap account is the same one as the stored in the instance.
+        self.check_token_swap_pool(accounts.token_swap_pool)?;
+        let token_swap = get_token_swap_instance(accounts.token_swap_pool)?;
```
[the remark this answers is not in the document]
enriquefynn: That would prevent us from using the `get_token_swap_instance` to get, for instance, the `new_token_swap` when changing the instance.

anker/tests/tests/manager.rs
enriquefynn: We can't, cause we borrow context as mutable before.

ruuda: Hmm, right. What we could do instead is to not take the manager as an argument in `try_change_token_swap_pool`, but use `self.solido_context.manager` directly inside that function. Then if we want to simulate an unauthorized call with the wrong manager, we still can, we just need to overwrite `context.solido_context.manager` with a new pubkey from the test. I think that will be a bit cleaner to read.

anker/src/state.rs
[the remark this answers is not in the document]
ruuda: Hmm yeah. We could still do it, but then we need to shulffle some arguments around. Deserialize the old one first, then overwrite `self.token_swap_pool`, then do the check, and only then save. Then the check can take the old and new instance as arguments. But then you have this code where the check comes after the mutation, which works because it runs in a transaction, but it can still be a bit counter-intuitive. I’ll leave it up to you then.

ruuda (approved):
LGTM with one optional idea.

Issue #470, opened 2021-12-02
[Anchor] Store the Terra rewards destination address


ruuda:
We need to store the Terra address where we send the rewards in the UST. For now we opted just for a Solana `Pubkey` field, but Terra addresses are not the same as Solana ones.

Some resources I was able to find:

* https://docs.rs/cosmwasm-terra/0.7.2/cosmwasm_terra/types/struct.CanonicalAddr.html
* https://docs.cosmos.network/master/basics/accounts.html#keys-accounts-addresses-and-signatures

But in the end, what matters is the format that Wormhole accepts, so we should dive into that.

PR #473, opened 2021-12-04
[Anker] Send UST tokens via wormhole.

### LINKED ISSUE #445: [Anchor] ClaimRewards + SendRewards

enriquefynn:
Implement Wormhole instruction to send UST tokens to the Terra chain.
It was difficult to include the wormhole project. This rewrites some code from their repository.

~I don't understand the `nonce` and `fee` parameters. We should figure out if these are malleable before approving this PR.~
Fee should be set to 0: that's the fee payed in form of the transferred token in the target chain, to incentivize relayers.
Nonce is a random number, used to differentiate transactions with the same parameter, this is set by the caller.

We can leave implementing tests in a later PR, as this is already extensive.

Closes #445, #470
Review carefully 🕵️

### DIFF
--- a/anker/src/instruction.rs
+++ b/anker/src/instruction.rs
@@ -41,8 +41,7 @@ pub enum AnkerInstruction {
     SellRewards,
     /// Transfer from the UST reserve to terra through Wormhole.
     SendRewards {
-        nonce: u32, // FIXME: Check this nonce business.
-        fee: u64,   // FIXME: I don't know if this can be variable or not. Check it.
+        wormhole_nonce: u32, // Random number used to differentiate similar transaction.
     },
     /// Change the Anker's rewards destination address on Terra:
     /// `terra_rewards_destination`.
@@ -489,10 +488,9 @@ accounts_struct! {
 pub fn send_rewards(
     program_id: &Pubkey,
     accounts: &SendRewardsAccountsMeta,
-    nonce: u32,
-    fee: u64,
+    wormhole_nonce: u32,
 ) -> Instruction {
-    let data = AnkerInstruction::SendRewards { nonce, fee };
+    let data = AnkerInstruction::SendRewards { wormhole_nonce };
     Instruction {
         program_id: *program_id,
         accounts: accounts.to_vec(),
--- a/anker/src/instruction.rs
+++ b/anker/src/instruction.rs
@@ -406,11 +406,6 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
-        // So we can get the UST minter.
-        pub token_swap_pool {
-            is_signer: false,
-            is_writable: false,
-        },
         pub solido {
             is_signer: false,
             is_writable: false,
@@ -437,11 +432,11 @@ accounts_struct! {
             is_writable: false,
         },
         // From should be the UST token address.
-        pub from {
+        pub ust_reserve_account {
             is_signer: false,
             is_writable: true,
         },
-        pub mint {
+        pub ust_mint {
             is_signer: false,
             is_writable: true,
         },
--- a/anker/src/instruction.rs
+++ b/anker/src/instruction.rs
@@ -37,6 +37,11 @@ pub enum AnkerInstruction {
 
     /// Sell rewards to the UST reserve.
     SellRewards,
+    /// Transfer from the UST reserve to terra through Wormhole.
+    SendRewards {
+        nonce: u32, // FIXME: Check this nonce business.
+        fee: u64,   // FIXME: I don't know if this can be variable or not. Check it.
+    },
     /// Change the Anker's rewards destination address on Terra:
     /// `terra_rewards_destination`.
     ChangeTerraRewardsDestination,
@@ -70,6 +75,15 @@ accounts_struct! {
             is_writable: false,
         },
         pub solido_program {
+            is_signer: false,
+            is_writable: false,
+        },
+        // Store wormhole program ids
+        pub wormhole_core_bridge_program_id {
+            is_signer: false,
+            is_writable: false,
+        },
+        pub wormhole_token_bridge_program_id {
             is_signer: false,
             is_writable: false,
         },
@@ -375,6 +389,105 @@ pub fn change_token_swap_pool(
     accounts: &ChangeTokenSwapPoolAccountsMeta,
 ) -> Instruction {
     let data = AnkerInstruction::ChangeTokenSwapPool;
+    Instruction {
+        program_id: *program_id,
+        accounts: accounts.to_vec(),
+        data: data.to_vec(),
+    }
+}
+
+accounts_struct! {
+    SendRewardsAccountsMeta, SendRewardsAccountsInfo {
+        // Needs to be writtable in order to save new Token Pool address.
+        pub anker {
+            is_signer: false,
+            is_writable: true,
+        },
+        // So we can get the UST minter.
+        pub token_swap_pool {
+            is_signer: false,
+            is_writable: false,
+        },
+        pub solido {
+            is_signer: false,
+            is_writable: false,
+        },
+        pub reserve_authority {
+            is_signer: false,
+            is_writable: false,
+        },
+        // Accounts for Wormhole swap.
+        pub wormhole_token_bridge_program_id {
+            is_signer: false,
+            is_writable: false,
+        },
+        pub wormhole_core_bridge_program_id {
+            is_signer: false,
+            is_writable: false,
+        },
+        pub payer {
+            is_signer: true,
+            is_writable: true,
+        },
+        pub config_key {
+            is_signer: false,
+            is_writable: false,
+        },
+        // From should be the UST token address.
+        pub from {
+            is_signer: false,
+            is_writable: true,
+        },
+        pub mint {
+            is_signer: false,
+            is_writable: true,
+        },
+        pub custody_key {
+            is_signer: false,
+            is_writable: true,
+        },
+        pub authority_signer_key {
+            is_signer: false,
+            is_writable: false,
+        },
+        pub custody_signer_key {
+            is_signer: false,
+            is_writable: false,
+        },
+        pub bridge_config {
+            is_signer: false,
+            is_writable: true,
+        },
+        pub message {
+            is_signer: true,
+            is_writable: true,
+        },
+        pub emitter_key {
+            is_signer: false,
+            is_writable: false,
+        },
+        pub sequence_key {
+            is_signer: false,
+            is_writable: true,
+        },
+        pub fee_collector_key {
+            is_signer: false,
+            is_writable: true,
+        },
+        const sysvar_clock = sysvar::clock::id(),
+        const sysvar_rent = sysvar::rent::id(),
+        const system_program = system_program::id(),
+        const spl_token = spl_token::id(),
+    }
+}
+
+pub fn send_rewards(
+    program_id: &Pubkey,
+    accounts: &SendRewardsAccountsMeta,
+    nonce: u32,
+    fee: u64,
+) -> Instruction {
+    let data = AnkerInstruction::SendRewards { nonce, fee };
     Instruction {
         program_id: *program_id,
         accounts: accounts.to_vec(),
--- a/anker/src/instruction.rs
+++ b/anker/src/instruction.rs
@@ -402,10 +402,9 @@ pub fn change_token_swap_pool(
 
 accounts_struct! {
     SendRewardsAccountsMeta, SendRewardsAccountsInfo {
-        // Needs to be writtable in order to save new Token Pool address.
         pub anker {
             is_signer: false,
-            is_writable: true,
+            is_writable: false,
         },
         // So we can get the UST minter.
         pub token_swap_pool {
--- a/anker/src/instruction.rs
+++ b/anker/src/instruction.rs
@@ -431,7 +431,6 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
-        // From should be the UST token address.
         pub ust_reserve_account {
             is_signer: false,
             is_writable: true,
--- a/anker/src/instruction.rs
+++ b/anker/src/instruction.rs
@@ -11,12 +11,14 @@ use solana_program::{
     system_program, sysvar,
 };
 
-use crate::token::BLamports;
+use crate::{token::BLamports, wormhole::ForeignAddress};
 
 #[repr(C)]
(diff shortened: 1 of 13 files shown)


anker/src/processor.rs
```diff
@@ -369,6 +374,59 @@ fn process_change_token_swap_pool(
...
+            accounts.message.clone(),
+            accounts.emitter_key.clone(),
+            accounts.sequence_key.clone(),
+            accounts.fee_collector_key.clone(),
+            accounts.sysvar_clock.clone(),
+            accounts.sysvar_rent.clone(),
+            accounts.system_program.clone(),
+            accounts.wormhole_core_bridge_program_id.clone(),
+            accounts.spl_token.clone(),
+            accounts.wormhole_token_bridge_program_id.clone(),
```
enriquefynn, line 430: We can't make so many clones. Wonder why?! Maybe we reached some limits, and the compiler is trying to be clever.
The BPF program is all screwed up if we include this code

ruuda: What do you mean “The BPF program is all screwed up”? I don’t expect the compiler to behave any differently depending on the amount of clones.

enriquefynn: There are all sort of errors in the tests, e.g., I put some `msg!` to debug and the errors change. Access violations, etc.
If I put 1 account less in this function, the errors don't happen anymore.
Maybe we don't need the last account `accounts.wormhole_token_bridge_program_id.clone(),` and we can get away with it.

ruuda: I suspect something funky going on with `solana_program_test` or something like that more than rustc miscompiling anything. Maybe a size limit of some sort somewhere? Also, we had issues in the past with inexplicable errors where a clean build fixed things.

> Maybe we don't need the last account accounts.wormhole_token_bridge_program_id.clone(), and we can get away with it.

It sounds very dangerous to live that close to the edge, especially if we don’t know why the edge is there. We should understand this a bit better before we proceed. Maybe it’s possible to make a minimal example? What if we make an otherwise empty program that takes as many accounts as this instruction, and that tries to call Wormhole? Maybe we can isolate the issue a bit?

anker/src/wormhole.rs
```diff
@@ -0,0 +1,213 @@
...
+        Pubkey::from_str("SysvarRent111111111111111111111111111111111").unwrap(),
+        Pubkey::from_str("11111111111111111111111111111111").unwrap(),
+        Pubkey::from_str("worm2ZoG2kUd4vFXhvjh93UUH596ayRfgQ2MgjNMTth").unwrap(),
+        Pubkey::from_str("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA").unwrap(),
+    ];
+    let expected_data = hex::decode("058c3800000065cd1d00000000000000000000000000000000000000000000000029fc5aacd613410b68c9c08d4e1656e3c890e4820200").unwrap();
+    let accounts: Vec<Pubkey> = instruction.accounts.iter().map(|acc| acc.pubkey).collect();
+    assert_eq!(expected_accounts, accounts);
+    assert_eq!(expected_data, instruction.data);
+}
```
ruuda, line 213: :+1: Thanks for adding a test!

ruuda (commented):
Looks mostly good! I have a few suggestions.

anker/src/instruction.rs
```diff
@@ -37,9 +39,16 @@ pub enum AnkerInstruction {
 
     /// Sell rewards to the UST reserve.
     SellRewards,
+    /// Transfer from the UST reserve to terra through Wormhole.
+    SendRewards {
+        nonce: u32, // FIXME: Check this nonce business.
+        fee: u64,   // FIXME: I don't know if this can be variable or not. Check it.
```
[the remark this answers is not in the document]
enriquefynn: =I asked Dave and he answered:
> Nonce is random
So it's fine to leave it caller's control.
> Fee is a portion of the tokens that will pay the relayer, e.g. if the sender doesn't have any tokens in the chain that it's sending to.
So we should definitely set this to 0, to avoid getting sniped by a malicious user that can act as a relayer and steal all the tokens 😅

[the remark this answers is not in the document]
ruuda: > So we should definitely set this to 0, to avoid getting sniped by a malicious user that can act as a relayer and steal all the tokens sweat_smile

Ah yes :fearful: we should hard-code it to 0 and not make it a parameter!

anker/src/instruction.rs
```diff
@@ -381,3 +400,102 @@ pub fn change_token_swap_pool(
         data: data.to_vec(),
     }
 }
+
+accounts_struct! {
+    SendRewardsAccountsMeta, SendRewardsAccountsInfo {
+        // Needs to be writtable in order to save new Token Pool address.
```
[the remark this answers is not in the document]
enriquefynn: Right! That was copy-paste mistake, thanks!

anker/src/instruction.rs
```diff
@@ -381,3 +400,102 @@ pub fn change_token_swap_pool(
...
+        },
+        pub custody_signer_key {
+            is_signer: false,
+            is_writable: false,
+        },
+        pub bridge_config {
+            is_signer: false,
+            is_writable: true,
+        },
+        pub message {
```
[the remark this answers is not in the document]
enriquefynn: It's where the wormhole program is going to do its magic to save things. We should investigate how many bytes should be allocated there.

anker/src/instruction.rs
```diff
@@ -381,3 +400,102 @@ pub fn change_token_swap_pool(
...
+
+accounts_struct! {
+    SendRewardsAccountsMeta, SendRewardsAccountsInfo {
+        // Needs to be writtable in order to save new Token Pool address.
+        pub anker {
+            is_signer: false,
+            is_writable: true,
+        },
+        // So we can get the UST minter.
+        pub token_swap_pool {
```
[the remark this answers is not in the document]
enriquefynn: That's true, I modified it

anker/src/instruction.rs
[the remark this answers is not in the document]
enriquefynn: Good catch! Thanks

anker/src/instruction.rs
```diff
@@ -381,3 +400,102 @@ pub fn change_token_swap_pool(
...
+        pub payer {
+            is_signer: true,
+            is_writable: true,
+        },
+        pub config_key {
+            is_signer: false,
+            is_writable: false,
+        },
+        // From should be the UST token address.
+        pub from {
```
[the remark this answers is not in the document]
enriquefynn: Done!

anker/src/instruction.rs
```diff
@@ -381,3 +400,102 @@ pub fn change_token_swap_pool(
...
+            is_signer: false,
+            is_writable: true,
+        },
+        pub mint {
```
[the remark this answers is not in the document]
enriquefynn: Done!

anker/src/instruction.rs
```diff
@@ -381,3 +400,102 @@ pub fn change_token_swap_pool(
...
+            is_signer: false,
+            is_writable: true,
+        },
+        pub custody_key {
+            is_signer: false,
+            is_writable: true,
+        },
+        pub authority_signer_key {
```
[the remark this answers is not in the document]
enriquefynn: No idea, this is a derived account needed by Wormhole 😬

anker/src/wormhole.rs
```diff
@@ -0,0 +1,213 @@
...
+
+/// Wormhole's Terra chain id.
+pub const WORMHOLE_CHAIN_ID_TERRA: u16 = 3;
+
+/// Position of the native transfer code at the Wormhole project:
+/// https://github.com/certusone/wormhole
+/// solana/modules/token_bridge/program/src/lib.rs
+const WORMHOLE_NATIVE_TRANSFER_CODE: u8 = 5;
+
+pub type ForeignAddress = [u8; 32];
```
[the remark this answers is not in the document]
enriquefynn: Great, I modified it, implementing the `FromString` trait. This helps with the client initialization as well.

anker/src/wormhole.rs
```diff
@@ -0,0 +1,213 @@
...
+pub struct Payload {
+    pub nonce: u32,
+    pub amount: MicroUst,
+    pub fee: u64,
+    pub foreign_address: ForeignAddress,
+    pub target_chain: u16,
+}
+
+impl Payload {
+    pub fn new(nonce: u32, amount: MicroUst, fee: u64, foreign_address: ForeignAddress) -> Payload {
```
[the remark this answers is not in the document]
enriquefynn: That's solved now, we set the fee to 0 always.

anker/src/instruction.rs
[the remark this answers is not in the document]
ruuda: Do we need to allocate it ourselves? How can we figure this out? Also, why does it store anything in an account at all? Couldn’t it store whatever it needs in the transaction?

anker/src/instruction.rs
```diff
@@ -381,3 +399,95 @@ pub fn change_token_swap_pool(
...
+        },
+        pub payer {
+            is_signer: true,
+            is_writable: true,
+        },
+        pub config_key {
+            is_signer: false,
+            is_writable: false,
+        },
+        // From should be the UST token address.
```
ruuda, line 434: This comment makes little sense now after renaming the account. Maybe say something like

> The `ust_reserve_account` is the account that we send the funds from for the wormhole transfer.

anker/src/instruction.rs
```diff
@@ -381,3 +399,95 @@ pub fn change_token_swap_pool(
...
+        pub ust_reserve_account {
+            is_signer: false,
+            is_writable: true,
+        },
+        pub ust_mint {
+            is_signer: false,
+            is_writable: true,
```
ruuda, line 441: Why does the mint need to be writable? Does the SPL token program require this for transfers? I would expect it to only be writable for mint/burn.

anker/src/wormhole.rs
```diff
@@ -0,0 +1,283 @@
...
+impl FromStr for ForeignAddress {
+    type Err = FromHexError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut foreign_address = [0; 32];
+        // Allow for array starting with 0x
+        let array_vec = if s.len() >= 2 && &s[..2] == "0x" {
+            hex::decode(&s[2..])
+        } else {
+            hex::decode(s)
+        }?;
```
ruuda, line 36: We can make this a bit nicer with [`strip_prefix`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix):

```rust
match s.strip_prefix("0x") {
    Some(suffix) => hex::decode(suffix),
    None => hex::decode(s),
}?;
```
or even
```rust
hex::decode(s.strip_prefix("0x").unwrap_or(s))?;
```

anker/src/wormhole.rs
```diff
@@ -0,0 +1,283 @@
...
+        if array_vec.len() > 32 {
+            return Err(FromHexError::InvalidStringLength);
+        }
+        let start = 32 - array_vec.len();
+        foreign_address[start..].clone_from_slice(&array_vec[..(32 - start)]);
```
ruuda, line 41: The slicing on the `array_vec` is redundant here: `[..32 - start] = [..32 - (32 - array_vec.len())] = [..array_vec.len()] = [..]`.

anker/src/wormhole.rs
```diff
@@ -0,0 +1,283 @@
...
+    assert_eq!(zero_addr, zero_addr_start_0x);
+
+    let eth_addr =
+        ForeignAddress::from_str("29fc5aacd613410b68c9c08d4e1656e3c890e4820200").unwrap();
+    assert_eq!(
+        eth_addr,
+        ForeignAddress([
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 41, 252, 90, 172, 214, 19, 65, 11, 104, 201, 192, 141,
+            78, 22, 86, 227, 200, 144, 228, 130, 2, 0
+        ])
```
ruuda, line 249: :+1: Thanks for testing this case.

ruuda (commented):
I have a few more comments but aside from those this looks good!

anker/src/wormhole.rs
ruuda: Actually, don’t bother with this, I think we shouldn’t convert to/from hexadecimal, we should use Bech32 which is used by Terra. (No need to do that in this PR, it’s already big enough; we can change this later.)

anker/src/instruction.rs
enriquefynn: It's needed for Wormhole here: https://github.com/certusone/wormhole/blob/57919f6675ed5e0123d83f6abed09f06e0390f09/solana/modules/token_bridge/program/src/instructions.rs#L371
Maybe it calls something that need this?!

anker/src/instruction.rs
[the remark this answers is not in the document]
enriquefynn: >Do we need to allocate it ourselves? How can we figure this out?

I believe so. I guess we need to allocate this structure. We should take a look at it.

>why does it store anything in an account at all? Couldn’t it store whatever it needs in the transaction?

I guess it's the same reason we have to keep metrics. It's complicated and cumbersome to rely on re-executing transactions.

anker/src/wormhole.rs
enriquefynn: Done in #471

anker/src/wormhole.rs
enriquefynn: I see there's something in #471 already. Will leave it.

anker/src/instruction.rs
ruuda: Looks like an oversight on their end, they only read from it and call `spl_token::initialize`, which doesn’t need the mint to be writable either: https://github.com/certusone/wormhole/blob/d678cb4662f5eb9b9590c0c9e9238d137b5e4226/solana/modules/token_bridge/program/src/api/transfer.rs#L142-L172

ruuda: I opened https://github.com/certusone/wormhole/issues/607 for this.

ruuda (approved):
LGTM

Issue #474, opened 2021-12-08
Fix test coverage CI step


ruuda:
We have a CI step that builds programs with coverage collection enabled and then generates a coverage report.

Unfortunately, it started failing a while ago due to some external breaking change. I spent some time trying to fix it (#461) but so far I was not successful, and right now we have more urgent things to work on.

For now I’ll disable the CI step, and I’m opening this issue as a reminder to fix it later.

PR #475, opened 2021-12-08
Disable coverage CI step for now

ruuda:
This step has been failing for a while due to some breaking change somewhere in the toolchain, not sure where exactly. I spent some time trying to fix it, but so far I did not manage to. We should fix this, but in the meantime, I think it is counterproductive to have CI always fail, because you can't tell when the situation got worse. So for now, disable this CI step.

To restore the step later, we can `git revert` the commit.

I opened #474 so we remember to re-enable the step.


enriquefynn (approved):
LGTM

PR #476, opened 2021-12-08
[Anker] Add metrics

enriquefynn:
Log the total number of StSOL and UST sold by the `SellRewards`
function.

Add some logging and tests.

### DIFF
--- a/anker/src/metrics.rs
+++ b/anker/src/metrics.rs
@@ -13,7 +13,7 @@ pub struct Metrics {
     #[serde(rename = "swaped_rewards_total_st_lamports")]
     pub swapped_rewards_st_sol_total: StLamports,
 
-    /// Total swapped amount of UST to StSOL.
+    /// Total amount of UST received through swaps.
     #[serde(rename = "swapped_rewards_ust_total_microust")]
     pub swapped_rewards_ust_total: MicroUst,
 }
--- a/anker/src/metrics.rs
+++ b/anker/src/metrics.rs
@@ -1,0 +1,39 @@
+use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
+use lido::token::StLamports;
+use serde::Serialize;
+
+use crate::token::{self, MicroUst};
+
+#[repr(C)]
+#[derive(
+    Clone, Debug, Default, BorshDeserialize, BorshSerialize, BorshSchema, Eq, PartialEq, Serialize,
+)]
+pub struct Metrics {
+    /// Total swapped amount of StSOL to UST.
+    #[serde(rename = "swaped_rewards_total_st_lamports")]
+    pub swapped_rewards_st_sol_total: StLamports,
+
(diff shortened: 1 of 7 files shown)


anker/src/metrics.rs
```diff
@@ -0,0 +1,39 @@
...
+#[repr(C)]
+#[derive(
+    Clone, Debug, Default, BorshDeserialize, BorshSerialize, BorshSchema, Eq, PartialEq, Serialize,
+)]
+pub struct Metrics {
+    /// Total swapped amount of StSOL to UST.
+    #[serde(rename = "swaped_rewards_total_st_lamports")]
+    pub swapped_rewards_st_sol_total: StLamports,
+
+    /// Total swapped amount of UST to StSOL.
```
ruuda, line 16: Nit: not really UST swapped for stSOL, right? Maybe write “Total amount of UST received through swaps”?

anker/src/metrics.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[serde(rename = "swapped_rewards_ust_total_microust")]
+    pub swapped_rewards_ust_total: MicroUst,
+}
```
ruuda, line 19: :+1: Great idea to have stSOL and UST totals, if we sample that over time then we can compute the average exchange rate we got over any time window we want, neat.

anker/src/processor.rs
```diff
@@ -236,17 +238,27 @@ fn process_sell_rewards(program_id: &Pubkey, accounts_raw: &[AccountInfo]) -> Pr
...
-    swap_rewards(program_id, rewards, &anker, &accounts)
+
+    // Get the amount of UST that we had.
+    let ust_before = MicroUst(Anker::get_token_amount(accounts.ust_reserve_account)?);
+    swap_rewards(program_id, rewards, &anker, &accounts)?;
+    // Get new UST amount.
+    let ust_after = MicroUst(Anker::get_token_amount(accounts.ust_reserve_account)?);
+    let swapped_ust = (ust_after - ust_before)?;
+
+    msg!("Swapped {} for {}.", st_sol_amount, swapped_ust);
```
ruuda, line 256: Nice to also print what happened. I think in hindsight one of the things I run into with the Solido program is that it is hard to see on block explorers what kind of transaction something was (aside from deposit/withdraw where we also print). Printing some hint of what we are doing fixes that. :+1:

anker/src/state.rs
```diff
@@ -566,6 +570,13 @@ impl Anker {
         )?;
         Ok(Box::new(wormhole_transfer_args))
     }
+
+    /// Get the `amount` of tokens from the SPL account defined by `account`.
+    /// Does not perform any checks, fails if not able to decode an SPL account.
```
ruuda, line 575: We could at least check that the account is owned by the SPL token program before we deserialize it.

ruuda (approved):
I found one misleading doc comment, LGTM aside from that!

PR #478, opened 2021-12-10
added solido.js library to interact with solido program

hritique:


js/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Run TS watch script
+```
+$ npm run watch:ts
+```
+
+### Run tests watch script
+```
+$ npm run watch:test
+```
```
ruuda, line 36: What’s the difference between `watch:ts` and `watch:test`? One just typechecks the TypeScript and the other one also runs the tests?

js/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Build and Release
+TODO
+
+
+## Usage
+
+### Javascript
+```js
+const SolidoJS = require("@chorusone/solido.js");
```
ruuda, line 46: Is it common to use the `JS` suffix even in names in Javascript?

js/README.md
```diff
@@ -0,0 +1,73 @@
...
+const snapshot = SolidoJS.getSnapshot(...params);
```
ruuda, line 47: We should give a full example that works. I think it needs a connection and also the `SolidoJS.MAINNET_PROGRAM_ADDRESSES`, right?

js/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@chorusone/solido.js",
+  "version": "1.0.0",
+  "description": "Typescript SDK for interacting with Solido Program",
```
ruuda, line 4: Shall we say “for interacting with Lido for Solana (Solido)” to make it more discoverable?

js/src/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+  solidoProgramId: new PublicKey(
+    'CrX7kMhLC3cSsXJdT7JDgqrRVWGnUpX3gfEfxxU2NVLi'
+  ),
+  solidoInstanceId: new PublicKey(
+    '49Yi1TKkNyYjPAFdR9LBvoHcUjuPX4Df5T5yv39w2XTn'
+  ),
+  stSolMintAddress: new PublicKey(
+    '7dHbWXmci3dT8UFYWYZweBLXgycu7Y3iL6trKn1Y7ARj'
+  ),
+};
```
ruuda, line 14: :+1: I cross-checked these addresses against https://docs.solana.lido.fi/deployments and I can confirm that they are correct.

ruuda (commented):
I looked at the first few files but I haven’t gotten to the meat of it yet. Is there a tool that I can run to generate API reference docs? (Something like what `rustdoc` is for Rust?) I see https://solana-labs.github.io/solana-web3.js/ is generated by Typedoc, is that the one that the standard one? Do we need to configure anything to set it up?

js/src/utils.ts
```diff
@@ -0,0 +1,56 @@
+import { PublicKey } from '@solana/web3.js';
+import type BN from 'bn.js';
+import type { ProgramAddresses, Snapshot } from './types';
+
+export const findAuthorityProgramAddress = async (
+  programAddresses: ProgramAddresses,
+  additionalSeedString: string
```
AlexBHarley, line 7: Can this be more strictly typed? AFAICT the only relevant strings are `reserve_account` and `mint_authority`. Are `rewards_withdraw_authority` and `sol_reserve` also valid?

AlexBHarley:
Looks great. A couple questions,

1. Is it worth adding a guard to the `createAssociatedTokenAccount` call if the user already has one?
2. As the Lido validator set grows, how costly will pulling the snapshot get? Moreover, is there any need to pull the whole snapshot when doing a `withdraw`?

js/src/utils.ts
hritique: Yeah good idea, we can add all the available options to it like "reserve_account" | "mint_authority" ...
To help the user(editor suggestions) and to make it more strictly typed

js/package.json
hritique: Sure we should do this

js/README.md
hritique: Not really, but I just used it in the example here

js/README.md
hritique: `watch:ts` compile the ts files on every file change
`watch:test` runs the test on every file change

ruuda:
> Is it worth adding a guard to the `createAssociatedTokenAccount` call if the user already has one?

There is a separation between constructing instructions and transactions (which are lists of instructions), and querying the on-chain state. I think it’s nicer to keep the functions that build instructions pure, without side effects. If all instruction builders query the chain state themselves, this can result in torn reads, and observing inconsistent states, because they all query the chain at different times, and this makes it difficult to compose different instructions into one transaction.

We should make a convenience function though that returns all instructions that you need to do a deposit, and that can then include the instruction to create the ATA if it doesn’t exist yet.

> As the Lido validator set grows, how costly will pulling the snapshot get?

It depends on the RPC node configuration how many accounts you can query in a single `GetMultipleAccounts` call. The default configuration is to allow fetching 100 accounts in one call, so we can still grow the validator set by more than 5× before we run into issues there. Also, we really only need to fetch the validator stake accounts to be able to determine the maximum amount to withdraw; we could also set a lower limit that is on the safe side, and then we only need to fetch the single account that holds the Solido state.

> Moreover, is there any need to pull the whole snapshot when doing a withdraw?

To be able to determine which validator to withdraw from, we only need the Solido state. But to be able to determine the maximum amount to withdraw, we also need the stake account. However, if we first fetch the Solido state, and _then_ fetch the stake account, you run into the torn reads again, and then we build transactions based on an inconsistent view of the chain. Right now the only way to get a consistent view through the Solana RPC, is to use `GetMultipleAccounts`, and then we have to get all of the accounts that we might possibly need.

I don’t think it will be a big issue though; we should be able to get all the data we need with just two RPC calls, and that will serve us up to ~100 validators, and there is a lot we can optimize at that point to scale further if we need to. (For example, only fetch the top _n_ stake accounts, because we don’t expect them to change quickly anyway. Or don’t look at the stake accounts at all and just set a slightly lower per-instruction withdraw limit.)

ruuda:
@hritique I ran `yarn` in the `js` directory and it produced a `yarn.lock` that is not currently tracked by Git. Shouldn’t it be under source control so we all use the same package versions?

js/src/instructions/utils.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  ASSOCIATED_TOKEN_PROGRAM_ID,
+  Token,
+  TOKEN_PROGRAM_ID,
+} from '@solana/spl-token';
+import { PublicKey, TransactionInstruction } from '@solana/web3.js';
+
+export const getATAInitializeInstruction = async (
```
ruuda, line 8: Nit, I prefer to capitalize abbreviations like regular words because I find `JsonRpcXmlHttpRequestBackend` more readable than `JSONRPCXMLHTTPRequestBackend`. Also, it may not be obvious what an “ATA” is, although
`getInitializeAssociatedTokenAccountInstruction` is quite long ... I have no strong opinion on this, but I think I would just write it out fully. If you are new to the Solana ecosystem and you read code that calls `getAtaInitializeInstruction` you have no clue what it does and it may be difficult to find. If it says “associated token account”, then you at least know that it’s a token account, and searching for “associated token account” is more obvious than just “ATA”.

js/src/instructions/withdraw.ts
```diff
@@ -0,0 +1,102 @@
...
+import BN from 'bn.js';
+import type { Lamports, ProgramAddresses, Snapshot } from '../types';
+import {
+  findAuthorityProgramAddress,
+  getHeaviestValidatorStakeAccount,
+} from '../utils';
+
+export const getWithdrawInstruction = async (
+  snapshot: Snapshot,
+  programAddresses: ProgramAddresses,
```
ruuda, line 19: Let’s make `ProgramAddresses` a field of `Snapshot` and initialize it when we create the snapshot. That’s one less thing for callers to worry about, and also makes it impossible to mix a snapshot created from say testnet `ProgramAddresses` with mainnet `ProgramAddresses` here.

js/src/instructions/withdraw.ts
```diff
@@ -0,0 +1,102 @@
...
+  stSolAccountOwnerAddress: PublicKey,
```
ruuda, line 20: Nit: This is the owner of `senderStSolAccountAddress`, right? Can we call it`senderStSolAccountOwnerAddress` for consistency and to clarify?

js/src/instructions/withdraw.ts
```diff
@@ -0,0 +1,102 @@
...
+) => {
+  const stakeAuthorityAddress = await findAuthorityProgramAddress(
+    programAddresses,
+    'stake_authority'
+  );
+  const heaviestValidatorStakeAccount =
+    getHeaviestValidatorStakeAccount(snapshot);
+
+  if (amount.lamports.lte(snapshot.stakeAccountRentExemptionBalance.lamports)) {
+    throw new Error('Amount must be greater than the rent exemption balance');
```
ruuda, line 33: Where does `Error` come from? Is it something that we define, or is it a general TypeScript error type? If it’s the latter, does it make sense to define our own error type so callers can catch just these errors without accidentally swallowing problems that happen elsewhere?

We should also document in the doc comment that the function may throw this type.

js/src/instructions/withdraw.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  const maxWithdrawAmount = heaviestValidatorStakeAccount.balance.lamports
+    .div(new BN(10))
+    .add(new BN(10));
```
ruuda, line 38: The max withdraw amount is 1/10 of the balance + 10 **SOL**, not 10 Lamports! This should be `10_000_000_000`. (Does TypeScript have numeric underscores? Or is there a `LAMPORTS_PER_SOL` constant somewhere?)

js/src/instructions/withdraw.ts
```diff
@@ -0,0 +1,102 @@
...
+    'stake_authority'
+  );
+  const heaviestValidatorStakeAccount =
+    getHeaviestValidatorStakeAccount(snapshot);
+
+  if (amount.lamports.lte(snapshot.stakeAccountRentExemptionBalance.lamports)) {
+    throw new Error('Amount must be greater than the rent exemption balance');
+  }
+
+  const maxWithdrawAmount = heaviestValidatorStakeAccount.balance.lamports
```
ruuda, line 36: To clarify that the amount is in SOL, we can wrap the expression on the right-hand side in `Lamports()`. These wrapper types are not just useful for users, also for us!

js/src/instructions/withdraw.ts
```diff
@@ -0,0 +1,102 @@
...
+  getHeaviestValidatorStakeAccount,
+} from '../utils';
+
+export const getWithdrawInstruction = async (
+  snapshot: Snapshot,
+  programAddresses: ProgramAddresses,
+  stSolAccountOwnerAddress: PublicKey,
+  senderStSolAccountAddress: PublicKey,
+  recipientStakeAccountAddress: PublicKey,
+  amount: Lamports
```
ruuda, line 23: Amount is in `StLamports`, not in `Lamports`!

js/src/instructions/withdraw.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  if (amount.lamports.lte(snapshot.stakeAccountRentExemptionBalance.lamports)) {
+    throw new Error('Amount must be greater than the rent exemption balance');
+  }
+
+  const maxWithdrawAmount = heaviestValidatorStakeAccount.balance.lamports
+    .div(new BN(10))
+    .add(new BN(10));
+
+  if (amount.lamports.gte(maxWithdrawAmount)) {
```
ruuda, line 40: This is mixing things in different units (and that is where the different wrapper types come in, to prevent this). Is it possible to overload the `<=` operator in TypeScript? Then we can make sure that for `Lamports` on the left-hand side, the right-hand side must also be `Lamports`, and the compiler will reject `amount < maxWithdrawAmount` with a type error, because `amount: StLamports` but `maxWithdrawAmount: Lamports`. This way we can leverage the type system to prevent bugs :-)

To fix the bug, we need to convert the stSOL amount to SOL, according to the exchange rate in the snapshot.

ruuda (commented):
I reviewed a few more files, but I haven’t gotten through all of it yet.

js/src/instructions/withdraw.ts
hritique: Ahhh yes, that make sense. Will try to fix it

js/src/instructions/withdraw.ts
hritique: Yes, stumbled upon this today. Will fix this soon

js/src/instructions/withdraw.ts
hritique: Ohhh yes, thanks for pointing this. Would have missed this

js/src/instructions/withdraw.ts
hritique: Yes, we'll add a list of possible errors so that users can catch and handle them better

js/src/snapshot.ts
```diff
@@ -0,0 +1,453 @@
...
+  return deserialized;
+};
+
+export const getSnapshot = async (
+  connection: Connection,
+  solidoInstanceAccountInfo: AccountInfo<Buffer>,
+  reserveAccountInfo: AccountInfo<Buffer>,
+  programAddresses: ProgramAddresses
+): Promise<Snapshot> => {
+  const solido = getSolido(solidoInstanceAccountInfo.data);
```
ruuda, line 386: Now we still do a torn read, the `getMultipleAccounts` later on can read from a later slot than where we got the `solido` account from, and they may have changed in the meantime. To get a consistent view, we have to re-read the `solido` account as part of the `getMultipleAccounts`. One way to do this is in a loop:

* Start with just the Solido account, do a `getMultipleAccounts`, put the results in a dictionary keyed on accounts.
* Start parsing the Solido state and the other accounts, but instead of calling the RPC, read from the dictionary.
* If an account is not present in the dictionary, add that address to the list of accounts to retrieve, and start again. To be more efficient, we can add all validator stake accounts to the list immediately. Then we should only have to do two reads (the first one to get the Solido instance, and the second one that gets everything we need). In the extremely unlucky case where you run this just as we add a validator, it will have to do three iterations.

js/src/snapshot.ts
```diff
@@ -0,0 +1,453 @@
...
+
+  const reserveAccountBalance: Lamports = {
+    lamports: new BN(
+      reserveAccountInfo.lamports - reserveAccountRentExemptionBalance
+    ),
+  };
+
+  const {
+    value: { amount },
+  } = await connection.getTokenSupply(programAddresses.stSolMintAddress);
```
ruuda, line 401: We can’t use the `getTokenSupply` if we want to avoid torn reads. It might happen that we read the Solido account, and we find that the stake accounts together hold 100 SOL (and 100 stSOL is minted for that, but that is not stored in the Solido instance). Then somebody withdraws, now there is only 50 SOL left and 50 stSOL minted. Then we call `getTokenSupply`, and we find 50 stSOL. Now the snapshot claims to have 100 SOL but only 50 stSOL. But that’s not true, and even worse, there never even was a time at which it _was_ true!

To avoid this, we have to use `getMultipleAccounts`. That’s the only tool the Solana RPC gives us to avoid torn reads. We can include the stSOL mint in the accounts to query, and then use `encoding: 'jsonParsed`. It returns data of this shape for SPL token mints:

```json
"value": [
  {
    "data": {
      "parsed": {
        "info": {
          "decimals": 9,
          "freezeAuthority": null,
          "isInitialized": true,
          "mintAuthority": "8kRRsKezwXS21beVDcAoTmih1XbyFnEAMXXiGXz6J3Jz",
          "supply": "1367470251242903"
        },
        "type": "mint"
      },
      "program": "spl-token",
      "space": 82
    },
    "executable": false,
    "lamports": 1461600,
    "owner": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
    "rentEpoch": 268
  }
]
```

(If you want to be really strict, we also need to avoid `connection.getMinimumBalanceForRentExemption`, because it performs a separate read, and read and parse `SYSVAR_RENT` instead. But the rent is unlikely to change, so I can live with that being a separate read. I expect it would be faster to do everything in a single `getMultipleAccounts` read though, it saves a network round trip.)

ruuda (commented):
I reviewed a few more files, only a few left to go!

js/src/tests/data/snapshot.ts
```diff
@@ -0,0 +1,768 @@
...
+      new BN(
+        'fedef235d5df2a94343d6348287dbfc85f31ffc815d800ce9d6b471971cb7162',
+        'hex',
+        'le'
+      )
+    ),
+    exchange_rate: {
+      computed_in_epoch: new BN('0100', 'hex'),
+      st_sol_supply: new BN('35e06bf954dd5', 'hex'),
+      sol_balance: new BN('36a84a87b4459', 'hex'),
```
ruuda, line 25: This is fine, but it may be more readable to just use base 10 here, i.e.

```javascript
computed_in_epoch: new BN(256),
st_sol_supply: new BN(947_808_007_179_733),
sol_balance: new BN(961_542_925_010_009),
```

Or is this data generated and this is the default way in which it’s printed? If so, please add a comment to explain how to re-generate the data if needed.

js/src/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+  const [stakeAccountAddress] = await PublicKey.findProgramAddress(
+    bufferArray,
+    solidoProgramId
+  );
+
+  return stakeAccountAddress;
+};
+
+/**
+ * Get the stake account that has the most amount of SOL staked(heaviest)
```
ruuda, line 58:
```suggestion
 * Get the stake account that has the most amount of SOL staked (heaviest)
```

js/src/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @param snapshot Snapshot of the Solido stats
+ * @returns Heaviest stake account
+ */
+export const getHeaviestValidatorStakeAccount = (
+  snapshot: Snapshot
+): Snapshot['validatorsStakeAccounts'][0] => {
+  const sortedValidatorStakeAccounts = snapshot.validatorsStakeAccounts.sort(
```
ruuda, line 65: To get the validator with the most stake, we don’t need to sort (which has O(n log n) complexity, where n is the number of validators.) A single pass is sufficient. (Which has O(n) complexity.)

```typescript
let heaviestStakeAccount = snapshot.validatorsStakeAccounts[0];
for (stakeAccount in snapshot.validatorsStakeAccounts) {
    if (stakeAccount.balance > heaviestStakeAccount.balance) {
        heaviestStakeAccount = stakeAccount;
    }
}
return heaviestStakeAccount;
```
(Assuming we can overload `operator>` for `Lamports`.)

ruuda (commented):
I did a first pass over all files now!

js/src/stats.ts
```diff
@@ -53,56 +60,55 @@ export const getStSolSupply = (
...
-  return {
-    solBalance: { lamports: totalSolInLamports },
-    stSolSupply: { stLamports: stSolSupplyInLamports },
-  };
+  return totalSolInLamports.toNumber() / stSolSupplyInStLamports.toNumber();
 };
 
 /**
- * Get the number of all the token accounts exist for the token specified by the mint address
+ * Get the number of  token accounts that exist for the token specified by the mint address
```
ruuda, line 71:
```suggestion
 * Get the number of token accounts that exist for the token specified by the mint address
```

js/src/stats.ts
```diff
@@ -0,0 +1,123 @@
...
+};
+
+/**
+ * Get the number of all the token accounts exist for the token specified by the mint address
+ * @param connection Connection to the cluster
+ * @param tokenMintAddress Address of the token mint account
+ * @param tokenProgramId Address of the Token program
+ * @returns Number of token accounts
+ */
+export const getTotalNumberOfTokenAccounts = async (
```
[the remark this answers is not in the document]
ruuda: This comment is not yet addressed.

js/src/utils.ts
```diff
@@ -55,21 +57,46 @@ export const calculateStakeAccountAddress = async (
...
+
+  return new StLamports(amount.lamports.toNumber() / exchangeRate);
+};
+
+/**
+ * Exchange stSOL to SOL
+ * @param snapshot Snapshot of the Solido stats
+ * @param amount stSOL to exchange
+ */
+export const exchangeStSol = (snapshot: Snapshot, amount: StLamports) => {
```
ruuda, line 98: We should add `Lamports` as the return type.

js/src/utils.ts
```diff
@@ -55,21 +57,46 @@ export const calculateStakeAccountAddress = async (
...
+  const exchangeRate = getExchangeRate(snapshot);
+
+  return new Lamports(amount.stLamports.toNumber() * exchangeRate);
```
ruuda, line 101: We can be more precise here by using

```
(amount.stLamports * snapshot.solido.exchange_rate.sol_balance).toNumber() / snapshot.solido.exchange_rate.st_sol_supply.toNumber();
```

This way the conversion to float doesn’t happen until after multiplying the two numbers, and we don’t divide by a quotient.

js/src/tests/updateSnapshot.ts
```diff
@@ -0,0 +1,66 @@
...
+import { getSnapshot } from '../snapshot';
+
+const updateSnapshot = async () => {
+  const connection = new Connection(clusterApiUrl('mainnet-beta'));
+
+  const snapshot = await getSnapshot(connection, MAINNET_PROGRAM_ADDRESSES);
+
+  let rawString = util.inspect(snapshot, true, 10, false);
+
+  // Replace the ugly <BN: ...> public keys with readable ones
```
ruuda, line 16: :+1: Nice, this makes the tests easier to read indeed.

ruuda (approved):
I have a few small comments, but aside from those, let’s get this merged indeed. We can iterate on it from here, and we still need to test thoroughly either way before making a public release.

js/src/tests/mainnet.test.ts
```diff
@@ -94,6 +97,22 @@ describe('Utility functions', () => {
...
+    expect(stSolReceived.stLamports.toString()).toBe('977650658');
+  });
+
+  it('exchanges stSOL to get SOL', async () => {
+    const stSolToExchange = new StLamports('977650659');
+
+    const solReceived = exchangeStSol(snapshotDump, stSolToExchange);
+
+    expect(solReceived.lamports.toString()).toBe('1000000000');
+  });
```
ruuda, line 115: :+1: Thanks for adding a test.

ruuda (approved):
LGTM!

Issue #459, opened 2021-11-16
[Anchor] Make maintenance bot call ClaimRewards on Anker instance


ruuda:
We have the maintenance bot that submits maintenance transactions for Solido. Once the Anchor integration is ready, something needs to periodically call `ClaimRewards` if the value of the reserve in SOL exceeds the bSOL supply. The existing maintenance bot could do this; it would already watch the Anker instance for #458, so we can also make it submit the `ClaimRewards` transaction if it notices there are rewards to claim.

enriquefynn:
We split the `ClaimRewards` into `SellRewards` and `SendRewards`.
Selling rewards is being implemented in #480.

ruuda:
This has been implemented in #480 and #481.

PR #480, opened 2021-12-13
Perform `SellRewards` on the maintenance bot

enriquefynn:
Periodically checks if it's possible to sell rewards from the Anker instance. If it's possible, sell the rewards.
Steps towards #459


cli/src/maintenance.rs
```diff
@@ -692,6 +709,21 @@ impl SolidoState {
         None
     }
 
+    /// If there is a validator ready for removal, try to remove it.
```
ruuda, line 712: This comment is copy-pasted from elsewhere, please add an up-to-date doc comment.

cli/src/maintenance.rs
```diff
@@ -692,6 +709,21 @@ impl SolidoState {
...
+    pub fn try_sell_anker_rewards(&self) -> Option<(Instruction, MaintenanceOutput)> {
+        if self.anker_state.ust_reserve_balance > MicroUst(0) {
```
ruuda, line 714: This is the condition for _sending_ the rewards, not selling. For selling we need to compute the SOL value of the stSOL in the reserve according to Solido's exchange rate, and if that exceeds the bSOL supply, then we sell the rewards.

ruuda (commented):
I haven’t done a full review yet, but I started building on top of this for calling `SendRewards`, and two things stood out to me.

cli/src/maintenance.rs
enriquefynn: Yeah, that's true. Thanks! I got sidetracked with another PR to setup a token pool, Will come back to this.

cli/src/anker_state.rs
```diff
@@ -0,0 +1,126 @@
...
+    ) -> Result<Self> {
+        let (anker_instance, _anker_bump_seed) =
+            find_instance_address(&anker_program_id, solido_address);
+        let anker = config.client.get_anker(&anker_instance)?;
+
+        let token_swap_account = config.client.get_account(&anker.token_swap_pool)?;
+        let token_swap = spl_token_swap::state::SwapV1::unpack(token_swap_account.data())?;
+
+        let (anker_ust_reserve, _anker_ust_reserve_bump_seed) =
+            find_ust_reserve_account(&anker_program_id, solido_address);
```
ruuda, line 44: So we discovered during testing that these `solido_address`es should be `anker_instance` instead, but we already fixed that in #481 so I’m not going to comment on it further in this PR.

cli/src/anker_state.rs
```diff
@@ -0,0 +1,126 @@
...
+        anker_program_id: Pubkey,
+        solido_address: &Pubkey,
+        solido: &Lido,
+    ) -> Result<Self> {
+        let (anker_instance, _anker_bump_seed) =
+            find_instance_address(&anker_program_id, solido_address);
+        let anker = config.client.get_anker(&anker_instance)?;
+
+        let token_swap_account = config.client.get_account(&anker.token_swap_pool)?;
+        let token_swap = spl_token_swap::state::SwapV1::unpack(token_swap_account.data())?;
```
ruuda, line 41: We have a function to make this a bit nicer:
```rust
let token_swap: tpl_token_swap::state::SwapV1 = config.client.get_unpack(&anker.token_swap_pool)?;
```

cli/src/anker_state.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        let (anker_st_sol_reserve, _anker_st_sol_reserve_bump_seed) =
+            find_st_sol_reserve_account(&anker_program_id, solido_address);
+        let st_sol_reserve_balance =
+            StLamports(config.client.get_spl_token_balance(&anker_st_sol_reserve)?);
+
+        let b_sol_mint_account = config.client.get_spl_token_mint(&anker.b_sol_mint)?;
+        let b_sol_total_supply_amount = BLamports(b_sol_mint_account.supply);
+
+        let (ust_account, ust_mint, st_sol_account) =
```
ruuda, line 56: Maybe call these `swap_ust_account` and `swap_st_sol_account`, to avoid confusing them with the reserve?

cli/src/anker_state.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        let (st_sol_reserve_account, _st_sol_reserve_bump_seed) =
+            find_st_sol_reserve_account(&self.anker_program_id, &solido_address);
+
+        let (reserve_authority, _reserve_authority_bump_seed) =
+            find_reserve_authority(&self.anker_program_id, &solido_address);
+
+        let (token_pool_authority, _authority_bump_seed) = Pubkey::find_program_address(
+            &[&self.anker.token_swap_pool.to_bytes()[..]],
+            &anker::orca_token_swap_v2::id(),
```
ruuda, line 103: We can get this from the owner of the token swap pool account, and that one we store in the Anker instance. That way we don’t have to hard-code the program address here, and that allows us to test on testnet/devnet even if the program is deployed at a different address.

cli/src/config.rs
```diff
@@ -652,6 +656,10 @@ cli_opt_struct! {
         #[clap(long)]
         solido_program_id: Pubkey,
 
+        /// Address of the Anker program.
+        #[clap(long, value_name = "address")]
+        anker_program_id: Pubkey,
```
ruuda, line 661: If we make it optional for `perform-maintenance`, we might as well make it optional here and keep things uniform.

cli/src/daemon.rs
```diff
@@ -63,6 +63,9 @@ struct MaintenanceMetrics {
 
     /// Number of times we performed `Unstake` on an active validator for balancing purposes.
     transactions_unstake_from_active_validator: u64,
+
+    /// Number of times we performed `SellRewards` on the Anker instance.
+    transactions_sell_rewards: u64,
```
ruuda, line 68: This also needs a few new lines below in `write_prometheus`, but I fixed that in #481 already.

cli/src/maintenance.rs
```diff
@@ -692,6 +717,32 @@ impl SolidoState {
...
+    /// Try to sell the extra stSOL rewards for UST tokens.
+    pub fn try_sell_anker_rewards(&self) -> Option<(Instruction, MaintenanceOutput)> {
+        let anker_state = self.anker_state.as_ref()?;
+        let reserve_st_sol = anker_state.st_sol_reserve_balance;
+        let st_sol_amount = self
+            .solido
+            .exchange_rate
+            .exchange_sol(Lamports(anker_state.b_sol_total_supply_amount.0))
+            .unwrap();
+        if let Ok(rewards) = reserve_st_sol - st_sol_amount {
```
ruuda, line 729: A nicer way to write
```rust
if let Ok(x) = f() {
    Some(g(x))
} else {
    None
}
```
is
```
f().ok().map(g)
```
though if `g` is an ugly closure, in this case because the function returns `Option`, we can also write
```rust
let x = f().ok()?;
g(x)
```
So concretely, we can write
```rust
let rewards = (reserve_st_sol - st_sol_amount).ok()?;
```
and dedent the remainder.

cli/src/maintenance.rs
```diff
@@ -692,6 +717,32 @@ impl SolidoState {
...
 
+    /// Try to sell the extra stSOL rewards for UST tokens.
+    pub fn try_sell_anker_rewards(&self) -> Option<(Instruction, MaintenanceOutput)> {
+        let anker_state = self.anker_state.as_ref()?;
+        let reserve_st_sol = anker_state.st_sol_reserve_balance;
+        let st_sol_amount = self
+            .solido
+            .exchange_rate
+            .exchange_sol(Lamports(anker_state.b_sol_total_supply_amount.0))
+            .unwrap();
```
ruuda, line 728: Note to self: this unwrap happens only if the rational multiplication in `exchange_sol` overflows. In the first epoch with zero deposits, it should still work fine.

@enriquefynn can you add a comment to clarify this, and use `.expect()` instead of `.unwrap()` to also codify that reason? (And also to avoid people saying “unwrap is baaad, mkay?”)

ruuda (commented):
Looks mostly good! I have a few comments.

cli/src/anker_state.rs
enriquefynn: Good point, do you think we should also change it on `instruction.rs` instead of passing a const?

cli/src/maintenance.rs
enriquefynn: Done like it... We still need to check if the rewards is > 0, because it could be 0, but we should return None if it is so.

cli/src/anker_state.rs
enriquefynn: Since we now check for the owner to figure out the token pool program id, this is the best option.

cli/src/anker_state.rs
ruuda: Yes, we should.

ruuda (approved):
LGTM!

PR #481, opened 2021-12-15
Call `SendRewards` from maintainer

ruuda:
This modifies the maintainer daemon to call `SendRewards` in case there is UST in the UST reserve.

Also contains a number of fixes that @enriquefynn and me pair-programmed while trying to get this to work on Solana devnet.


anker/src/lib.rs
```diff
@@ -43,6 +43,10 @@ pub fn find_instance_address(anker_program_id: &Pubkey, solido_instance: &Pubkey
     Pubkey::find_program_address(&[solido_instance.as_ref()], anker_program_id)
 }
 
+// TODO(ruuda): Use this below.
```
ruuda, line 46: I started working on this, but it is going to be a pretty big and invasive change, so I split it off into a separate PR: https://github.com/ChorusOne/solido/pull/488

cli/src/maintenance.rs
```diff
@@ -742,6 +768,39 @@ impl SolidoState {
...
+        }
+
+        // Use the low 32 bits of the current slot as the nonce. This ensures
+        // that we don't repeat the nonce for at least 2^32 slots, which is 327
+        // years even if the block times were 100ms, but in practice block times
+        // are 600-700ms, so it’s even longer. Do we never send multiple
+        // transactions based on the same slot though? If we do, then the state
+        // we observe should also be identical, and only one of the transactions
+        // should succeed.
+        let wormhole_nonce = (self.clock.slot & 0xffff_ffff) as u32;
```
enriquefynn, line 785: Great 👍

cli/src/maintenance.rs
```diff
@@ -742,6 +768,39 @@ impl SolidoState {
...
+
+        let (instruction, additional_signer) = anker_state.get_send_rewards_instruction(
+            self.solido_address,
+            self.maintainer_address,
+            wormhole_nonce,
+        );
+
+        let ust_amount = anker_state.ust_reserve_balance;
+        let output = MaintenanceOutput::SendRewards { ust_amount };
+        println!("Trying to send rewards: {:?}", &output);
```
enriquefynn, line 795: Is this a debug msg? Cause it can mess the tests (if we have one)

cli/src/maintenance.rs
```diff
@@ -1510,14 +1569,19 @@ pub fn try_perform_maintenance(
...
-        Some((instruction, output)) => {
-            // For maintenance operations, the maintainer is the only signer,
-            // and that should be sufficient.
-            config.sign_and_send_transaction(&[instruction], &[config.signer])?;
-            Ok(Some(output))
+        Some(maintenance_instruction) => {
+            // Usually the maintainer is the only signer, but in some cases we
+            // need to generate a new fresh keypair, which then also is a signer.
+            let mut signers: Vec<&dyn Signer> = vec![config.signer];
+            for keypair in &maintenance_instruction.additional_signers {
```
enriquefynn, line 1580: That's how we get additional signers, nice!

enriquefynn (commented):
Looks mostly good, I just have a few questions

anker/src/state.rs
```diff
@@ -532,19 +536,9 @@ impl Anker {
             accounts.config_key.key,
         )?;
         check_wormhole_account(
-            "custody key",
-            &wormhole_transfer_args.custody_key,
-            accounts.custody_key.key,
-        )?;
-        check_wormhole_account(
```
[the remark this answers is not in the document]
ruuda: Very good catch! Fixed. I think what happened here is that I removed the custody key and custody signer key, because they are only used for native tokens, not wrapped tokens like UST, but I was a bit too eager and also removed this check in between that I shouldn’t have removed.

anker/src/processor.rs
```diff
@@ -392,63 +400,104 @@ fn process_change_token_swap_pool(
...
+
+    // Stack space is scarce in this function, so we put as many things as we can
+    // in a scope to make sure the stack space of the temporaries is reclaimed.
+    {
+        // Wormhole signs the SPL token transfer with its "authority signer key",
+        // which means we need to authorize that key to modify our UST reserve.
+        let instr = Box::new(spl_token::instruction::approve(
+            accounts.spl_token.key,
+            accounts.ust_reserve_account.key,
+            accounts.authority_signer_key.key,
```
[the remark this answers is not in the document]
ruuda: I’m not you, but we do check it now since c5f3d30c52e95289bc068160d4311fb95846954d :-)

cli/src/maintenance.rs
ruuda: Yes, good catch, that is a debugging leftover, removed now. We shouldn’t have bare `println!`s in the CLI code because it will mess up the `--output json` mode.

enriquefynn (approved):
LGTM!

PR #482, opened 2021-12-15
Create token pool instance

enriquefynn:
Create the token pool instance from the client, used for testing purposes only.


ruuda (approved):
LGTM! I rebased and added one more commit to clarify that the `solido anker create-token-pool` command is intended for testing only.

PR #483, opened 2021-12-16
Add script to print details of past multisig LDO transfers

ruuda:
We funded incentive programs three times now, and it is becoming a bit of a mess to see what happened, to find transactions, and to check if addresses match previous addresses.

I tried putting the transactions in a table in Notion, but doing all of that manually is a big hassle and very prone to human error.

To address this, I made a script that prints a Markdown table with all transaction details. We can then put this table in the docs.

### DIFF
--- a/scripts/print_incentive_history.py
+++ b/scripts/print_incentive_history.py
@@ -175,9 +175,7 @@ def main() -> None:
         assert (
             details.amount / 1e8 == entry.amount_ldo
         ), f'Expected amount in script to match transaction for {entry}'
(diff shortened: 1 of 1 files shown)


ruuda:
I went ahead and deployed the docs page, but I’d still appreciate a review on this before we merge it.

enriquefynn (approved):
LGTM!

One small suggestion would be to separate the entries in another file and load the file (csv?) from the script. I'm also ok with it being in the script.

PR #484, opened 2021-12-16
Upgrade Solana to 1.7.15

enriquefynn:
Tests were failing for Solana SDK 1.7.8. Updating it to 1.7.15 seems to fix the problem.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,14 +59,14 @@ jobs:
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
           target/
-        key: build-1.7.8-v2-${{ hashFiles('Cargo.lock') }}
-        restore-keys: build-1.7.8-v2
+        key: build-1.7.15-v2-${{ hashFiles('Cargo.lock') }}
+        restore-keys: build-1.7.15-v2
 
     - name: Install development tools
       run: |
         sudo apt update
         sudo apt-get install -y libudev-dev
-        sh -c "$(curl -sSfL https://release.solana.com/v1.7.8/install)"
+        sh -c "$(curl -sSfL https://release.solana.com/v1.7.15/install)"
 
     - name: Run unit tests
       run: |
@@ -153,8 +153,8 @@ jobs:
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
           target/
-        key: lint-1.7.8-v1-${{ hashFiles('Cargo.lock') }}
-        restore-keys: lint-1.7.8-v1
+        key: lint-1.7.15-v1-${{ hashFiles('Cargo.lock') }}
+        restore-keys: lint-1.7.15-v1
 
     - name: Install linters
       run: |
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -23,15 +23,15 @@ jobs:
         path: |
           target
           multisig/target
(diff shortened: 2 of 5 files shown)


ruuda (commented):
There are many places that still reference 1.7.8, can you update those as well?

```
$ git grep '1\.7\.8' -- :^Cargo.lock :^CHANGELOG.md
Cargo.toml:# We take our fork of Solana 1.7.8, which has one patch applied:
anker/Cargo.toml:solana-program = "1.7.8"
anker/Cargo.toml:solana-program-test = "1.7.8"
anker/Cargo.toml:solana-sdk = "1.7.8"
cli/Cargo.toml:solana-account-decoder = "1.7.8"
cli/Cargo.toml:solana-clap-utils = "1.7.8"
cli/Cargo.toml:solana-cli-config = "1.7.8"
cli/Cargo.toml:solana-client = "1.7.8"
cli/Cargo.toml:solana-config-program = "1.7.8"
cli/Cargo.toml:solana-logger = "1.7.8"
cli/Cargo.toml:solana-program = "1.7.8"
cli/Cargo.toml:solana-remote-wallet = "1.7.8"
cli/Cargo.toml:solana-sdk = "1.7.8"
cli/Cargo.toml:solana-stake-program = "1.7.8"
cli/Cargo.toml:solana-vote-program = "1.7.8"
program/Cargo.toml:solana-program = "1.7.8"
program/Cargo.toml:solana-program-test = "1.7.8"
program/Cargo.toml:solana-sdk = "1.7.8"
program/Cargo.toml:solana-vote-program = "1.7.8"
testlib/Cargo.toml:solana-program = "1.7.8"
testlib/Cargo.toml:solana-program-test = "1.7.8"
testlib/Cargo.toml:solana-sdk = "1.7.8"
testlib/Cargo.toml:solana-vote-program = "1.7.8"
```

The one thing I am worried about is that Anchor-the-eDSL pins the Rust version to a particular version, so it’s hard to update this without updating Anchor. If it’s a problem we can fork Anchor and change the version bounds to be less strict.

enriquefynn:
> There are many places that still reference 1.7.8, can you update those as well?

We can update just the binary utils (what I did here). Would you like to update all reference to 1.7.8 as well? I think we can do this at a later time, it's going to cause more pain than simply updating the Solana binaries.

ruuda:
> We can update just the binary utils (what I did here). Would you like to update all reference to 1.7.8 as well? 

I’m a bit afraid that if we don’t keep them in sync, we will run into other inexplicable issues down the line. What was the original issue with 1.7.8 that you ran into? Maybe there’s a different way to sidestep it?

enriquefynn:
> > We can update just the binary utils (what I did here). Would you like to update all reference to 1.7.8 as well?
> 
> I’m a bit afraid that if we don’t keep them in sync, we will run into other inexplicable issues down the line. What was the original issue with 1.7.8 that you ran into? Maybe there’s a different way to sidestep it?

Couldn't download it I guess, see #482, fails with `Error: Unknown release: 1.7.8` 🤷

ruuda:
Ugh, they must have changed their install script. Well it’s kind of our own fault for relying on `curl | bash`, we should have finished packaging Solana for Nixpkgs, then we can keep it cached on Cachix as long as we like, and others will not be able to break our build at inconvenient times. But sorting that out will take time in itself :confused: 

Let’s update then for now ...

PR #488, opened 2021-12-22
[WIP] Make addresses more type-safe

ruuda:
As @enriquefynn and me discovered the hard way, it is easy to confuse addresses when everything is a `Pubkey`. Add some newtype wrappers to make them harder to confuse.

Note: This is going to be a pretty invasive change. (Mostly trivial edits, but still.) So I’m opening this as a separate pull request.

-------

<sub>Side note, the state of Solana tooling is like raw assemply programming even before the invention of C, or like C where you treat everything as a `void*`. Solana addresses are like void pointers: they tell you the address of some piece of data, but nothing about what is stored there, you have to cast (deserialize, in the case of Solana) to the right type. If you get it wrong your program crashes, and there is no typechecker to help. (Which is ironic, because Rust has a strong type system, but it’s hard to leverage it due to the way the Solana runtime APIs work in Rust.)</sub>

<sub>Anchor-the-eDSL is a big step forward here. I think there is room for a full-blown non-embedded DSL that is designed to compile types to accounts, and which is built from the start to make static analysis easy (e.g. first class preconditions/postconditions, invariants on types, etc).</sub>


ruuda:
I continued with this, but it is even more invasive than I expected, and I am no longer convinced that this is worth doing. 

* It helps to prevent mixing up accounts in many places, but then it moves that to other places. It is more obvious to spot there (especially something like `AnkerAddress(solido)` should be a red flag), but it does also clutter the code and make things a bit less readable in general.
* We end up with a _lot_ of wrapping and unwrapping these types.
* We end up with inconsistencies about where we use it. For example, it would be nice if we could use the typed addresses in our `accounts_struct!` macro, but currently we can’t, and it would be difficult to add.

I think typed addresses are very valuable, but they need to be pervasive enough that you don’t have to wrap/unwrap them all the time, before the benefit starts to outweigh the hassle.

I’ll close this then, and we’ll have to rely on tests to catch cases of mixing up accounts.

PR #492, opened 2022-01-06
Add test for `solido anker show`

### LINKED ISSUE #456: [Anchor] Add command to create a new Anker instance

ruuda:
This extends the test script to call `solido anker show`, which acts two-fold:

* It tests that `solido anker show` works.
* We can use it in the remainder of the test to confirm that the instance looks as expected.

This is also the final step for #456, so closes #456.


enriquefynn (approved):
LGTM!
It does more than test the solido anker show.

PR #493, opened 2022-01-06
Clarify maintainer CLI options, remove unused option

ruuda:
The unused option was a leftover from when we did not have the maintainer duties yet. Also clarify that `--stake-time` is only intended for testing.


enriquefynn (approved):
LGTM!

PR #495, opened 2022-01-06
Avoid sending/selling small amounts of Anker rewards

ruuda:
The maintainer pays for these transactions, it doesn't make sense to pay if the amount moved is not much larger than the transaction fee. Especially for Wormhole transfers, which are expensive due to the accounts it allocates.


enriquefynn (approved):
LGTM

ruuda:
I added another check to make sure we don’t sell the rewards if the *output* would be too small (which would cause it to round down to 0).

PR #496, opened 2022-01-07
Allow Anker deposits during initial Solido epoch

ruuda:
On mainnet this will not happen, but in local testing, it might happen that you set up a Solido instance with zero balance (0 SOL, 0 stSOL), and therefore its exchange rate from the balances is undefined. Then if you try to do an Anker deposit in that initial epoch, previously that would fail. This is not a concern on mainnet, but for testing this is annoying, so do allow deposits during the initial epoch, by defining the exchange rate to be 1 stSOL = 1 bSOL, just like Solido defines 1 stSOL =1 SOL if it has zero balance.

### DIFF
--- a/anker/tests/tests/deposit.rs
+++ b/anker/tests/tests/deposit.rs
@@ -13,7 +13,24 @@ use testlib::assert_solido_error;
 const TEST_DEPOSIT_AMOUNT: StLamports = StLamports(1_000_000_000);
 
 #[tokio::test]
-async fn test_successful_deposit() {
+async fn test_successful_deposit_during_first_epoch() {
+    let mut context = Context::new_with_undefined_exchange_rate().await;
+    let (_owner, recipient) = context.deposit(Lamports(TEST_DEPOSIT_AMOUNT.0)).await;
+
+    let reserve_balance = context
+        .solido_context
+        .get_st_sol_balance(context.st_sol_reserve)
+        .await;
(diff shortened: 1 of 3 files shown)


ruuda:
(The clippy error on CI will be fixed by #492.)

ruuda:
@enriquefynn I am going to go ahead and merge this to get it in before the audit, but please do still take a look once you get back.

anker/tests/tests/deposit.rs
```diff
@@ -13,7 +13,24 @@ use testlib::assert_solido_error;
 const TEST_DEPOSIT_AMOUNT: StLamports = StLamports(1_000_000_000);
 
 #[tokio::test]
-async fn test_successful_deposit() {
+async fn test_successful_deposit_during_first_epoch() {
```
enriquefynn, line 16: Nice to add a test for it 👍

testlib/src/anker_context.rs
```diff
@@ -193,7 +193,7 @@ pub struct Context {
 const INITIAL_DEPOSIT: Lamports = Lamports(1_000_000_000);
 
 impl Context {
-    pub async fn new() -> Self {
+    pub async fn new_with_undefined_exchange_rate() -> Self {
```
enriquefynn, line 196: 👍

enriquefynn (commented):
LGTM!

PR #497, opened 2022-01-07
Add `solido anker deposit` command, use it in Anker test

ruuda:
* Add a new command to the CLI, `solido anker deposit`, to make a deposit. The recipient is the associated token account of the signer, and this account will be created if it doesn’t exist.
* (In hindsight, the `solido deposit` command for Solido deposits can be simplified to not take multiple snapshots, but let’s not do that right now.)
* Use this to extend the Anker test.


enriquefynn (approved):
LGTM

PR #501, opened 2022-01-07
Add script to set up Anker instance on devnet or a local test validator

ruuda:
Add a script to deploy an Anker instance to devnet. This is very similar to the `test_anker.py` script, and it shares a lot with the old `prepare_test_environment.py` that I renamed to `deploy_test_solido.py`. The differences are:

* It uses the actual devnet Wormhole programs, Orca program, and UST address.
* On a local validator, it will create a UST mint and upload the Orca program.
* It outputs a config file so you can easily interact with the instance later on.


tests/deploy_test_anker.py
```diff
@@ -0,0 +1,281 @@
...
+    DEVNET_WORMHOLE_TOKEN_BRIDGE_PROGRAM_ID,
+    '--terra-rewards-address',
+    DEVNET_TERRA_REWARDS_ADDRESS,
+)
+anker_address = result['anker_address']
+print(f'> Created instance at {anker_address}.')
+
+sol_balance_post = float(solana('balance').split(' ')[0])
+total_cost_sol = sol_balance_pre - sol_balance_post
+print(f'\nDeployment cost {total_cost_sol:.3f} SOL.')
```
enriquefynn, line 261: Would be nice to see how much each of the Anker, multisig, and Solido deployments cost.

enriquefynn (approved):
LGTM!

tests/deploy_test_anker.py
ruuda: Would be nice indeed. I intended to still add it for a long time, but this PR has been sitting here for a long time, I’ll just merge it now without this feature.

PR #504, opened 2022-01-12
Delete nightly CI run

ruuda:
This workflow has been failing for months; it fails to airdrop sufficient SOL to run the tests. Let's just delete it then, a test that always fails is not useful, and fixing this is out of our control.


enriquefynn (approved):
Yeah, let's do it. Thanks

PR #505, opened 2022-01-12
Restructure client functions in library

enriquefynn:
For the upcoming listener, we can use the same snapshot functions
instead of rewriting them.

The listener features will be added on top of this PR. Let's not merge it until all features are done


ruuda (approved):
:+1: So far so good. I don’t see any harm in merging this already, there’s less risk of nasty merge conflicts if we merge things early and keep the changes small.

ruuda:
Ah Clippy is unhappy though, can you fix that?

PR #506, opened 2022-01-12
Add simple functions

enriquefynn:
Log into SQLite and calculate APY.


cli/listener/src/main.rs
```diff
@@ -0,0 +1,190 @@
...
+                let p1 = exchange_rate_last.token_a as f32 / exchange_rate_last.token_b as f32;
+                let apy = (p1 / p0).powf(1. / (fraction_of_year));
+                Ok(Some((apy - 1f32) * 100f32))
+            }
+        }
+        _ => Ok(None),
+    }
+}
+
+pub fn log_price(conn: &Connection, exchange_rate: ExchangeRate) -> rusqlite::Result<()> {
```
ruuda, line 132: Nit: "log" sounds to me like it logs something to a log file or stdout/stderr, I would call this `insert_price`.

ruuda (commented):
Good start! I have some comments.

cli/listener/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+authors = ["Chorus One <techops@chorus.one>"]
+description = "Solido Utility for indexing price data"
+license = "GPL-3.0"
+edition = "2018"
+name = "listener"
+version = "1.2.0"
+
+[dependencies]
+rusqlite = "0.26.3"
```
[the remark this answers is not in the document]
enriquefynn: That was the first one I've found, we can switch to `sqlite`, I don't really have a preference, and we won't probably need fancy things.

cli/listener/src/main.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// Whether to output text or json. [default: "text"]
+    // Overwritten by `GeneralOpts` if None.
+    #[clap(long = "output", possible_values = &["text", "json"])]
+    output_mode: Option<OutputMode>,
+
+    /// Unique name for identifying
+    #[clap(long, default_value = "solido")]
+    pool: String,
+
+    /// Poll frequency in seconds, defaults to 86400s = 1 day.
```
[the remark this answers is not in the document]
enriquefynn: Yeah, that looks reasonable :)

cli/listener/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    // Get first logged minimal logged data based on timestamp that is greater than `from_time`.
+    let mut stmt_first = conn.prepare(
+        "SELECT *, MIN(timestamp) FROM exchange_rate WHERE pool = :pool AND timestamp > :t",
+    )?;
+    let mut row_iter = stmt_first.query_map([opts.pool.clone(), from_time.to_string()], row_map)?;
+    let first = row_iter.next();
+
+    // Get first logged maximal logged data based on timestamp that is smaller than `to_time`.
+    let mut stmt_last = conn.prepare(
```
ruuda, line 154: If we issue multiple reads, we should do so in a transaction, to make sure that we get a consistent view of the data. A nice way to enforce that is to have an RAII wrapper type for transactions, and accept that as a function argument, instead of `Connection`. (This also makes it compose better: suppose you wanted to read the APY over different time periods. If this function would start the transaction itself, you still couldn’t do atomic reads of multiple different time periods. But if we accept the transaction as an argument, then they can all happen within the same transaction. (SQL does allow nested transactions, but the syntax is different :hankey: Imagine that in Rust you had to open the first block in a function with `{` but within that you open a block with `(`. Yeah ...))

Would be nice if the crate offers something like a transaction type, but otherwise we can create one.

cli/listener/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+        self.annual_growth_factor().mul_add(100.0, -100.0)
+    }
+}
+
+pub fn get_apy_for_period(
+    conn: &Connection,
+    opts: &Opts,
+    from_time: chrono::DateTime<chrono::Utc>,
+    to_time: chrono::DateTime<chrono::Utc>,
+) -> rusqlite::Result<Option<f64>> {
```
ruuda, line 130: We should return the `IntervalPrices` here, and delay converting to a float as far as possible. A float itself requires an out-of-band description of what time period the APY is measured over. But the `IntervalPrice` type is self-descriptive: if you receive a value of type `IntervalPrice`, and that is all you have, you know what it means. If you receive an `f64`, even if you know it’s an APY, you don’t know over what period.

cli/listener/src/main.rs
```diff
@@ -138,24 +138,51 @@ pub fn get_apy_for_period(
...
+                SELECT *
+                FROM exchange_rate
+                WHERE epoch = (SELECT MIN(epoch) from exchange_rate where pool = :pool AND timestamp > :t)
+              )
+              SELECT
+                *
+              FROM
+                prices_epoch
+              ORDER BY
+                timestamp ASC
```
ruuda, line 159: Small note: we might add a `LIMIT 1` as well. It wouldn’t affect the outcome as we only run `.next()` once anyway, but I can imagine it might affect the query plan. For now I wouldn’t bother changing it and if we ever find that our queries are too slow we can investigate.

cli/listener/src/main.rs
```diff
@@ -138,24 +138,51 @@ pub fn get_apy_for_period(
...
-        "SELECT *, MAX(timestamp) FROM exchange_rate WHERE pool = :pool AND timestamp < :t",
-    )?;
-    let mut row_iter = stmt_last.query_map([opts.pool.clone(), to_time.to_string()], row_map)?;
-    let last = row_iter.next();
+        let mut row_iter =
+            stmt_last.query_map([opts.pool.clone(), to_time.to_string()], row_map)?;
+        let last = row_iter.next();
+        (first, last)
+    };
+    tx.commit()?;
```
ruuda, line 185: Why the commit here? This is a read-only function right? It shouldn’t modify the transaction. I’m not sure how the types work, but it would be nice if we can take the `rusqlite::Transaction` as a non-mutable reference to enforce that.

ruuda (commented):
I have one more comment about the transaction, aside from that this looks good!

cli/listener/src/main.rs
enriquefynn: Yeah... Also we might need to change it to be the first entries of epoch + 200. I'll put a TODO

ruuda: Ah yes, good point :+1:

ruuda (approved):
LGTM

PR #509, opened 2022-01-25
Run daemon on listener

enriquefynn:
Runs the daemon on the listener.


cli/listener/src/main.rs
```diff
@@ -1,32 +1,41 @@
...
     /// URL of cluster to connect to (e.g., https://api.devnet.solana.com for solana devnet)
-    // Overwritten by `GeneralOpts` if None.
     #[clap(long, default_value = "http://127.0.0.1:8899")]
     cluster: String,
 
     /// Whether to output text or json. [default: "text"]
-    // Overwritten by `GeneralOpts` if None.
     #[clap(long = "output", possible_values = &["text", "json"])]
-    output_mode: Option<OutputMode>,
+    output_mode: OutputMode,
```
ruuda, line 38: Hmm now that I’m looking at this again ... why do we have this here? For `solido` one-off commands it makes sense, but for just a daemon that watches the chain ... it shouldn’t output anything aside from logging and for that we don’t need this switch right now. Is the `output_mode` used at all in this crate?

cli/listener/src/main.rs
```diff
@@ -36,24 +45,13 @@ pub struct Opts {
     #[clap(long, default_value = "300")]
     poll_frequency_seconds: u32,
 
-    /// Location of the SQLite DB file.
+    /// Location of the SQLite DB file, defaults to "listener.db".
     #[clap(long, default_value = "listener.db")]
```
ruuda, line 49: Nit: I prefer to use `.sqlite3` extension to make it more obvious what type of file it is.

cli/listener/src/main.rs
```diff
@@ -36,24 +45,13 @@ pub struct Opts {
...
-    pub fn new(
-        config: &mut SnapshotConfig,
-        solido_program_id: &Pubkey,
-        solido_address: &Pubkey,
-    ) -> Result<Self, SnapshotError> {
-        let solido = config.client.get_solido(solido_address)?;
-        Ok(State { solido })
-    }
+    /// Listen address and port for the http server that serves a /metrics endpoint. Defaults to 0.0.0.0:8923.
+    #[clap(long, default_value = "0.0.0.0:8923")]
```
ruuda, line 53: We should pick a different default port than the one used by `solido run-maintainer`. How about 8929, the next prime after 8923? :grimacing:

cli/listener/src/main.rs
```diff
@@ -220,35 +218,309 @@ pub fn insert_price(conn: &Connection, exchange_rate: ExchangeRate) -> rusqlite:
     Ok(())
 }
 
+struct Daemon<'a, 'b> {
+    config: &'a mut SnapshotClientConfig<'b>,
+    opts: &'a Opts,
+
+    /// Mutex where we publish the latest snapshot for use by the webserver.
+    snapshot_mutex: Arc<SnapshotMutex>,
```
ruuda, line 226: So for `solido run-maintainer`, we keep the latest snapshot of the state in memory in an Arc, so we can serve the metrics from it. But for this daemon, we plan to instead read from the state and insert a row into the database based on that, and for any endpoints, we will read from the database. So we could keep the `Metrics` in an Arc, but all of the on-chain data should go in the database instead; we don’t need this snapshot here.

ruuda, line 226: So for `solido run-maintainer`, we keep the latest snapshot of the state in memory in an Arc, so we can serve the metrics from it. But for this daemon, we plan to instead read from the state and insert a row into the database based on that, and for any endpoints, we will read from the database. So we could keep the `Metrics` in an Arc, but all of the on-chain data should go in the database instead; we don’t need this snapshot here.

cli/listener/src/main.rs
```diff
@@ -220,35 +218,309 @@ pub fn insert_price(conn: &Connection, exchange_rate: ExchangeRate) -> rusqlite:
...
+            .rng
+            .gen_range(Duration::from_secs(0)..target_sleep_time);
+        println!("Sleeping {:?} after error ...", sleep_time);
+        std::thread::sleep(sleep_time);
+    }
+
+    /// Publish a new snapshot that from now on will be served by the http server.
+    ///
+    /// This also updates the block time estimator, if applicable.
+    fn publish_snapshot(&mut self, exchange_rate: Option<ExchangeRate>) {
```
ruuda, line 292: Likewise, this is not needed here.

cli/listener/src/main.rs
```diff
@@ -220,35 +218,309 @@ pub fn insert_price(conn: &Connection, exchange_rate: ExchangeRate) -> rusqlite:
...
+    /// Number of times that we checked the price.
+    polls: u64,
+
+    /// Number of times that we tried to get the exchange rate, but encountered an error.
+    errors: u64,
+}
+
+impl Metrics {
+    pub fn write_prometheus<W: std::io::Write>(&self, out: &mut W) -> std::io::Result<()> {
+        Ok(())
```
ruuda, line 331: We should write `solido_listener_polls_total` and `solido_listener_errors_total` here then. (Do we call it `solido_listener`? Maybe `solido_pricedb` would be clearer?)

ruuda (commented):
Looks mostly good! It feels like it’s a lot of duplication between the maintainer daemon we already have, but I also feel that trying to deduplicate that would only complicate things, and it’s mostly boilerplate that is duplicated anyway, so I am fine with it.

cli/listener/src/main.rs
enriquefynn: I think would be nice to show some log, but that can be done just with some `println` 👍

cli/listener/src/main.rs
enriquefynn: ohhh... that, definitely 😬

cli/listener/src/main.rs
ruuda: Yeah exactly, and we can even still have the enum for output instead of raw printlns in the middle of the code if you like, but no need to have it be switchable between `Display` and `serde::Serialize`, just `Display` is sufficient.

cli/listener/src/main.rs
```diff
@@ -213,44 +223,332 @@ pub fn get_apy_for_period(
     // Ok(Some(1.0))
 }
 
-pub fn insert_price(conn: &Connection, exchange_rate: ExchangeRate) -> rusqlite::Result<()> {
+pub fn insert_price(conn: &Connection, exchange_rate: &ExchangeRate) -> rusqlite::Result<()> {
     conn.execute("INSERT INTO exchange_rate (timestamp, slot, epoch, pool, price_lamports_numerator, price_lamports_denominator) VALUES (?1, ?2, ?3, ?4, ?5, ?6)", 
     params![exchange_rate.timestamp.to_string(), exchange_rate.slot, exchange_rate.epoch, exchange_rate.pool,
```
ruuda, line 228: I noticed when inspecting the database that it produces strings of the form `2022-02-01 11:58:39.923080375 UTC`. That’s not ISO-8601. We should use [`to_rfc3339`](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#method.to_rfc3339) to get an ISO-8601-compatible format.

I suppose it also needs a change when we read the value to use `parse_from_rfc3339`. The reason to use ISO-8601 is that it is widely compatible with other applications, (e.g. Python has `datetime.isoformat()` and `fromisoformat()`), and SQLite’s built-in date functions can handle it as well.

cli/listener/src/main.rs
```diff
@@ -36,24 +40,13 @@ pub struct Opts {
     #[clap(long, default_value = "300")]
     poll_frequency_seconds: u32,
 
-    /// Location of the SQLite DB file.
-    #[clap(long, default_value = "listener.db")]
+    /// Location of the SQLite DB file, defaults to "listener.db".
```
ruuda, line 43: Clap will already append the default, we shouldn’t mention it in the doc comment. Right now it prints this:

```
        --db-path <db-path>
            Location of the SQLite DB file, defaults to "listener.db" [default: listener.sqlite3]
```

cli/listener/src/main.rs
```diff
@@ -12,22 +15,23 @@ use solana_sdk::{
...
     #[clap(long, default_value = "http://127.0.0.1:8899")]
     cluster: String,
 
-    /// Whether to output text or json. [default: "text"]
-    // Overwritten by `GeneralOpts` if None.
-    #[clap(long = "output", possible_values = &["text", "json"])]
-    output_mode: Option<OutputMode>,
-
     /// Unique name for identifying
     #[clap(long, default_value = "solido")]
```
ruuda, line 36: Why do we pass this in? I think it would be better to hard-code, the code path for fetching it is already specific to Solido, making what we insert into the database variable only results in incompatibilities.

program/src/token.rs
```diff
@@ -41,24 +41,11 @@ impl PartialOrd for Rational {
...
-        let denominator = self
-            .denominator
-            .checked_mul(rhs.numerator)
-            .ok_or(ArithmeticError)?;
-        Ok(Rational {
-            numerator,
-            denominator,
-        })
+        (self.numerator as f64 * rhs.denominator as f64)
+            / (self.denominator as f64 * rhs.numerator as f64)
```
ruuda, line 48: Can you add a test for this method as well? The previous version behaved unexpectedly, let’s make sure that this one does not. Also, please add a comment to explain why the output type is `f64` and not `Rational`. (Because the product of two rationals would need to be twice as wide, or lose precision, so we choose to make the loss of precision explicit in the types.)

cli/listener/src/main.rs
```diff
@@ -12,56 +15,44 @@ use solana_sdk::{
...
-impl State {
-    pub fn new(
-        config: &mut SnapshotConfig,
-        solido_program_id: &Pubkey,
-        solido_address: &Pubkey,
-    ) -> Result<Self, SnapshotError> {
-        let solido = config.client.get_solido(solido_address)?;
-        Ok(State { solido })
-    }
+    /// Listen address and port for the http server that serves a /metrics endpoint.
```
ruuda, line 45:
```suggestion
    /// Listen address and port for the http server.
```

In this case it’s not just the metrics, the entire point of this daemon is to serve the APY API.

cli/listener/src/main.rs
```diff
@@ -12,56 +15,44 @@ use solana_sdk::{
...
+    listen: String,
 }
 
 #[derive(Debug)]
 pub struct ExchangeRate {
     /// Id of the data point.
     id: i32,
     /// Time when the data point was logged.
-    timestamp: chrono::DateTime<chrono::Utc>,
+    timestamp: chrono::DateTime<chrono::FixedOffset>,
```
ruuda, line 55: Any reason for using `FixedOffset` instead of `Utc`? Not that it matters much, they are still unambiguous at least, but we also have no reason to deal with non-zero UTC offsets.

cli/listener/src/main.rs
```diff
@@ -145,6 +149,8 @@ pub fn get_apy_for_period(
 
     let (first, last) = {
         // Get first logged minimal logged data based on timestamp that is greater than `from_time`.
+        // TODO: Do not limit the query below, but select the first data point
```
ruuda, line 152: Can you open an issue for the to do so we don’t forget?

cli/listener/src/main.rs
```diff
@@ -210,74 +215,417 @@ pub fn get_apy_for_period(
...
+        let sleep_time = self
+            .rng
+            .gen_range(Duration::from_secs(0)..target_sleep_time);
+        println!("Sleeping {:?} after error ...", sleep_time);
+        sleep_time
+    }
+
+    pub fn get_sleep_time(&self) -> Duration {
+        // Sleep until is time to get the next exchange rate.
+        let sleep_time = std::time::Duration::from_secs(self.opts.poll_frequency_seconds as u64);
```
ruuda, line 340:
```suggestion
        let sleep_time = std::time::Duration::from_secs(self.opts.poll_interval_seconds as u64);
```

Technically it’s the poll interval, not the poll frequency. Frequency has units of time<sup>-1</sup>.

cli/listener/src/main.rs
```diff
@@ -210,74 +215,417 @@ pub fn get_apy_for_period(
...
+            name: "solido_pricedb_errors_total",
+            help: "Number of times we encountered an error while trying to get the exchange rate, since launch.",
+            type_: "counter",
+            metrics: vec![Metric::new(self.errors)]
+        })?;
+        if let Some(interval_price) = &self.solido_average_interval_price {
+            write_metric(
+                out,
+                &MetricFamily {
+                    name: "average_apy",
```
ruuda, line 382:
```suggestion
                    name: "solido_pricedb_average_apy",
```

The name should also have `solido_pricedb_` prefix, because Prometheus doesn’t by itself namespace the metrics; applications have to choose their names to not clash themselves.

Also, we are making the same mistake now that we are building this thing for to avoid :see_no_evil: “average APY” ... over what time period? We should clarify that too, e.g. `solido_pricedb_30d_average_apy`.

But maybe it’s better to not return the `IntervalPrices` from the sampling loop, because the interval we choose is arbitrary. We could just return the current price. And if you have the current price as a metric — but we have that already anyway from the maintainer daemon — you can easily compute the desired APY through e.g. `(solido_pricedb_st_sol_price_in_sol / solido_pricedb_st_sol_price_in_sol offset 30d)^(365/30)`.

cli/listener/src/main.rs
```diff
@@ -210,74 +215,417 @@ pub fn get_apy_for_period(
...
+                    help: "Average APY",
+                    type_: "counter",
```
ruuda, line 384: Average APY is not a counter, it’s a gauge. See [the Prometheus docs about metrics type](https://prometheus.io/docs/concepts/metric_types/), they are very helpful.

cli/listener/src/main.rs
```diff
@@ -210,74 +215,417 @@ pub fn get_apy_for_period(
...
+                err, &opts.listen,
+            );
+            std::process::exit(1);
+        }
+    };
+
+    println!("Http server listening on {}", &opts.listen);
+
+    // Spawn a number of http handler threads, so we can handle requests in
+    // parallel. This server is only used to serve metrics, it can be super basic,
```
ruuda, line 438: This will not really be true for this daemon though. Still, 1 thread per core should be fine. Maybe 2 threads per core if we expect them to be IO-bound. Maybe eventually we need to make it configurable on the command line, but for now 1 thread per core would be fine. We shouldn’t lie in the comment though.

cli/listener/src/main.rs
```diff
@@ -210,74 +215,417 @@ pub fn get_apy_for_period(
...
+    insert_price(&conn, &exchange_rate).unwrap();
+
+    let apy = get_interval_price_for_period(
+        conn.unchecked_transaction().unwrap(),
+        chrono::Utc.ymd(2020, 7, 7).and_hms(0, 0, 0),
+        chrono::Utc.ymd(2022, 7, 8).and_hms(0, 0, 0),
+        SOLIDO_ID.to_owned(),
+    )
+    .expect("Failed when getting APY for period");
+    apy.unwrap().growth_factor();
```
ruuda, line 630: Shouldn’t we also check the returned value? :sweat_smile:

ruuda (commented):
I have a few small comments/suggestions, but aside from those this looks good!

cli/listener/src/main.rs
enriquefynn: We can do it in UTC, the problem was that `parse_from_rfc3339` returns `FixedOffset`.

cli/listener/src/main.rs
enriquefynn: Great! That's much better, thanks. I just did `now() - 30d` and used the APY function to calculate it. Should be working.

cli/listener/src/main.rs
```diff
@@ -108,33 +100,52 @@ impl IntervalPrices {
...
     tx: rusqlite::Transaction,
-    opts: &Opts,
     from_time: chrono::DateTime<chrono::Utc>,
     to_time: chrono::DateTime<chrono::Utc>,
+    pool: String,
 ) -> rusqlite::Result<Option<IntervalPrices>> {
     let row_map = |row: &Row| {
-        let timestamp_iso8601: String = row.get(1)?;
+        let timestamp: String = row.get(1)?;
+        let timestamp_iso8601 = chrono::DateTime::from_utc(
```
ruuda, line 140: We should be able to use [`DateTime::<Utc>::from`](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#impl-From%3CDateTime%3CFixedOffset%3E%3E), then we don’t have to go through the scary `.naive_utc()`. To be clear, the current code does look correct to me, but I would try to avoid naive datetimes as much as possible.

cli/listener/src/main.rs
```diff
@@ -210,74 +222,396 @@ pub fn get_apy_for_period(
...
+        })?;
+        if let Some(interval_price) = &self.solido_average_30d_interval_price {
+            write_metric(
+                out,
+                &MetricFamily {
+                    name: "solido_pricedb_30d_average_apy",
+                    help: "Average 30d APY",
+                    type_: "gauge",
+                    metrics: vec![Metric::new(interval_price.annual_percentage_rate())
+                        .with_label("solido", "APY".to_string())],
```
ruuda, line 390: What is the point of this label? Labels are key-value pairs that you attach to a metric to distinguish multiple instances of the same metrics. For example, you might have two instances of an http server, each with their own request counter, and you can have labels `instance="server1"` and `instance="server2"`.

In this particular case, if in the future we want to support multiple pools, it would make sense to add a `pool="solido"` label. Then you can query `solido_pricedb_30d_average_apy` in Prometheus or Grafana, and it will return the 30d APYs of all pools as separate series in the same chart.

cli/listener/src/main.rs
```diff
@@ -210,74 +222,396 @@ pub fn get_apy_for_period(
...
+fn serve_request(request: Request, metrics_mutex: &MetricsMutex) -> Result<(), std::io::Error> {
+    // Take the current snapshot. This only holds the lock briefly, and does
+    // not prevent other threads from updating the snapshot while this request
+    // handler is running.
+    let metrics = metrics_mutex.lock().unwrap().clone();
+
+    // We don't even look at the request, for now we always serve the metrics.
+
+    let mut out: Vec<u8> = Vec::new();
+    let is_ok = metrics.write_prometheus(&mut out).is_ok();
```
ruuda, line 408: We can assume that this succeeds; we are wirting to a `Vec<u8>`. If we were writing to an `io::Write` backed by a file or a socket or something, we might get an IO error, but with a vector, the worst thing that could happen is running out of memory, but we panic in that case either way.

program/src/token.rs
```diff
@@ -365,4 +355,29 @@ pub mod test {
...
+        let x = Rational {
+            numerator: 18446744073709551557,
+            denominator: 20116405751046403,
+        };
+        let y = Rational {
+            numerator: 69088929115017047,
+            denominator: 18446744073709551533,
+        };
+        let div_result = x / y;
+        assert_eq!(div_result, 244838.99999999997); // Checked with WolframAlpha and adjusted to Rust's precision.
```
ruuda, line 370: :+1:

ruuda (commented):
LGTM aside from a few small comments.

cli/listener/src/main.rs
enriquefynn: True, didn't see that `DateTime::<Utc>` implemented `FromStr`, thanks!

cli/listener/src/main.rs
enriquefynn: Yes... For now we can use it without a label, since we have only Solido.

cli/listener/src/main.rs
ruuda: Not `FromStr`, `From`!

cli/listener/src/main.rs
```diff
@@ -137,12 +138,8 @@ pub fn get_interval_price_for_period(
 ) -> rusqlite::Result<Option<IntervalPrices>> {
     let row_map = |row: &Row| {
         let timestamp: String = row.get(1)?;
-        let timestamp_iso8601 = chrono::DateTime::from_utc(
-            chrono::DateTime::parse_from_rfc3339(&timestamp)
-                .expect("Invalid timestamp format.")
-                .naive_utc(),
-            chrono::Utc,
-        );
+        let timestamp_iso8601 = chrono::DateTime::<chrono::Utc>::from_str(&timestamp)
```
ruuda, line 141: Ah but the `from_str` also works I guess.

cli/listener/src/main.rs
```diff
@@ -405,22 +401,20 @@ fn serve_request(request: Request, metrics_mutex: &MetricsMutex) -> Result<(), s
...
-            &b"text/plain; version=0.0.4; charset=UTF-8"[..],
-        )
-        .expect("Static header value, does not fail at runtime.");
-        request.respond(Response::from_data(out).with_header(content_type))
-    } else {
-        request.respond(Response::from_string("error").with_status_code(500))
-    }
+    metrics
+        .write_prometheus(&mut out)
+        .expect("Error when writing metrics to Vec, probably out of memory.");
```
ruuda, line 406:
```suggestion
        .expect("We must handle the error because of io::Write, but writing to a Vec does not fail.");
```

Rust's default allocator panics when it runs out of memory, it will not get here. (And then even on Linux there is overcommit by default, so before Rust’s allocator fails, your system will start trashing and OOM killing things.)

ruuda (approved):
LGTM!

cli/listener/src/main.rs
enriquefynn: Ah sure, but we can use [DateTime::<Utc>::from_str](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#impl-FromStr) to get the string from the data base.

ruuda (approved):

PR #511, opened 2022-01-27
Store historical stSOL prices in Anker state

ruuda:
This is a mitigation against sandwiching `Anker::SellRewards`, as pointed out by Neodyme in their audit. (I thought I had opened an issue for that, but it looks like I didn’t.)

To do:

* [x] Add tests.
* [x] Modify `solido anker show` to also print this data.
* [x] Use the price to set `min_out` in `SellRewards`.
* [x] Use the age to block `SellRewards` if the historical prices are too old or too young.


anker/src/processor.rs
```diff
@@ -240,6 +310,19 @@ fn process_sell_rewards(program_id: &Pubkey, accounts_raw: &[AccountInfo]) -> Pr
...
+    let oldest_sample = anker.historical_st_sol_prices.first();
+    let slots_elapsed = clock.slot.saturating_sub(oldest_sample.slot);
+    if slots_elapsed > POOL_PRICE_MAX_SAMPLE_AGE {
+        msg!(
+            "The oldest stSOL/UST price was sampled at slot {}. \
+            It must have been sampled more recently.",
+            oldest_sample.slot,
+        );
+        return Err(AnkerError::FetchPoolPriceNotCalledRecently.into());
+    }
```
ruuda, line 324: @enriquefynn I think we also need to enforce a minimum age on the youngest sample. Otherwise you can do the final update and `sell_rewards` in the same transaction. I’ll make a pull request for that.

ruuda: https://github.com/ChorusOne/solido/pull/549

enriquefynn: Great find! Added comments in the PR

PR #512, opened 2022-01-28
Anker: Check the token swap program id before we call it

ruuda:
This fixes a critical issue identified by Neodyme: we never checked the program id of the token swap program before calling it with `invoke_signed` in `SellRewards`. An attacker could call `SellRewards` and pass in their own patched version of the token swap program, which would then be able to steal the full stSOL reserve.

* Add a regression test that would have caught the missing check.
* Fix the bug by checking that the token swap program passed in to the call is the owner of the token swap pool stored in the Anker instance. The token swap pool can only be set by the manager, who we trust to set the right pool that is owned by the right token swap program.
* Add an additional check in `SellRewards` that checks that we don’t lose more stSOL than we intended to sell. If the owner check were not in place, this check would not have prevented an attacker from stealing some stSOL, but it would at least have limited the damage to stealing only the excess stSOL that we intended to sell, not the full stSOL reserve.

### DIFF
--- a/anker/src/processor.rs
+++ b/anker/src/processor.rs
@@ -250,10 +250,10 @@ fn process_sell_rewards(program_id: &Pubkey, accounts_raw: &[AccountInfo]) -> Pr
         StLamports(Anker::get_token_amount(accounts.st_sol_reserve_account)?);
 
     // Get StLamports corresponding to the amount of b_sol minted.
-    let st_sol_amount = solido.exchange_rate.exchange_sol(Lamports(b_sol_supply))?;
+    let b_sol_supply_value_in_st_sol = solido.exchange_rate.exchange_sol(Lamports(b_sol_supply))?;
 
-    // If `reserve_st_sol` < `st_sol_amount` something went wrong, and we abort the transaction.
-    let rewards = (reserve_st_sol_before - st_sol_amount)?;
+    // If this underflows, something went wrong, and we abort the transaction.
+    let rewards = (reserve_st_sol_before - b_sol_supply_value_in_st_sol)?;
 
     // Get the amount of UST that we had.
     let ust_before = MicroUst(Anker::get_token_amount(accounts.ust_reserve_account)?);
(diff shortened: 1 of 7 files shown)


anker/src/processor.rs
```diff
@@ -246,20 +246,36 @@ fn process_sell_rewards(program_id: &Pubkey, accounts_raw: &[AccountInfo]) -> Pr
...
     let ust_after = MicroUst(Anker::get_token_amount(accounts.ust_reserve_account)?);
+    let reserve_st_sol_after =
+        StLamports(Anker::get_token_amount(accounts.st_sol_reserve_account)?);
     let swapped_ust = (ust_after - ust_before)?;
+    let swapped_st_sol = (reserve_st_sol_before - reserve_st_sol_after)?;
+
+    // The token swap program should not take more stSOL than we told it to swap.
+    // As an extra line of defense, confirm this after the swap is done, and abort
+    // if some stSOL went missing.
+    if swapped_st_sol > st_sol_amount {
```
enriquefynn, line 271: Thanks for the check, this would have minimized the impact of the bug.

anker/tests/tests/sell_rewards.rs
```diff
@@ -70,3 +70,18 @@ async fn test_rewards_fail_with_different_reserve() {
     let result = context.try_sell_rewards().await;
     assert_solido_error!(result, AnkerError::InvalidDerivedAccount);
 }
+
+#[tokio::test]
+async fn test_sell_rewards_fails_with_different_token_swap_program() {
```
enriquefynn, line 75: Great to add a test for it. Thanks ❤️

enriquefynn (approved):
Great fix! Thanks for working on it, LGTM!

anker/src/processor.rs
ruuda: This check is incorrect! The swap amount is `rewards`, not `st_sol_amount`!

enriquefynn: Yes, this is embarrassing. Sorry about that, need to be more careful.

PR #515, opened 2022-02-02
Answer a request for an interval_price query

enriquefynn:
WIP for answering queries from clients.
Answers the query /interval_price?from=<from_date>?to=<to_date>.


cli/listener/src/main.rs
```diff
@@ -394,7 +402,169 @@ impl Metrics {
...
+        if method_name == "interval_price" {
+            match get_date_params(&vec_parameters[1..]) {
+                Ok(res) => Some(res),
+                Err(err) => return Some(get_error_response(err)),
+            }
+        } else if method_name.is_empty() {
+            None
+        } else {
+            return Some(get_error_response(ResponseError {
+                error: format!("Method not supported: {}, use \"interval_price?from_date=<from_date_iso8601>?to_date=<to_date_iso8601>\"", method_name)
```
ruuda, line 512: Query params tend to have `?` to separate the path part from the query params, and then `&` to separate different query params, not a second `?`, e.g.

```
/interval_price?from_date=X&to_date=Y
```

cli/listener/src/main.rs
```diff
@@ -392,7 +399,169 @@ impl Metrics {
...
+}
+
+/// Get an interval price, consume it and returns a `ResponseBox` with the
+/// provided interval price and computed annual percentage rate.
+fn get_success_response(interval_prices: IntervalPrices) -> ResponseBox {
+    let response_interval = ResponseInterval {
+        annual_percentage_rate: interval_prices.annual_percentage_rate(),
+        interval_prices,
+    };
+    let content_type = Header::from_bytes(&b"Content-Type"[..], &b"application/json"[..])
```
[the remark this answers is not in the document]
ruuda: You added this to the function above, but not this one.

cli/listener/src/main.rs
```diff
@@ -392,7 +403,154 @@ impl Metrics {
...
+    let parsed_url = match Url::parse(request_url) {
+        Ok(parsed_url) => parsed_url,
+        Err(err) => {
+            return Some(get_error_response(ResponseError {
+                error: err.to_string(),
+            }))
+        }
+    };
+    let method_name = parsed_url.path_segments()?.last()?;
+    if method_name != "interval_price" {
```
ruuda, line 487: Bikeshed: shall we call it `apy`? That’s what this service is for after all :D

cli/listener/src/main.rs
```diff
@@ -392,7 +403,154 @@ impl Metrics {
...
+        return Some(get_error_response(ResponseError {
+                        error: format!("Method not supported: {}, use \"/interval_price?from=<from_date_iso8601>&to=<to_date_iso8601>\"", method_name)
```
ruuda, line 489: Also we now use `from` and `to` here, but `begin` and `end` in the response. We should be consistent, let’s rename these to `begin` and `end`.

cli/listener/src/main.rs
```diff
@@ -392,7 +403,154 @@ impl Metrics {
...
+                // Error while getting the interval prices.
+                Err(err) => Some(get_error_response(ResponseError {
+                    error: err.to_string(),
+                })),
+                Ok(interval_prices_opt) => {
+                    if let Some(interval_prices) = interval_prices_opt {
+                        // Got interval prices.
+                        Some(get_success_response(interval_prices))
+                    } else {
+                        // No interval price could be calculated, probably because of few data points.
```
ruuda, line 521: If we have a single data point, shall we return a success response with `APY` set to `0`, and the `begin_*` and `end_*` values set to the same value? This way the service doesn’t return errors for an entire epoch, and then we can use this endpoint for health checks.

cli/listener/src/main.rs
```diff
@@ -392,7 +403,154 @@ impl Metrics {
...
-fn serve_request(request: Request, metrics_mutex: &MetricsMutex) -> Result<(), std::io::Error> {
+#[derive(Serialize, PartialEq, Debug)]
+struct ResponseError {
+    error: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+struct ResponseInterval {
+    interval_prices: IntervalPrices,
+    annual_percentage_rate: f64,
```
ruuda, line 414:
```suggestion
    annual_percentage_yield: f64,
```

[According to Investopedia](https://www.investopedia.com/terms/a/apr.asp#toc-apr-vs-annual-percentage-yield-apy) the value we return is an APY, not an APR, because we do take compounding into account.

cli/listener/src/main.rs
```diff
@@ -531,81 +694,168 @@ fn main() {
...
+        };
+        insert_price(&conn, &exchange_rate).unwrap();
+        let apy = get_interval_price_for_period(
+            conn.unchecked_transaction().unwrap(),
+            chrono::Utc.ymd(2020, 7, 7).and_hms(0, 0, 0),
+            chrono::Utc.ymd(2021, 7, 8).and_hms(0, 0, 0),
+            SOLIDO_ID.to_owned(),
+        )
+        .expect("Failed when getting APY for period");
+        assert_eq!(apy.unwrap().annual_percentage_rate(), 4.7989255185326485);
```
ruuda, line 733:
```suggestion
        assert_eq!(apy.unwrap().annual_percentage_yield(), 4.7989255185326485);
```
We should also rename this for consistency, see also my earlier comment.

cli/listener/src/main.rs
```diff
@@ -531,81 +694,168 @@ fn main() {
...
+            dates,
+            Ok(parse_utc_iso8601("2022-02-04T11:40:02.683960+00:00")
+                ..parse_utc_iso8601("2022-02-07T14:22:08.826526+00:00")),
+        );
+    }
+
+    #[test]
+    fn test_get_interval_from_url() {
+        // check for correct and specific prices.
+        fn check_correct_url(conn: &Connection, url: &str) {
```
ruuda, line 798: It might be a bit nicer to split this into two tests, one that tests parsing of the url query params, and a different one that tests range selection. Then it’s easier to to see which part is broken when a test fails, and if all of the individual pieces pass the tests, we assume that the combination is good as well. (It is also good to test the full flow, but in that case I would try to test it externally, i.e. start the actual binary with some prepared database, make an http request to it, and verify the response. For our tiny server I think that’s overkill though, just the unit tests and some manual testing will do for now.)

ruuda (commented):
I have one suggestion for making the query params parsing more robust. Aside from that I only have nits.

cli/listener/src/main.rs
```diff
@@ -392,7 +403,154 @@ impl Metrics {
...
+
+/// Returns a `DateFromTo` with the Result from parsing the iso8601 from
+/// `input`. `input` is the url query in bytes.
+fn get_date_params(parsed_inputs: Vec<(Cow<str>, Cow<str>)>) -> Result<DateFromTo, ResponseError> {
+    let (first_k, first_v) = parsed_inputs.get(0).ok_or(ResponseError {
+        error: "First parameter \"from\" or \"to\" not provided.".to_owned(),
+    })?;
+    let (second_k, second_v) = parsed_inputs.get(1).ok_or(ResponseError {
+        error: "Second parameter \"from\" or \"to\" not provided.".to_owned(),
+    })?;
```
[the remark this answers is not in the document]
enriquefynn: That's a nice approach, thanks!

cli/listener/src/main.rs
```diff
@@ -392,7 +403,154 @@ impl Metrics {
...
+
+/// Returns a Request response with an error.
+/// With the status code set to 400.
+fn get_error_response(err_res: ResponseError) -> ResponseBox {
+    let content_type = Header::from_bytes(
+        &b"Content-Type"[..],
+        &b"application/json; charset=UTF-8"[..],
+    )
+    .expect("Static header value, does not fail at runtime.");
+    Response::from_data(serde_json::to_vec(&err_res).expect("Serialization shouldn't fail"))
```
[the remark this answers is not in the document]
enriquefynn: What do you suggest? We return just the message or nothing?

cli/listener/src/main.rs
enriquefynn: Yeah, can be. Since we're going to pre-populate the db with the data for when Solido was first deployed I think it wouldn't make a practical difference, but I changed it. Now it fails only if no data point is present.

cli/listener/src/main.rs
enriquefynn: Why should we split this one? It only tests the successful queries. We can make another that tests for parsing errors.
We can also just delete this test, but I see no reason why, now that it is done 😬

ruuda: This test tests two things at once: parsing the query params, and running the query. We can test two smaller units independently. That helps to pinpoint a problem when one of them fails. But I don’t have a strong opinion on it.

cli/listener/src/main.rs
```diff
@@ -392,7 +405,170 @@ impl Metrics {
...
+            }
+            _ => continue,
+        }
+    }
+
+    let begin = match begin_opt {
+        Some(t) => t,
+        None => {
+            return Err(ResponseError::new_bad_request(
+                "Missing query parameter: 'from'.",
```
ruuda, line 459:
```suggestion
                "Missing query parameter: 'begin'.",
```

cli/listener/src/main.rs
```diff
@@ -392,7 +405,170 @@ impl Metrics {
...
+            ))
+        }
+    };
+    let end = match end_opt {
+        Some(t) => t,
+        None => {
+            return Err(ResponseError::new_bad_request(
+                "Missing query parameter: 'to'.",
```
ruuda, line 467:
```suggestion
                "Missing query parameter: 'end'.",
```

cli/listener/src/main.rs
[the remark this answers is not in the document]
ruuda: Yeah I would do just `text/plain`, and the message or `"internal server error"`.

cli/listener/src/main.rs
```diff
@@ -392,7 +405,170 @@ impl Metrics {
...
+/// Returns a Request response with an error.
+/// With the status code set to 400.
+fn get_error_response(err_res: ResponseError) -> ResponseBox {
+    let content_type = Header::from_bytes(
+        &b"Content-Type"[..],
+        &b"application/json; charset=UTF-8"[..],
+    )
+    .expect("Static header value, does not fail at runtime.");
+    Response::from_data(serde_json::to_vec(&err_res).expect("Serialization shouldn't fail"))
+        .with_status_code(400)
```
ruuda, line 483: Should be 500 for internal server error.

cli/listener/src/main.rs
```diff
@@ -392,7 +405,170 @@ impl Metrics {
...
+}
+
+fn serve_request(
+    db_connection: &Connection,
+    request: Request,
+    metrics_mutex: &MetricsMutex,
+) -> Result<(), std::io::Error> {
+    if let Some(res) = get_interval_price_request(db_connection, request.url()) {
+        return request.respond(res);
+    };
```
ruuda, line 570: Hmm, so now we serve the APY response on any url that ends with `/apy`, and the metrics on anything else. I think it would be safer to turn that around, and serve the metrics only on `/metrics`.

cli/listener/src/main.rs
enriquefynn: Yes, but parsing the query is tested already when testing `get_date_params`.

cli/listener/src/main.rs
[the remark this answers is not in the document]
enriquefynn: Sometimes is a formatting error from the client, can we do instead:
```
match err_res {
        ResponseError::BadRequest(_msg) => {
            let content_type = Header::from_bytes(
                &b"Content-Type"[..],
                &b"application/json; charset=UTF-8"[..],
            )
            .expect("Static header value, does not fail at runtime.");
            Response::from_data(serde_json::to_vec(&err_res).expect("Serialization shouldn't fail"))
                .with_status_code(400)
                .with_header(content_type)
                .boxed()
        }
        ResponseError::InternalServerError => {
            let content_type =
                Header::from_bytes(&b"Content-Type"[..], &b"text/plain; charset=UTF-8"[..])
                    .expect("Static header value, does not fail at runtime.");
            Response::from_string("internal server error")
                .with_status_code(500)
                .with_header(content_type)
                .boxed()
        }
    }
```
Or return every error as `"internal server error"`?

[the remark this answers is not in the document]
ruuda: Yes your snippet above is what I meant! (Or alternatively, in the `BadRequest`, we could omit the json wrapper and just send the text, either is fine with me.)

cli/listener/src/main.rs
```diff
@@ -506,93 +506,109 @@ fn get_success_response(interval_prices: IntervalPrices) -> ResponseBox {
...
         }
         Ok(interval_prices_opt) => {
             if let Some(interval_prices) = interval_prices_opt {
                 // Got interval prices.
-                Some(get_success_response(interval_prices))
+                get_success_response(interval_prices)
             } else {
                 // No interval price could be calculated, probably because of few data points.
-                Some(get_error_response(ResponseError::new_bad_request(
+                get_error_response(ResponseError::new_bad_request(
```
ruuda, line 535: This is not really a bad request then, it’s not something that the client can do something about. Not sure what the best thing to return is here. Maybe 404. But let’s get this pull request merged first, we can tweak this later.

cli/listener/src/main.rs
```diff
@@ -506,93 +506,109 @@ fn get_success_response(interval_prices: IntervalPrices) -> ResponseBox {
...
+        "apy" => {
+            let query_params =
+                form_urlencoded::parse(parsed_url.query()?.as_bytes()).collect::<Vec<_>>();
+            let dates = match get_date_params(query_params) {
+                Ok(res) => res,
+                Err(err) => return Some(ParseUrlResponse::Error(err)),
+            };
+            Some(ParseUrlResponse::IntervalPriceRequest(dates))
+        }
+        "metrics" => Some(ParseUrlResponse::Metrics),
```
ruuda, line 565: Now we will expose metrics on e.g. `/foo/bar/metrics`. If the reverse proxy does a blanked forward of `/api/apy*` -> `our-apy-daemon/api/apy/*`, it means that users can access the metrics on `/api/apy/metrics`. If we only accept `/metrics`, then we can keep them internal.

ruuda (approved):
Let’s merge this, we can tweak the routing and status codes in a follow-up PR.

PR #517, opened 2022-02-04
Add version byte for Anker

### LINKED ISSUE #510: Add version byte to the `Anker` struct
Just like in the `Lido` state struct, which starts with a version byte, we should start the `Anker` struct with a version byte, so that parsers that extract the account data can bail out in case a future version changes the data layout.

enriquefynn:
Adds a version byte for the Anker state in the same way it's done for Solido.
Closes #510

### DIFF
--- a/anker/src/state.rs
+++ b/anker/src/state.rs
@@ -22,7 +22,8 @@ use spl_token_swap::state::SwapV1;
 use crate::token::{self, BLamports};
 
 /// Size of the serialized [`Anker`] struct, in bytes.
-pub const ANKER_LEN: usize = 233;
+pub const ANKER_LEN: usize = 234;
+pub const ANKER_VERSION: u8 = 0;
 
 #[repr(C)]
 #[derive(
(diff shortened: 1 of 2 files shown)


anker/src/state.rs
```diff
@@ -40,6 +41,9 @@ pub struct WormholeParameters {
     Clone, Debug, Default, BorshDeserialize, BorshSerialize, BorshSchema, Eq, PartialEq, Serialize,
 )]
 pub struct Anker {
+    /// Version number for Anker.
+    pub version: u8,
```
ruuda, line 45: I wondered where it got set, but then I realized it’s the `Default` impl. That’s a bit subtle, can we initialize it explicitly in `process_initialize`?

ruuda (approved):
LGTM with one suggestion!

Also I wonder now, we do check the Solido version byte before we deserialize it, right? We need to double-check.

anker/src/state.rs
ruuda: Never mind, it is there, I missed it. LGTM then!

ruuda (approved):

PR #519, opened 2022-02-08
Added anker related functions to solido.js

hritique:


js/tsconfig.json
```diff
@@ -13,7 +13,7 @@
     "strictNullChecks": true,
     "strictFunctionTypes": true,
     "noImplicitThis": true,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
```
ruuda, line 16: Why does it need to be allowed?

js/src/ankerSnapshot.ts
```diff
@@ -0,0 +1,206 @@
...
+
+      Object.assign(this, parsedData);
+    } catch (error) {
+      console.log('error', error);
+    }
+  }
+}
+
+class Metrics {
+  swapped_rewards_st_sol_total: BN;
```
ruuda, line 72: These numbers are in lamports and µUST, I would rename them to `swapped_rewards_st_lamports_total` and `swapped_rewards_micro_ust_total`, if we can’t use strong types here.

js/src/ankerSnapshot.ts
```diff
@@ -0,0 +1,206 @@
...
+    ],
+    programAddresses.ankerProgramId
+  );
+
+  accountsInfoRecord[stSolReserveAccountAddress.toString()] = null;
+
+  // Iterates over all the required accounts data to be fetched and updates the accountsInfo record
+  // Once, all we have the record of all the accounts data, we can create the snapshot
+  // This way, we avoid torn reads, and fetch all the required data in one go with `getMultipleAccountInfo`
+  while (true) {
```
ruuda, line 160: I don’t see a `continue` or `break` in this loop, it looks like it will either throw or return in the first iteration.

If you want to mimick what we do in https://github.com/ChorusOne/solido/blob/3faa01864a444197324d209b4c2810f203781022/cli/common/src/snapshot.rs, I would do something like:

* In the functions that read accounts, instead of getting them directly from the `connection`, read them from a hash map. If the account is not part of the hashmap, throw `AccountsMissingException(List[Pubkey])`.
* In the loop, catch this, add the missing accounts to the ones to fetch in the `GetMultipleAccounts` call, and then `continue`.

js/src/instructions/anker/withdraw.ts
```diff
@@ -0,0 +1,90 @@
+import * as BufferLayout from '@solana/buffer-layout';
+import { TOKEN_PROGRAM_ID } from '@solana/spl-token';
+import { PublicKey, TransactionInstruction } from '@solana/web3.js';
+import { BLamports, ProgramAddresses } from '../../types';
+
+/**
+ * Returns the instruction to withdraw stSOL from the Anker program
+ * @param senderBSolAccountOwnerAddress Address of the owner of the sender's bSOL SPL token account
+ * @param senderBSolAccountAddress Address of the bSOL SPL token account, whose bSOL balance will be decreased
+ * @param recipientStSolAccountAddress Address of the recipient bSOL SPL token account
```
ruuda, line 10:
```suggestion
 * @param recipientStSolAccountAddress Address of the recipient stSOL SPL token account, whose stSOL balance will be increased
```

js/src/instructions/anker/deposit.ts
```diff
@@ -0,0 +1,90 @@
+import * as BufferLayout from '@solana/buffer-layout';
+import { TOKEN_PROGRAM_ID } from '@solana/spl-token';
+import { PublicKey, TransactionInstruction } from '@solana/web3.js';
+import { ProgramAddresses, StLamports } from '../../types';
+
+/**
+ * Returns the instruction to deposit stSOL into the Anker program and mint bSOL tokens
+ * @param senderStSolAccountOwnerAddress Address of the owner of the sender's stSOL SPL token account
```
ruuda, line 8:
```suggestion
 * @param senderStSolAccountOwnerAddress Address of the owner of the sender's stSOL SPL token account. Must be a signer of the transaction.
```

js/src/instructions/anker/withdraw.ts
```diff
@@ -0,0 +1,90 @@
+import * as BufferLayout from '@solana/buffer-layout';
+import { TOKEN_PROGRAM_ID } from '@solana/spl-token';
+import { PublicKey, TransactionInstruction } from '@solana/web3.js';
+import { BLamports, ProgramAddresses } from '../../types';
+
+/**
+ * Returns the instruction to withdraw stSOL from the Anker program
+ * @param senderBSolAccountOwnerAddress Address of the owner of the sender's bSOL SPL token account
```
ruuda, line 8:
```suggestion
 * @param senderBSolAccountOwnerAddress Address of the owner of the sender's bSOL SPL token account. Must be a signer of the transaction.
```

js/src/ankerSnapshot.ts
hritique: Yeah this is because in this case, we have all the accounts' addresses already in the `accountsInfo` record/hashmap before the loop starts.
So it fetches all the data in the first iteration and returns the snapshot so no checking and `continue`.

ruuda: But then why is the loop there?

js/src/ankerSnapshot.ts
```diff
@@ -95,7 +105,7 @@ const schema = new Map([
       kind: 'struct',
       fields: [
         ['swapped_rewards_st_sol_total', 'u64'],
```
ruuda, line 107: This should be `st_lamports`, right?

hritique: I've now used the type StLamports for this. Should it still be renamed?

js/src/ankerSnapshot.ts
hritique: Yeah that is there just to follow the process we followed for the Solido Snapshot. We can remove it or probably leave it there for future(any other property needs to be added to the snapshot that requires multi-step reads)

ruuda: It does nothing right now, let’s remove it, we can add it again if it’s needed.

ruuda:
Anchor died, this is no longer relevant.

PR #523, opened 2022-02-18
Return message when no data point could be calculated

enriquefynn:
As @ruuda pointed out in #515, we shouldn't return an error when no data points are present for calculating the APY. We instead should return a message with 404 status code.


ruuda (commented):
Apologies for the long delay. It looks good, but it has a conflict now. Since this PR we introduced `ResponseError::NotFound`, so we can use that.

PR #524, opened 2022-02-21
Limit queries

### LINKED ISSUE #516: Limit queries to select `epoch + offset`
Limit queries on `get_interval_price_for_period` to select the minimum/maximum epoch that starts/begins from the given timestamps.

To give time for the exchange rate to be updated, we shall introduce an offset that we can sum to the database entries to select the n-th entry for a given epoch.

enriquefynn:
Limit query for `const QUERY_INTERVAL_OFFSET: usize = 200` from the
beginning of the epoch.
This is done so the exchange rate could be updated in time.

Close #516

### DIFF
--- a/cli/listener/src/main.rs
+++ b/cli/listener/src/main.rs
@@ -212,9 +212,7 @@ pub fn get_interval_price_for_period(
             "SELECT MAX(epoch) from exchange_rate where pool = :pool and timestamp < :t",
         )?;
         let epoch = match last_epoch_stmt
-            .query_row([pool.clone(), to_time.to_rfc3339()], |row| {
-                row.get::<usize, u64>(0)
-            })
+            .query_row([pool, to_time.to_rfc3339()], |row| row.get::<usize, u64>(0))
             .ok()
         {
             Some(epoch) => epoch,
--- a/cli/listener/src/main.rs
+++ b/cli/listener/src/main.rs
@@ -191,41 +191,41 @@ pub fn get_interval_price_for_period(
         let epoch = match first_epoch_stmt
             .query_row([pool.clone(), from_time.to_rfc3339()], |row| {
                 row.get::<usize, u64>(0)
-            })
-            .ok()
-        {
-            Some(epoch) => epoch,
-            None => return Ok(None),
+            }) {
+            Ok(epoch) => epoch,
+            Err(_) => return Ok(None),
         };
         let minimum_slot = epoch_schedule.get_first_slot_in_epoch(epoch) + QUERY_SLOT_OFFSET;
 
         // Get the first row from `epoch` in which the slot is greater than `minimum_slot`.
         let mut exchange_rate_stmt = tx.prepare(
-            "SELECT * from exchange_rate WHERE epoch = :epoch AND slot >= :slot_min LIMIT 1",
+            "SELECT * from exchange_rate WHERE pool = :pool AND epoch = :epoch AND slot >= :slot_min LIMIT 1",
         )?;
         let first_exchange_rate = exchange_rate_stmt
-            .query_map([epoch, minimum_slot], row_map)?
+            .query_map(
+                [pool.clone(), epoch.to_string(), minimum_slot.to_string()],
+                row_map,
+            )?
             .next();
 
         // Get maximum epoch in which timestamp is smaller than `from_time`.
         let mut last_epoch_stmt = tx.prepare(
             "SELECT MAX(epoch) from exchange_rate where pool = :pool and timestamp < :t",
         )?;
-        let epoch = match last_epoch_stmt
-            .query_row([pool, to_time.to_rfc3339()], |row| row.get::<usize, u64>(0))
-            .ok()
-        {
-            Some(epoch) => epoch,
-            None => return Ok(None),
+        let epoch = match last_epoch_stmt.query_row([pool.clone(), to_time.to_rfc3339()], |row| {
+            row.get::<usize, u64>(0)
+        }) {
+            Ok(epoch) => epoch,
+            Err(_) => return Ok(None),
         };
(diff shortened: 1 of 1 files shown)


cli/listener/src/main.rs
```diff
@@ -27,8 +27,12 @@ use std::str::FromStr;
 use tiny_http::{Header, Request, Response, ResponseBox, Server};
 use url::{form_urlencoded, Url};
 
+// Name put in the solido table `pool`.
 const SOLIDO_ID: &str = "solido";
 
+// Offset from the first Epoch's entry to use as data point.
```
ruuda, line 33: Ideally, we would offset it from the start of the epoch, not from the first entry. There is probably some code in `solana_program::clock` that we can use to compute it. What do you think?

**Edit**: Also, it wasn’t clear to me that this was about rows, not an offset in slots. We shouldn’t do it on rows because the poll interval is configurable.

enriquefynn:
Can you please take another look @ruuda. Maybe there are some more idiomatic ways to do the modifications proposed.
I also think that we should get the timestamp from the cluster, but we can do that later.

ruuda:
> I also think that we should get the timestamp from the cluster

Unfortunately it cannot be trusted as long as the max-allowable-drift is in place (https://github.com/solana-labs/solana/issues/18992). The time on the machine itself is under our control, so we assume it to be trusted.

cli/listener/src/main.rs
```diff
@@ -165,46 +170,55 @@ pub fn get_interval_price_for_period(
...
-            ",
+        // Get minimum epoch in which timestamp is greater than `from_time`.
+        let mut first_epoch_stmt = tx.prepare(
+            "SELECT MIN(epoch) from exchange_rate where pool = :pool and timestamp > :t",
+        )?;
+        let epoch = match first_epoch_stmt
+            .query_row([pool.clone(), from_time.to_rfc3339()], |row| {
+                row.get::<usize, u64>(0)
+            })
+            .ok()
```
ruuda, line 185: We can skip the `.ok()` here and match directly on `Ok(epoch)` / `Err(..) => return Ok(None)` below.

cli/listener/src/main.rs
```diff
@@ -165,46 +170,55 @@ pub fn get_interval_price_for_period(
...
+        {
+            Some(epoch) => epoch,
+            None => return Ok(None),
+        };
+        let minimum_slot = epoch_schedule.get_first_slot_in_epoch(epoch) + QUERY_SLOT_OFFSET;
```
ruuda, line 190: Nice!

cli/listener/src/main.rs
```diff
@@ -165,46 +170,55 @@ pub fn get_interval_price_for_period(
...
+
+        // Get the first row from `epoch` in which the slot is greater than `minimum_slot`.
+        let mut exchange_rate_stmt = tx.prepare(
+            "SELECT * from exchange_rate WHERE epoch = :epoch AND slot >= :slot_min LIMIT 1",
```
ruuda, line 194: This query is missing a filter on `AND pool = :pool`. For now we only have a single pool, but if we ever add a second one, this might return a row for a different pool than the one we tried to query.

cli/listener/src/main.rs
```diff
@@ -165,46 +170,55 @@ pub fn get_interval_price_for_period(
...
+        )?;
+        let first_exchange_rate = exchange_rate_stmt
+            .query_map([epoch, minimum_slot], row_map)?
+            .next();
+
+        // Get maximum epoch in which timestamp is smaller than `from_time`.
```
ruuda, line 200:
```suggestion
        // Get maximum epoch in which timestamp is smaller than `to_time`.
```

cli/listener/src/main.rs
```diff
@@ -165,46 +170,55 @@ pub fn get_interval_price_for_period(
...
+            .ok()
+        {
+            Some(epoch) => epoch,
+            None => return Ok(None),
+        };
+        let minimum_slot = epoch_schedule.get_first_slot_in_epoch(epoch) + QUERY_SLOT_OFFSET;
+
+        // Get the first row from `epoch` in which the slot is greater than `minimum_slot`.
+        let mut exchange_rate_stmt = tx.prepare(
+            "SELECT * from exchange_rate WHERE epoch = :epoch AND slot >= :slot_min LIMIT 1",
```
ruuda, line 215: This one also needs a filter on the pool.

cli/listener/src/main.rs
```diff
@@ -165,46 +170,55 @@ pub fn get_interval_price_for_period(
...
-
-        let mut row_iter = stmt_last.query_map([pool, to_time.to_rfc3339()], row_map)?;
-        let last = row_iter.next();
-
-        (first, last)
+        let last_exchange_rate = exchange_rate_stmt
+            .query_map([epoch, minimum_slot], row_map)?
+            .next();
+
+        (first_exchange_rate, last_exchange_rate)
```
ruuda, line 221: To factiltate these queries, we should create an index either on `timestamp` and `slot`, or on `(pool, timestamp)` and `(pool, slot)`. Otherwise it will need to do a full table scan every time. Probably the overhead of the extra column in the index is not worth scanning through a few extra rows to find a matching `pool`, I would keep it simple and index just `timestamp` and `slot`.

We can add this to the startup:

```sql
CREATE INDEX IF NOT EXISTS ix_exchange_rate_timestamp ON exchange_rate (timestamp);
CREATE INDEX IF NOT EXISTS ix_exchange_rate_slot ON exchange_rate (slot);
```

ruuda (commented):
The logic looks almost good to me! Only a filter on `pool` is missing. Also, we should add indexes to facilitate these queries.

cli/listener/src/main.rs
enriquefynn: Yeah, great idea :)

ruuda (approved):
LGTM

PR #525, opened 2022-02-21
Make /metrics internal.

### LINKED ISSUE #522: [APY daemon] Make /metrics internal
As @ruuda mentioned on #515

> Now we will expose metrics on e.g. /foo/bar/metrics. If the reverse proxy does a blanked forward of /api/apy* -> our-apy-daemon/api/apy/*, it means that users can access the metrics on /api/apy/metrics. If we only accept /metrics, then we can keep them internal.

enriquefynn:
Only answers /metrics if is in the form `hostname:port`/metrics.
Making it internal if we rewrite the public url with `../api/..`

Closes #522


ruuda (approved):
LGTM! It’s a bit of an unusual formulation but it should do the trick. :+1:

PR #528, opened 2022-02-24
Apy daemon query params

hritique:
**Changes:**
- Added 2 query params to add support for queries like `/apy?days=5` and `/apy?since_launch`.

**Question:**
Currently the program loops over all the query parameters and overrides the `begin_opt` and `end_opt` on the way.
So if I do `/apy?days=2&since_launch`, it will consider begin and end values based on `since_launch` parameter.

Is this expected behaviour, or should we break out of the loop once we have valid `begin_opt` and `end_opt` values?

### DIFF
--- a/cli/listener/src/main.rs
+++ b/cli/listener/src/main.rs
@@ -6,6 +6,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+use chrono::TimeZone;
 use clap::Clap;
 use lido::token::Rational;
 use rand::{rngs::ThreadRng, Rng};
@@ -468,6 +469,13 @@ fn get_date_params<'a, I: IntoIterator<Item = (Cow<'a, str>, Cow<'a, str>)>>(
                     Expected e.g. '30'.",
                         )
                     })?;
+                begin_opt = Some(begin);
+                end_opt = Some(end);
+            }
+            "since_launch" => {
+                let begin = chrono::Utc.ymd(2021, 09, 01).and_hms(00, 00, 00); // Solido Launch Date
+                let end = chrono::Utc::now();
+
                 begin_opt = Some(begin);
                 end_opt = Some(end);
             }
--- a/cli/listener/src/main.rs
+++ b/cli/listener/src/main.rs
@@ -452,6 +452,25 @@ fn get_date_params<'a, I: IntoIterator<Item = (Cow<'a, str>, Cow<'a, str>)>>(
                 })?;
                 end_opt = Some(t);
             }
+            "days" => {
+                let days = v.parse::<i64>().map_err(|_| {
+                    ResponseError::BadRequest(
+                        "Invalid number of days in 'days' query parameter. \
+                    Expected e.g. '30'.",
+                    )
+                })?;
+                let end = chrono::Utc::now();
+                let begin = end
+                    .checked_sub_signed(chrono::Duration::days(days))
+                    .ok_or_else(|| {
+                        ResponseError::BadRequest(
+                            "Invalid number of days in 'days' query parameter. \
+                    Expected e.g. '30'.",
+                        )
+                    })?;
+                begin_opt = Some(begin);
+                end_opt = Some(end);
+            }
             _ => continue,
         }
     }


ruuda:
> Currently the program loops over all the query parameters and overrides the `begin_opt` and `end_opt` on the way.
> So if I do `/apy?days=2&since_launch`, it will consider begin and end values based on `since_launch` parameter.

That’s a fair point. It’s a bit of a footgun when you pass multiple query params that all try to set the bounds. “Last write wins” sounds reasonable to me, but even better would be to disallow it. What we could do:

* Before we set `begin_opt` or `end_opt`, check if they are already set, and if so, return `ResponseError::BadRequest` with an explanation that you should only provide one of `days`, `since_launch`, and `begin`/`end`.
* Maybe a slightly nicer way to do that is to change `begin_opt` and `end_opt` to `begin: Vec<DateTime<Utc>>` and `end: Vec<DateTime<Utc>>` and push the value on to it while we loop over the query params. Then after we processed all of them, we check:
   * If `begin.len() != 1`: Return error that says that exactly one of `begin`, `days` or `since_launch` is needed.
   * If we didn’t return an error, then we have a unique value for `begin`.
   * Same for `end`.

ruuda:
CI is failing on the `cargo clippy` linter step: https://github.com/ChorusOne/solido/runs/5322085952?check_suite_focus=true#step:5:68. You can also run `cargo clippy` locally to see these warnings locally.

cli/listener/src/main.rs
```diff
@@ -452,6 +453,32 @@ fn get_date_params<'a, I: IntoIterator<Item = (Cow<'a, str>, Cow<'a, str>)>>(
...
                 end_opt = Some(t);
             }
+            "days" => {
+                let days = v.parse::<i64>().map_err(|_| {
+                    ResponseError::BadRequest(
+                        "Invalid number of days in 'days' query parameter. \
+                    Expected e.g. '30'.",
+                    )
+                })?;
+                let end = chrono::Utc::now();
```
ruuda, line 463: If we call `now()` in this function, then this function will behave differently every time you call it: the return value will be different, even if the arguments are the same. This makes the function [impure](https://en.wikipedia.org/wiki/Pure_function). Impure functions are difficult to test, because their output is a moving target. Pure functions are much nicer to test, because you can just write assertions of the form `assert_eq(f(inputs), expected_output);`.

Aside from that, if we call `now()` directly, we may call it in multiple places and end up with inconsistencies. It’s not the case here, but imagine we had `begin_days_ago` and `end_days_ago` query params that set `begin` and `end` to `now().checked_sub_signed(begin_days_ago)` and `now().checked_sub_signed(end_days_ago)`. I would expect, if I pass `begin_days_ago=1&end_days_ago=0` that `end - begin` is a duration of exactly one day. But this is not the case! Between the first call to `now()`, and the second call to `now()`, some time elapsed, so we end up with a duration that is more than 1 day. Again, in this case it’s no big deal, but in general this can lead to very subtle bugs, so I prefer to avoid this. (An exception to this is if you want to measure how long some operation takes, but then you should use a monotonic clock and not `DateTime<Utc>`.)

The solution to both problems is to push the impurity to the call site. Instead of calling `now()` directly, add an argument `now: DateTime<Utc>` to the function, and let the caller deal with it.

* Now our function is pure, and therefore easy to test.
* If we reference `now` in multiple places, they will be consistent.

This only moved the problem though, but you can usually continue moving the problem, and then we will end up at a very natural place to call `now()`. In this case that would be the entry point of the request handler, `serve_request`.

ruuda (commented):
I have a suggestion to make this function easier to test. If you implement that, can you also add a test for the new query params?

You can use this test for inspiration: https://github.com/ChorusOne/solido/blob/752e4b00d19d8a3a7be7a06574ea38fe8703388b/cli/listener/src/main.rs#L826-L844

cli/listener/src/main.rs
```diff
@@ -452,6 +453,32 @@ fn get_date_params<'a, I: IntoIterator<Item = (Cow<'a, str>, Cow<'a, str>)>>(
...
+                let begin = end
+                    .checked_sub_signed(chrono::Duration::days(days))
+                    .ok_or_else(|| {
+                        ResponseError::BadRequest(
+                            "Invalid number of days in 'days' query parameter. \
+                    Expected e.g. '30'.",
```
ruuda, line 469: If we get here, `days` is already a valid integer; the reason for the `checked_sub_signed` to fail is when it overflows. The reason for that is when you pass a ridiculously large number. We can improve the message by saying something like “Date range too large”.

cli/listener/src/main.rs
```diff
@@ -430,43 +432,83 @@ type DateBeginEnd = Range<chrono::DateTime<chrono::Utc>>;
...
-                begin_opt = Some(t);
+
+                begin_opt.push(t);
             }
             "end" => {
                 let t = parse_utc_iso8601(&v).map_err(|_| {
                     ResponseError::BadRequest(
                         "Invalid ISO 8601 timestamp in 'end' query parameter. \
-                    Expected e.g. '2022-02-15T23:59:59+00:00'.",
+                    Expected e.g. '2022-02-15T23:59:59Z'.",
```
enriquefynn, line 454: The previous example is still valid, right?

cli/listener/src/main.rs
```diff
@@ -4,8 +4,10 @@ use std::{
     sync::{Arc, Mutex},
     thread::JoinHandle,
     time::{Duration, Instant},
+    vec,
 };
 
+use chrono::TimeZone;
```
enriquefynn, line 10: You can remove this include from the test modules.
```rust
#[cfg(test)]
mod test {
    use super::*;
    use chrono::TimeZone;
```

enriquefynn (commented):
Looks almost good, just needs to address some of @ruuda's requests.

cli/listener/src/main.rs
```diff
@@ -430,43 +432,83 @@ type DateBeginEnd = Range<chrono::DateTime<chrono::Utc>>;
...
+                let begin = end
+                    .checked_sub_signed(chrono::Duration::days(days))
+                    .ok_or(ResponseError::BadRequest("Date range too large"))?;
+
+                begin_opt.push(begin);
+                end_opt.push(end);
+            }
+            "since_launch" => {
+                let begin = chrono::Utc.ymd(2021, 9, 1).and_hms(00, 00, 00); // Solido Launch Date
+                let end = chrono::Utc::now();
```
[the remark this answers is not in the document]
hritique: Yeah will do that

cli/listener/src/main.rs
hritique: I'm not really sure, it was giving me the error when I used the date format from the example

ruuda: It should accept both.

hritique: Yeah but it's throwing error `Invalid ISO 8601 timestamp in 'begin' query parameter. Expected e.g. '2022-02-15T23:59:59+00:00'.` for me when I try this on browser 
https://solana.testnet.fi/api/apy/apy?begin=2022-02-04T11:40:02.683960+00:00&end=2022-02-07T14:22:08.826526+00:00

ruuda: That’s because you are not url-escaping the +, :, and . in the url, try this:

https://solana.testnet.fi/api/apy/apy?begin=2022-02-04T11%3A40%3A02.683960%2B00%3A00&end=2022-02-07T14%3A22%3A08.826526%2B00%3A00

ruuda: Maybe we can add that caveat to the error message.

hritique: Ohhh, I thought browser would do that for me. 

> Maybe we can add that caveat to the error message.

 what should be the error message?

ruuda: > what should be the error message?

I would change it from

> Invalid ISO 8601 timestamp in 'begin' query parameter. Expected e.g. '2022-02-15T23:59:59+00:00'.

to

> Invalid ISO 8601 timestamp in 'begin' query parameter. Expected e.g. '2022-02-15T23:59:59+00:00'. Note that query parameters must be url-encoded.

ruuda (approved):
LGTM!

PR #530, opened 2022-03-03
Upgrade Solana SDK to 1.9.9

enriquefynn:


ruuda (approved):
Hah, I was about to comment on `buildimage.sh`, then my diff view was outdated and it was fixed already.

LGTM

PR #531, opened 2022-03-03
Upgrade Solana to 1.8.16

enriquefynn:
The 1.9.9 and 1.10.0 version broke the `solana validator-info publish` command. I didn't investigate much but I think it's related to the wrong number of bytes or lamports when funding the Info account for it to be rent exempt.

In `solana validator-info publish` we could pass `--info-pubkey` with an already funded account owned by the Config program and with sufficient bytes allocated. However, how could we do this from the Solana client tools 😡? Of course we could write some custom function in Solido, but should we? 😬 

I reverted the upgrade to `1.9.9` until some version fixes this issue. We can upgrade to `1.8.16` so it would download the binaries from Solana's repo.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,14 +59,14 @@ jobs:
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
           target/
-        key: build-1.7.15-v2-${{ hashFiles('Cargo.lock') }}
-        restore-keys: build-1.7.15-v2
+        key: build-1.8.16-v3-${{ hashFiles('Cargo.lock') }}
+        restore-keys: build-1.8.16-v3
 
     - name: Install development tools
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -108,7 +108,7 @@ jobs:
         tests/airdrop_lamports.sh
 
         tests/test_solido.py
-        kill $validator
+        kill -9 $validator
```
ruuda, line 111: Aah, good idea! This might be why CI sometimes gets stuck!

ruuda (approved):
LGTM

.github/workflows/build.yml
enriquefynn: It's not because of this, I tried some other things with no success :|

PR #533, opened 2022-03-08
Add type/methods for `HistoricalStSolPriceArray`

enriquefynn:
Make some tests for later use on slippage

### DIFF
--- a/anker/src/state.rs
+++ b/anker/src/state.rs
@@ -952,6 +952,30 @@ mod test {
     }
 
     #[test]
+    fn test_different_prices() {
+        let mut price_array = HistoricalStSolPriceArray::new();
+        // Prices in USD per Sol [100, 90, 95, 105, 101], median: 100
+        for (slot, price) in [100, 90, 95, 105, 101].iter().enumerate() {
+            price_array.insert_and_rotate(slot as Slot, MicroUst(price * 1_000_000));
+        }
+
+        price_array.insert_and_rotate(4, MicroUst(80_000_000));
+        // prices: [90, 95, 105, 101, 80], median: 95
+        let minimum_ust = price_array
+            .calculate_minimum_price(StLamports(331_000_000_000), 5000)
+            .unwrap();
+        assert_eq!(minimum_ust, MicroUst(15_722_500_000));
+
+        price_array.insert_and_rotate(5, MicroUst(70_000_000));
+        price_array.insert_and_rotate(6, MicroUst(85_000_000));
+        // prices: [70, 80, 85, 101, 105], median: 85
+        let minimum_ust = price_array
+            .calculate_minimum_price(StLamports(100_000_000_000), 9800)
+            .unwrap();
+        assert_eq!(minimum_ust, MicroUst(8_330_000_000));
+    }
+
+    #[test]
     fn test_historical_price_array_limits() {
         let mut price_array = HistoricalStSolPriceArray::new();
         // 100 UST for each StSol.
--- a/anker/src/state.rs
+++ b/anker/src/state.rs
@@ -12,7 +12,7 @@ use crate::{
 };
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
 use lido::state::Lido;
-use lido::token::{Lamports, Rational, StLamports};
+use lido::token::{ArithmeticError, Lamports, Rational, StLamports};
 use lido::util::serialize_b58;
 use serde::Serialize;
 use solana_program::program_error::ProgramError;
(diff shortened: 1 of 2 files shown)


anker/src/state.rs
```diff
@@ -90,6 +90,74 @@ pub struct HistoricalStSolPrice {
...
+    pub fn insert_and_rotate(&mut self, slot: Slot, st_sol_price_in_ust: MicroUst) {
+        // Maintain the invariant that samples are sorted by ascending slot number.
+        // The sample at index 0 is the oldest, so we remove it (well, move it to the
+        // end to be overwritten), and move everything else closer to the beginning
+        // of the array. Then we overwrite the last element with the current price
+        // and slot number, and we confirmed above that that slot number is larger
+        // than the slot number of the sample before it.
+        self.0.rotate_left(1);
+        self.0[POOL_PRICE_NUM_SAMPLES - 1].slot = slot;
+        self.0[POOL_PRICE_NUM_SAMPLES - 1].st_sol_price_in_ust = st_sol_price_in_ust;
```
ruuda, line 135: The invariant should be preserved if we pass in the current clock, but for defensive programming, let’s explicitly assert it as well:

```rust
assert!(self.0[POOL_PRICE_NUM_SAMPLES - 1].slot >= self.0[POOL_PRICE_NUM_SAMPLES - 2]);
```

anker/src/state.rs
```diff
@@ -90,6 +90,74 @@ pub struct HistoricalStSolPrice {
...
+
+    /// Calculate the minimum amount we are willing to pay for the `StLamports`
+    /// rewards based on the median price from the historical price information.
+    pub fn calculate_minimum_price(
+        &self,
+        rewards: StLamports,
+        sell_rewards_min_out_bps: u64,
+    ) -> Result<MicroUst, ArithmeticError> {
+        // Get median historical price.
+        let median_price = self.0[POOL_PRICE_NUM_SAMPLES / 2];
```
ruuda, line 146: This is not the median price! The elements are sorted chronologically, not by ascending price!

I made a function for this, but I see I did not put it in #511. Now I remember what I was doing :sweat_smile: I wrote this program to find a function that returns the median of 5 numbers in 7 transpositions:

<details>
<summary>Program</summary>

```rust
/// Call `f` on all permutations of `candidates`.
///
/// Set `current = vec![]` at the top-level call.
/// Exits early if `f` returns `false`.
fn permutations<F: Fn(&[u32]) -> bool>(
    current: &mut Vec<u32>,
    candidates: &mut Vec<u32>,
    f: &F,
) -> bool {
    for i in 0..candidates.len() {
        let x = candidates.remove(i);
        current.push(x);
        if candidates.len() == 0 {
            let is_ok = f(&current[..]);
            if !is_ok {
                return false;
            }
        } else {
            let is_ok = permutations(current, candidates, f);
            if !is_ok {
                return false;
            }
        }
        current.pop();
        candidates.insert(i, x);
    }
    true
}

fn main() {
    for a0 in 0..5 {
    for a1 in (a0 + 1)..5 {
        for b0 in 0..5 {
        for b1 in (b0 + 1)..5 {
            for c0 in 0..5 {
            for c1 in (c0 + 1)..5 {
                for d0 in 0..5 {
                for d1 in (d0 + 1)..5 {
                    for e0 in 0..5 {
                    for e1 in (e0 + 1)..5 {
                        for f0 in 0..5 {
                        for f1 in (f0 + 1)..5 {
                            for g0 in 0..5 {
                            for g1 in (g0 + 1)..5 {
                                let is_ok = permutations(
                                    &mut vec![],
                                    &mut vec![0, 1, 2, 3, 4],
                                    &|xs| {
                                        let mut ys: Vec<u32> = xs.iter().cloned().collect();
                                        let mut cross = |i, j| {
                                            if ys[i] > ys[j] {
                                                ys.swap(i, j);
                                            }
                                        };

                                        cross(a0, a1);
                                        cross(b0, b1);
                                        cross(c0, c1);
                                        cross(d0, d1);
                                        cross(e0, e1);
                                        cross(f0, f1);
                                        cross(g0, g1);

                                        // Continue searching if this sequence of
                                        // crossings did find the median (2) of this
                                        // permutation. Otherwise abort early.
                                        ys[2] == 2
                                    },
                                );

                                if is_ok {
                                    println!("cross({}, {});", a0, a1);
                                    println!("cross({}, {});", b0, b1);
                                    println!("cross({}, {});", c0, c1);
                                    println!("cross({}, {});", d0, d1);
                                    println!("cross({}, {});", e0, e1);
                                    println!("cross({}, {});", f0, f1);
                                    println!("cross({}, {});", g0, g1);
                                    return
                                }
                            }}
                        }}
                    }}
                }}
            }}
        }}
    }}
}
```

</details>

It outputs

```rust
cross(0, 1);
cross(2, 3);
cross(0, 2);
cross(1, 3);
cross(1, 2);
cross(1, 4);
cross(2, 4);
```
where `cross(i, j)` is defined as

```rust
let mut cross = |i, j| {
    if ys[i] > ys[j] {
        ys.swap(i, j);
    }
};
```
After this, `ys[2]` contains the median. I would also add a test that generates all 5^5 possible sequences involving the numbers `0..5` and exhaustively check that the function returns the correct median by comparing it against `.sort()`.

anker/src/state.rs
```diff
@@ -836,4 +904,52 @@ mod test {
...
+            .calculate_minimum_price(StLamports(80_000_000_000), 9500)
+            .unwrap();
+        assert_eq!(minimum_ust, MicroUst(7_600_000_000));
+
+        // 331 StSol rewards and 50% slippage
+        let minimum_ust = price_array
+            .calculate_minimum_price(StLamports(331_000_000_000), 5000)
+            .unwrap();
+        assert_eq!(minimum_ust, MicroUst(16_550_000_000));
+    }
```
ruuda, line 939: This test is good, but it tests the very special case where all samples are the same. We should have an independent test to confirm that finding the median works.

Also, we should have a test to confirm that `insert_and_rotate` kicks out the right value.

ruuda (commented):
Thanks for adding a wrapper type, much clearer like this indeed!

anker/src/state.rs
enriquefynn: I was with the impression that we get the median from the array, that would be the median of the chronologically sorted elements.

enriquefynn: But why not just sort it 😬 ?

ruuda: > But why not just sort it

NIH-ism and premature optimization on my part :grimacing: Getting the median doesn’t require a full sort, and for small arrays of known size, sorting networks can do better than a generic sorting algorithm, and for median it is possible to do even better than a sorting network.

But yeah I probably shouldn’t have gone down that rabbit hole ... You are right, we should just call `.sort()`, if we run into the compute limit we can use a specialized median-of-5.

anker/src/state.rs
```diff
@@ -90,6 +90,77 @@ pub struct HistoricalStSolPrice {
...
+        let minimum_ust_per_st_sol = (median_price.st_sol_price_in_ust
+            * Rational {
+                numerator: sell_rewards_min_out_bps,
+                denominator: 10_000,
+            })?;
+        let minimum_price = (rewards
+            * Rational {
+                numerator: minimum_ust_per_st_sol.0,
+                denominator: 1_000_000_000,
+            })?;
```
ruuda, line 159: Note to self: at first I thought we might pre-multiply the two rationals and get better precision, but then there is a risk of overflow in the 64-bit multiplication, so let’s keep it like this.

ruuda (approved):
LGTM with one suggestion and if you can correct one comment.

PR #534, opened 2022-03-08
Add minimum amount for slippage

enriquefynn:
Adjust existing tests

### DIFF
--- a/anker/src/logic.rs
+++ b/anker/src/logic.rs
@@ -1,4 +1,8 @@
-use crate::{error::AnkerError, token::BLamports, ANKER_MINT_AUTHORITY, ANKER_RESERVE_AUTHORITY};
+use crate::{
+    error::AnkerError,
+    token::{BLamports, MicroUst},
+    ANKER_MINT_AUTHORITY, ANKER_RESERVE_AUTHORITY,
+};
 use lido::{
     state::Lido,
     token::{Lamports, StLamports},
@@ -230,6 +234,7 @@ pub fn swap_rewards(
     amount: StLamports,
     anker: &Anker,
     accounts: &SellRewardsAccountsInfo,
+    minimum_ust_price: MicroUst,
 ) -> ProgramResult {
     if amount == StLamports(0) {
         msg!("Anker rewards must be greater than zero to be claimable.");
@@ -252,7 +257,7 @@ pub fn swap_rewards(
         None,
         Swap {
             amount_in: amount.0,
-            minimum_amount_out: 0,
+            minimum_amount_out: minimum_ust_price.0,
         },
     )?;
 
--- a/anker/src/logic.rs
+++ b/anker/src/logic.rs
@@ -234,7 +234,7 @@ pub fn swap_rewards(
     amount: StLamports,
     anker: &Anker,
     accounts: &SellRewardsAccountsInfo,
-    minimum_ust_price: MicroUst,
+    minimum_ust_out: MicroUst,
 ) -> ProgramResult {
     if amount == StLamports(0) {
         msg!("Anker rewards must be greater than zero to be claimable.");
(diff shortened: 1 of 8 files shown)


anker/src/logic.rs
```diff
@@ -230,6 +234,7 @@ pub fn swap_rewards(
     amount: StLamports,
     anker: &Anker,
     accounts: &SellRewardsAccountsInfo,
+    minimum_ust_price: MicroUst,
```
ruuda, line 237: This name looks wrong, it’s called `minimum_ust_price`, so I assume the unit would be something different than UST, e.g. the UST price in stSOL. If the unit is UST, then it’s really the price of stSOL (in UST).

Second, it’s passed as `minimum_amount_out`, so it’s not the price, it’s the price. Either we need to set `minimum_amount_out` to `amount * price`, or instead of price we take `minimum_ust_out` as an argument.

**Edit**: So from the calling code, it looks like this is the minimum stSOL price in UST. So we should call it `minimum_st_sol_price` or maybe even `minimum_st_sol_price_in_ust`, and we should set `minimum_amount_out` below to `amount * minimum_st_sol_price` (it needs some juggling with rationals to get the scale right).

anker/src/processor.rs
```diff
@@ -327,6 +329,19 @@ fn process_sell_rewards(program_id: &Pubkey, accounts_raw: &[AccountInfo]) -> Pr
         accounts.anker.key,
         accounts.st_sol_reserve_account,
     )?;
+
+    let clock = Clock::from_account_info(accounts.sysvar_clock)?;
+    let most_recent_sample = anker.historical_st_sol_prices.last();
```
ruuda, line 334: The max age should apply to the _least recent sample_, not to the most recent one! Otherwise, you can have _n_ - 1 very old samples, add one new sample, and then already call `SellRewards`. We need to compare against `historical_st_sol_prices[0]`.

Anyway, did I not implement this already? I thought I did. Maybe I didn’t push it :thinking:

anker/src/processor.rs
```diff
@@ -343,9 +358,14 @@ fn process_sell_rewards(program_id: &Pubkey, accounts_raw: &[AccountInfo]) -> Pr
     // If this underflows, something went wrong, and we abort the transaction.
     let rewards = (reserve_st_sol_before - b_sol_supply_value_in_st_sol)?;
 
+    // Get minimum amount we are willing to pay for the rewards in UST.
+    let minimum_ust_price = anker
```
ruuda, line 362: This is the minimum _stSOL_ price (in UST), not UST price!

anker/tests/tests/sell_rewards.rs
```diff
@@ -13,6 +13,7 @@ async fn test_successful_sell_rewards() {
     context
         .initialize_token_pool_and_deposit(Lamports(DEPOSIT_AMOUNT))
         .await;
+    context.fetch_pool_price().await;
```
ruuda, line 16: Hah, so this is a good test for the bad behavior now, we should keep a copy of this test but expect it to fail. You shouldn’t be able to successfully `sell_rewards()` after calling `fetch_pool_price()` just once, it needs to be called 5 times, the point was that an attacker would need to sandwich at least 3 of those 5 calls, which is expensive, and also less likely to succeed than sandwiching just one.

ruuda (changes requested):

anker/src/processor.rs
enriquefynn: Humm... That's what I was wondering. It's better like this, I was going to add another check for ensuring the past values were done in < minimum, but if we check the first one we ensure the newer entries are correct.

anker/src/logic.rs
enriquefynn: I changed it to `minimum_ust_out` and renamed the function to `minimum_ust_swap_amount`.
I think it's better to do the rewards calculation inside the function

ruuda (approved):
LGTM with one suggestion.

PR #535, opened 2022-03-08
Test fetch pool price

enriquefynn:
Add some tests for the `FetchPoolPrice` instruction.

### DIFF
--- a/anker/tests/tests/fetch_pool_price.rs
+++ b/anker/tests/tests/fetch_pool_price.rs
@@ -27,9 +27,9 @@ async fn test_successful_fetch_pool_price() {
     let anker = context.get_anker().await;
 
     // Initially, there are 10 stSOL and 10_000 UST in the pool.
-    // for maintaining the constant product k = 10 *10_000 = 100_000.
+    // for maintaining the constant product k = 10 * 10_000 = 100_000.
     // When selling 1 StSOL we should maintain the equality:
-    // (10 + 1) * (10_000 - x) = 100_000, x = 909.0909090909091
+    // (10 + 1) * (10_000 - x) = k, x = 909.0909090909091
     let current_ust_price = MicroUst(909_090_909);
 
     let mut expected_historical_st_sol_prices = (0..5)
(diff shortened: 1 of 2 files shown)


anker/tests/tests/fetch_pool_price.rs
```diff
@@ -0,0 +1,70 @@
...
+    context
+        .solido_context
+        .advance_to_normal_epoch(2 + POOL_PRICE_NUM_SAMPLES as u64);
+    context.fetch_pool_price().await;
+    let anker = context.get_anker().await;
+    assert_eq!(
+        anker.historical_st_sol_prices.0[..],
+        expected_historical_st_sol_prices
+    );
+}
```
ruuda, line 54: Can we also add a test that swaps against the pool to change the price, and check that we sample that new price as expected?

anker/tests/tests/fetch_pool_price.rs
```diff
@@ -0,0 +1,70 @@
...
+        .initialize_token_pool_and_deposit(Lamports(DEPOSIT_AMOUNT))
+        .await;
+    context.fetch_pool_price().await;
+    context
+        .solido_context
+        .context
+        .warp_to_slot(956256 + POOL_PRICE_MIN_SAMPLE_DISTANCE - 1)
+        .expect("Failed to warp to slot");
+    let result = context.try_fetch_pool_price().await;
+    assert_solido_error!(result, AnkerError::FetchPoolPriceTooEarly);
```
ruuda, line 69: To confirm that the test works, we should also warp to slot `956256 + POOL_PRICE_MIN_SAMPLE_DISTANCE` and confirm that that one slot later, we _can_ fetch the price.

ruuda (commented):
Thanks for adding tests! I have two suggestions for extending them.

enriquefynn:
@ruuda: can you check with me why test `test_fetch_pool_price_when_price_changed` doesn't pass?
There's `9090.909091 UST` and `11 stSOL`, so the price for swapping `1 stSOL` should be more or less`8333.33333 UST`:
```
9090.909091 * 11 = 100000.000001
8333.33333 * 12 = 99999.999996
```
But in the test it's `757.575757`, but `757.575757 * 12 = 9090.909083999999`
I'm probably missing something on how this AMM works.

anker/tests/tests/fetch_pool_price.rs
```diff
@@ -0,0 +1,132 @@
...
+    let mut context = Context::new().await;
+    context
+        .initialize_token_pool_and_deposit(Lamports(DEPOSIT_AMOUNT))
+        .await;
+    for epoch in 2..2 + POOL_PRICE_NUM_SAMPLES as u64 {
+        context.solido_context.advance_to_normal_epoch(epoch);
+        context.fetch_pool_price().await;
+    }
+    let anker = context.get_anker().await;
+    let current_ust_price = MicroUst(909090909);
```
ruuda, line 28: Can you add a comment to explain where this number comes from?

ruuda (commented):
The structure of the tests looks good, but I didn’t verify the numbers ... can you add a comment to explain how the initial stSOL price emerges? That would make it a bit easier for me to check.

ruuda (approved):
LGTM!

PR #537, opened 2022-03-09
Fix CI tests 🤞 

enriquefynn:
- Wait 60s instead of 5s for validator to be up.
- Fix solana command to check our own instance for when it's up.
- Kill validator with -9.
- Use different status codes when existing with error.


ruuda (approved):
:crossed_fingers:

PR #540, opened 2022-03-17
Implement fetch pool price in maintainer

enriquefynn:
Call `FetchPoolPrice` inside maintainer if the following conditions are met:
- current slot > 1st slot of next epoch - (`POOL_PRICE_MAX_SAMPLE_AGE + POOL_PRICE_MIN_SAMPLE_DISTANCE`) OR `rewards >= min_rewards_to_sell`
- slots elapsed from the last check > `POOL_PRICE_MIN_SAMPLE_DISTANCE`

### DIFF
--- a/cli/maintainer/src/maintenance.rs
+++ b/cli/maintainer/src/maintenance.rs
@@ -775,69 +775,11 @@ impl SolidoState {
         (reserve_st_sol - st_sol_amount).ok()
     }
 
-    /// Try to update the historical pool price exchange rate to protect us
+    /// Try to sell the extra stSOL rewards for UST tokens or
+    /// to update the historical pool price exchange rate to protect us
     /// against sandwiching attacks.
-    /// We try to update the historical price when we're near the epoch
-    /// boundaries.
-    pub fn try_fetch_pool_price(&self) -> Option<MaintenanceInstruction> {
-        let anker_state = self.anker_state.as_ref()?;
-        let current_slot = self.clock.slot;
-        let next_epoch_begin_slot = self
-            .epoch_schedule
-            .get_first_slot_in_epoch(self.clock.epoch + 1);
-
-        // We start fetching pool prices if we are before the epoch's
-        // end minus `POOL_PRICE_MAX_SAMPLE_AGE` and minus one sample distance,
-        // so we have some margin.
-        let slot_start_fetch_pool_price = next_epoch_begin_slot
-            .saturating_sub(POOL_PRICE_MAX_SAMPLE_AGE + POOL_PRICE_MIN_SAMPLE_DISTANCE);
-        let should_start_fetch = current_slot > slot_start_fetch_pool_price;
-
-        let last_price = anker_state.anker.historical_st_sol_prices.last();
-        let slots_elapsed = current_slot.saturating_sub(last_price.slot);
-        let passed_minimum_update_price = slots_elapsed >= POOL_PRICE_MIN_SAMPLE_DISTANCE;
-
-        // If we have some rewards to claim, let's try to keep the historical
-        // price updated until we claim the rewards.
-        let rewards = self.get_anker_rewards()?;
-        let min_rewards_to_sell = self
-            .solido
-            .exchange_rate
-            .exchange_sol(Self::MINIMUM_WITHDRAW_AMOUNT)
-            .expect("The price of a signature should be small enough that it doesn't overflow.");
-        let have_rewards_to_claim = rewards >= min_rewards_to_sell;
-
-        if (should_start_fetch || have_rewards_to_claim) && passed_minimum_update_price {
-            let expected_st_sol_price_in_ust = get_one_st_sol_for_ust_price_from_pool(
-                &anker_state.constant_product_calculator,
-                &anker_state.pool_st_sol_account,
-                &anker_state.pool_ust_account,
-                anker_state.pool_st_sol_balance,
-                anker_state.pool_ust_balance,
-            )
-            .ok()?;
-            Some(MaintenanceInstruction::new(
-                anker_state.get_fetch_pool_price_instruction(self.solido_address),
-                MaintenanceOutput::FetchPoolPrice {
-                    expected_st_sol_price_in_ust,
-                },
-            ))
-        } else {
-            None
-        }
-    }
-
-    /// Try to sell the extra stSOL rewards for UST tokens.
     pub fn try_sell_anker_rewards(&self) -> Option<MaintenanceInstruction> {
         let anker_state = self.anker_state.as_ref()?;
-        let oldest_sample = anker_state.anker.historical_st_sol_prices.first();
-        let slots_elapsed = self.clock.slot.saturating_sub(oldest_sample.slot);
-
-        // Check if we can sell Anker rewards for consistency. This should not
-        // happen if we call `try_fetch_pool_price` before this function.
-        if slots_elapsed > POOL_PRICE_MAX_SAMPLE_AGE {
-            return None;
-        }
 
         let rewards = self.get_anker_rewards()?;
         let min_rewards_to_sell = self
@@ -846,7 +788,7 @@ impl SolidoState {
             .exchange_sol(Self::MINIMUM_WITHDRAW_AMOUNT)
             .expect("The price of a signature should be small enough that it doesn't overflow.");
 
-        let ust_per_st_sol = get_one_st_sol_for_ust_price_from_pool(
+        let expected_st_sol_price_in_ust = get_one_st_sol_for_ust_price_from_pool(
             &anker_state.constant_product_calculator,
             &anker_state.pool_st_sol_account,
             &anker_state.pool_ust_account,
@@ -855,7 +797,7 @@ impl SolidoState {
         )
         .ok()?;
 
-        let expected_proceeds = MicroUst(rewards.0 * ust_per_st_sol.0);
+        let expected_proceeds = MicroUst(rewards.0 * expected_st_sol_price_in_ust.0);
         // We want at least 0.01 UST out if we are going to do the swap at all.
         let min_proceeds = MicroUst(10_000);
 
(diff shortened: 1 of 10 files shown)


cli/maintainer/src/maintenance.rs
```diff
@@ -754,33 +768,108 @@ impl SolidoState {
...
 
-        let rewards = (reserve_st_sol - st_sol_amount).ok()?;
+        (reserve_st_sol - st_sol_amount).ok()
+    }
 
+    /// Try to update the historical pool price exchange rate to protect us
+    /// against sandwiching attacks.
+    /// We try to update the historical price when we're near the epoch
+    /// boundaries.
+    pub fn try_fetch_pool_price(&self) -> Option<MaintenanceInstruction> {
```
ruuda, line 778: I think we can simplify this and keep only the `try_sell_anker_rewards` function. Or at least, keep only that function in the `.or_else` chain at line 1688. `try_sell_anker_rewards` would then do something like this (pseudocode):

```rust
if rewards == 0 {
    return None;
}

// Confirm that the price samples are up to date before we sell.
if oldest_price_sample.age() > MAX_AGE {
    if youngest_price_sample.age() < MIN_DISTANCE {
        // Too soon, can’t update or sell yet.
        // Possibly we would need something to not make it wait
        // too long before the next iteration.
        return None; 
    }
    
    return call_fetch_pool_price();
}

// If we get here, the price samples are up to date and we can sell,
// but it might still be too soon.
if youngest_price_sample.age() < MIN_DISTANCE {
    return None;
}

return call_sell_rewards();
```

It does create some delay versus fetching the price at the end of the previous epoch already, but I don’t think that’s a problem, the exact time we sell the rewards doesn’t matter as long as we can do it once per epoch. Also, if somebody else starts running a cranker bot and they do update the price at the end of the previous epoch ... good for them, then we can immediately sell the rewards.

ruuda (commented):
I would change the conditions for fetching the price a bit, to be independent of the absolute slot number, and make it depend only on the state and the relative age of the price samples.

cli/maintainer/src/maintenance.rs
```diff
@@ -754,33 +772,75 @@ impl SolidoState {
...
             .exchange_rate
             .exchange_sol(Self::MINIMUM_WITHDRAW_AMOUNT)
             .expect("The price of a signature should be small enough that it doesn't overflow.");
 
-        // Estimate the amount of UST that we will get when swapping against the constant product
-        // pool. This is only an approximation because it doesn’t uphold the constant product
-        // invariant, but this is a reasonable approximation when the rewards to sell are small
-        // compared to the pool balance. Also, this ignores swap fees. It would have been nice
-        // to use the actual `spl_token_swap::state::SwapV1` instance, but it is not `Send`, so
-        // we can’t store it in our state.
```
ruuda, line 770: I think this comment was useful for understanding why we do `rewards * expected_st_sol_price_of_one_ust` instead of computing the more precise value from the `SwapV1` instance, can we preserve it?

cli/maintainer/src/maintenance.rs
```diff
@@ -754,33 +772,75 @@ impl SolidoState {
...
+                None
+            } else {
+                Some(MaintenanceInstruction::new(
+                    anker_state.get_fetch_pool_price_instruction(self.solido_address),
+                    MaintenanceOutput::FetchPoolPrice {
+                        expected_st_sol_price_in_ust,
+                    },
+                ))
+            }
+        } else if expected_proceeds < min_proceeds
```
ruuda, line 841: This looks a bit buried in this long if-else chain, I would move it to just below the `min_proceeds` definition and make it an early return:

```rust
// We want at least 0.01 UST out if we are going to do the swap at all.
let min_proceeds = MicroUst(10_000);
if expected_proceeds < min_proceeds {
    return None;
}
```

If we are not going to swap at all, there is also no need to update the price.

cli/maintainer/src/maintenance.rs
```diff
@@ -754,33 +772,75 @@ impl SolidoState {
...
+        let minimum_ust_amount_for_rewards = anker_state
+            .anker
+            .historical_st_sol_prices
+            .minimum_ust_swap_amount(rewards, anker_state.anker.sell_rewards_min_out_bps)
+            .ok()?;
+
         // We should not call the instruction if the rewards are 0, or if the rewards are so small
         // that the transaction cost is a significant portion of the rewards.
-        if rewards < min_rewards_to_sell || expected_proceeds < min_proceeds {
+        if rewards < min_rewards_to_sell {
```
ruuda, line 815: We can move this check all the way to the top right after the definition of `min_rewards to sell`: if we don’t have enough, we don’t need to bother with any of the rest.

(I don’t have a strong opinion on this, but now that I’m reading this function front to back, I think it would be a bit easier to follow if it’s a list of check->action, instead of computing various values first and then having all the checks at the end.)

cli/maintainer/src/maintenance.rs
```diff
@@ -754,33 +772,75 @@ impl SolidoState {
...
+            } else {
+                Some(MaintenanceInstruction::new(
+                    anker_state.get_fetch_pool_price_instruction(self.solido_address),
+                    MaintenanceOutput::FetchPoolPrice {
+                        expected_st_sol_price_in_ust,
+                    },
+                ))
+            }
+        } else if expected_proceeds < min_proceeds
+            || expected_proceeds < minimum_ust_amount_for_rewards
```
ruuda, line 842: We can add an early return below the definition of `minimum_ust_amount_for_rewards` as well. I don’t have a strong opinion on it, it’s mostly a matter of preference, but I feel it would make the function a bit easier to read.

cli/maintainer/src/maintenance.rs
```diff
@@ -754,33 +772,75 @@ impl SolidoState {
...
+            if slots_elapsed_since_youngest_sample < POOL_PRICE_MIN_SAMPLE_DISTANCE {
+                None
+            } else {
+                Some(MaintenanceInstruction::new(
+                    anker_state.get_fetch_pool_price_instruction(self.solido_address),
+                    MaintenanceOutput::FetchPoolPrice {
+                        expected_st_sol_price_in_ust,
+                    },
+                ))
+            }
```
ruuda, line 840: :ok_hand: Great, this part with the sample ages is very clear!

ruuda (commented):
The structure of this looks good to me! I have some very small nits about how we could make it a bit easier to read, but I don’t have a strong opinion on it.

cli/maintainer/src/maintenance.rs
enriquefynn: Humm I see, My idea was to use the curve calculator so we don't do the `rewards * expected_st_sol_price_of_one_ust`, I think I forgot to include it. I'll do it.

cli/maintainer/src/maintenance.rs
enriquefynn: Yes! That's much better, thanks!

cli/maintainer/src/maintenance.rs
enriquefynn: Indeed, much better :)

cli/maintainer/src/maintenance.rs
```diff
@@ -754,34 +773,87 @@ impl SolidoState {
...
-            .map(|x| MicroUst(x.0))
-            .unwrap_or(MicroUst(0));
         // We want at least 0.01 UST out if we are going to do the swap at all.
         let min_proceeds = MicroUst(10_000);
 
-        // We should not call the instruction if the rewards are 0, or if the rewards are so small
-        // that the transaction cost is a significant portion of the rewards.
-        if rewards < min_rewards_to_sell || expected_proceeds < min_proceeds {
-            None
+        // We want at least 0.01 UST out if we are going to do the swap at all.
```
ruuda, line 811:
```suggestion
```
This comment is now duplicated with the one three lines above.

cli/maintainer/src/maintenance.rs
```diff
@@ -754,34 +773,87 @@ impl SolidoState {
...
+        }
+
+        // Check if we can sell the rewards with the preset slippage tolerance.
+        // Note that this might change when the instruction gets included in the block.
+        let minimum_ust_amount_for_rewards = anker_state
+            .anker
+            .historical_st_sol_prices
+            .minimum_ust_swap_amount(rewards, anker_state.anker.sell_rewards_min_out_bps)
+            .ok()?;
+        if expected_proceeds < minimum_ust_amount_for_rewards {
```
ruuda, line 823: I think this check is a bit misleading, because `expected_proceeds` is computed through `swap_without_fees`, but the `minimum_ust_swap_amount` is including the fees. So the `expected_proceeds` is an overestimation.

How difficult would it be to take the swap fees into account when computing `expected_proceeds`? If I recall correctly it’s just a fixed percentage, so if it’s one additional multiplication by a `Rational`, that should be doable.

ruuda (approved):
:ok_hand: `try_sell_anker_rewards` looks very straightforward now, it’s easy to follow what it does when you read it front to back :+1: 

I have one nit, and one comment that I’m not sure if it is something we should address here, but the computation of the expected proceeds does not take the swap fees into account, so it will be somewhat off. We could say that we’re fine with that, it is no big deal if the transaction fails, and even if the check passes there is a risk of the transaction failing. Depending on how busy the stSOL–UST pool is, the check may even be harmful, because there is a delay between when we fetch the state and when the transaction executes.

If it’s easy to take the fees into account, I would do that for now, and then once we run it in production, if we notice that it often doesn’t sell because of this check, then we can see if we need to change it.

cli/maintainer/src/maintenance.rs
enriquefynn: But at `get_one_st_sol_for_ust_price_from_pool` https://github.com/ChorusOne/solido/blob/7009abc6ba37063aaeb0a2ef3b374f4f20cb17b3/anker/src/logic.rs#L316 it's also swapping without fees, am I missing something?

ruuda: Yes, it is, but the `sell_rewards_min_out_bps` is the tolerance for slippage + fees. So if we set the `sell_rewards_min_out_bps` to 9800 bps (98%), then if the fees are 0.1% we would tolerate 1.9% slippage, but if the fees are 1%, we would tolerate only 1% slippage.

enriquefynn: Apparently the swap takes fees from the `source`:
https://github.com/solana-labs/solana-program-library/blob/95cd64ce022844d5ad2b2d4809c4b29c063162b3/token-swap/program/src/curve/base.rs#L80-L84
The swapped amount is going to remain the same, just some of the rewards are going to be `fees`.

ruuda: They compute the fees from the source amount, then subtract that from the source amount, and then put that in `swap_without_fees`. So the end result is still less output, then when you would put the full input amount in `swap_without_fees`.

enriquefynn: Indeed, I added the `Fees` struct to the `AnkerState` and subtracted the trade and owner fee from the rewards. The alternative is to either save the swap pool or store a closure in the struct.
The swap pool cannot be saved trivially in the struct because it has a trait object that doesn't implement `Send + Sync` that's needed in another part of the code.
I started adding the closure, but the changes propagate all throughout the code if I added it as a generic argument. So I think this is the simplest way of doing it, it's just a couple of lines.

ruuda (approved):
LGTM :rocket: 

The fmt lints are failing due to an unrelated problem which I fixed as a drive-by fix in #548.

PR #541, opened 2022-03-17
Test fetch pool price

enriquefynn:
Test that the maintainer updates the pool price 5 times before selling the rewards

### DIFF
--- a/tests/test_anker.py
+++ b/tests/test_anker.py
@@ -24,6 +24,7 @@ from util import (
     spl_token,
     spl_token_balance,
     create_spl_token_account,
+    wait_for_slots,
 )
 
 print('Creating test accounts ...')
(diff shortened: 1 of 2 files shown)


tests/start_test_validator.py
```diff
@@ -21,6 +21,7 @@
         'solana-test-validator',
     ],
     stdout=subprocess.DEVNULL,
+    shell=True,
```
ruuda, line 24: Why is this needed? It makes Python execute the command through a shell, so you can do things like put a `~` in the command and have it expand to the value of `$HOME`. But in this case our command doesn’t use any shell features, so I don’t think this does anything.

tests/test_anker.py
```diff
@@ -305,6 +306,22 @@ def deposit_solido_sol(amount_sol: float) -> str:
 assert result['st_sol_reserve_balance_st_lamports'] == 1_000_000_000
 print('> Anker stSOL reserve now contains 1 SOL.')
 
+print('\nPerforming 5 maintenance to populate the historical prices ...')
```
ruuda, line 309:
```suggestion
print('\nPerforming maintenance 5 times to populate the historical prices ...')
```

tests/util.py
```diff
@@ -322,3 +322,18 @@ def approve_and_execute(transaction_address: str) -> None:
...
+def wait_for_slots(slots: int) -> None:
+    import time
+
+    """
+    Blocks waiting until `slots` slots have passed.
+    """
+    slots_beginning = int(solana('get-slot'))
+    while True:
+        # Wait 1 second to poll next slot height (around 2 slots)
+        time.sleep(1)
```
ruuda, line 336: Hey, I finally wrote [a blog post](https://ruudvanasseldonk.com/2022/03/20/please-put-units-in-names) about this :grimacing: In this case it says "1 second" in the comment right next to it so I suppose that works as well, I don’t have a strong opinion in this case.

ruuda (commented):
Looks almost good to me!

tests/start_test_validator.py
enriquefynn: I'm not sure why, but this change makes our CI work, see https://github.com/ChorusOne/solido/pull/541/commits/57bd0fbf348fa299c494c635e4e9967ace212aa4

ruuda (approved):

PR #544, opened 2022-03-18
Show Anker instruction that require multisig

enriquefynn:
Show the following Anker instructions:
- ChangeTerraRewardsDestination
- ChangeTokenSwapPool
- ChangeSellRewardsMinBps

Test management functions

### DIFF
--- a/cli/maintainer/src/commands_anker.rs
+++ b/cli/maintainer/src/commands_anker.rs
@@ -246,6 +246,9 @@ struct ShowAnkerOutput {
     #[serde(serialize_with = "serialize_b58")]
     b_sol_mint_authority: Pubkey,
 
+    #[serde(serialize_with = "serialize_b58")]
+    token_swap_pool: Pubkey,
+
     #[serde(serialize_with = "serialize_bech32")]
     terra_rewards_destination: TerraAddress,
 
@@ -354,6 +357,7 @@ fn command_show_anker(
         solido_address: anker.solido,
         solido_program_id: anker.solido_program_id,
 
+        token_swap_pool: anker.token_swap_pool,
         terra_rewards_destination: anker.terra_rewards_destination,
         sell_rewards_min_out_bps: anker.sell_rewards_min_out_bps,
 
--- a/cli/maintainer/src/commands_anker.rs
+++ b/cli/maintainer/src/commands_anker.rs
@@ -287,6 +287,7 @@ impl fmt::Display for ShowAnkerOutput {
             "Rewards destination:   {}",
             self.terra_rewards_destination
         )?;
+        writeln!(f, "Token Swap Pool:       {}", self.token_swap_pool)?;
         if self.sell_rewards_min_out_bps <= 9999 {
             writeln!(f,
                      "Sell rewards min out:  {}.{:>02}% of the expected amount ({}.{:>02}% slippage + fees)",
--- a/cli/maintainer/src/commands_multisig.rs
+++ b/cli/maintainer/src/commands_multisig.rs
@@ -461,6 +461,7 @@ enum SolidoInstruction {
     },
 }
 
+#[allow(clippy::enum_variant_names)]
 #[derive(Serialize)]
 enum AnkerInstruction {
     ChangeTerraRewardsDestination {
@@ -1023,7 +1024,7 @@ fn show_transaction(
             }
         }
     } else if anker_program_id == Some(instr.program_id) {
-        match try_parse_anker_instruction(config, &&instr) {
+        match try_parse_anker_instruction(config, &instr) {
             Ok(instr) => instr,
             Err(SnapshotError::MissingAccount) => return Err(SnapshotError::MissingAccount),
             Err(SnapshotError::MissingValidatorIdentity(addr)) => {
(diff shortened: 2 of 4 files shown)


tests/test_anker.py
```diff
@@ -388,3 +388,82 @@ def deposit_solido_sol(amount_sol: float) -> str:
...
+    '--multisig-program-id',
+    multisig_program_id,
+    '--token-swap-pool',
+    new_token_pool_address,
+    keypair_path=test_addrs[0].keypair_path,
+)
+transaction_address = transaction_result['transaction_address']
+approve_and_execute(transaction_address)
+
+print('> Changing Terra rewards destination')
```
ruuda, line 454:
```suggestion
print('> Changing min out basis points')
```

tests/test_anker.py
```diff
@@ -388,3 +388,82 @@ def deposit_solido_sol(amount_sol: float) -> str:
...
+    '--multisig-address',
+    multisig_instance,
+    '--multisig-program-id',
+    multisig_program_id,
+    '--token-swap-pool',
+    new_token_pool_address,
+    keypair_path=test_addrs[0].keypair_path,
+)
+transaction_address = transaction_result['transaction_address']
+approve_and_execute(transaction_address)
```
ruuda, line 452: Shouldn’t we also do an `anker show` after this and confirm that the token pool actually changed? Also with the other changes below.

cli/maintainer/src/commands_multisig.rs
```diff
@@ -448,6 +461,46 @@ enum SolidoInstruction {
...
+        new_token_swap_pool: Pubkey,
+    },
+    ChangeSellRewardsMinBps {
+        #[serde(serialize_with = "serialize_b58")]
+        anker_instance: Pubkey,
+
+        #[serde(serialize_with = "serialize_b58")]
+        manager: Pubkey,
+
+        old_sell_rewards_min_bps: u64,
```
ruuda, line 498:
```suggestion
        old_sell_rewards_min_out_bps: u64,
```
Nit: for consistency with the field in the `Anker` struct. Also below.

ruuda (commented):
LGTM with two nits!

cli/maintainer/src/commands_anker.rs
```diff
@@ -246,6 +246,9 @@ struct ShowAnkerOutput {
     #[serde(serialize_with = "serialize_b58")]
     b_sol_mint_authority: Pubkey,
 
+    #[serde(serialize_with = "serialize_b58")]
+    token_swap_pool: Pubkey,
```
ruuda, line 250: If we put it in the struct, we should also update the `fmt::Display` impl below to print it.

tests/test_anker.py
```diff
@@ -388,3 +389,108 @@ def deposit_solido_sol(amount_sol: float) -> str:
...
+    'anker',
+    'change-terra-rewards-destination',
+    '--anker-address',
+    anker_address,
+    '--multisig-address',
+    multisig_instance,
+    '--multisig-program-id',
+    multisig_program_id,
+    '--terra-rewards-destination',
+    'terra14dycr8jm7e5kw88g4studekkzzw5xc5ffnp4hk',
```
ruuda, line 405: Let’s put this in a variable `new_terra_rewards_destination` or something like that, so we don’t have to repeat it below. Also we can assert that it is different from the previous address, so we are sure that we are actually changing something.

tests/test_anker.py
```diff
@@ -388,3 +389,108 @@ def deposit_solido_sol(amount_sol: float) -> str:
...
+    'anker',
+    'change-sell-rewards-min-out-bps',
+    '--anker-address',
+    anker_address,
+    '--multisig-address',
+    multisig_instance,
+    '--multisig-program-id',
+    multisig_program_id,
+    '--sell-rewards-min-out-bps',
+    '10',
```
ruuda, line 466: It wasn’t obvious to me at first that this value is different from the previous value, I had to scroll back to the init call to see that it is initialized with a min_out_bps of 0. We could do this instead:

```python
new_min_out_bps = anker['sell_rewards_min_out_bps'] + 10
solido(..., '--sell-rewards-min-out-bps', str(new_min_out_bps), ...)
```

and then it’s obvious that we are actually changing the value, and if we ever edit the init call, we don’t accidentally break this test if we set the min out basis points to 10 at init.

ruuda (commented):
Looks almost good, I have a few suggestions.

tests/test_anker.py
```diff
@@ -484,13 +486,13 @@ def deposit_solido_sol(amount_sol: float) -> str:
     'ust_reserve': authorities['ust_reserve_account'],
     'b_sol_mint_authority': authorities['b_sol_mint_authority'],
     'reserve_authority': authorities['reserve_authority'],
-    'terra_rewards_destination': 'terra14dycr8jm7e5kw88g4studekkzzw5xc5ffnp4hk',
+    'terra_rewards_destination': new_terra_rewards_destination,
     'token_swap_pool': new_token_pool_address,
     'sell_rewards_min_out_bps': 10,
```
ruuda, line 491: This should now be `new_min_out_bps`, right?

ruuda (approved):
LGTM with one more comment.

cli/maintainer/src/commands_anker.rs
```diff
@@ -284,6 +287,7 @@ impl fmt::Display for ShowAnkerOutput {
             "Rewards destination:   {}",
             self.terra_rewards_destination
         )?;
+        writeln!(f, "Token Swap Pool:       {}", self.token_swap_pool)?;
```
ruuda, line 290: This was one space too little, I fixed it in 65efad75.

PR #547, opened 2022-03-25
Add `Expires` header to query response.

### LINKED ISSUE #526: Add Expires header to APY responses
Right now we don’t include any caching headers in the responses from the APY daemon. However, APY only changes once per epoch, browsers and CDNs should be able to cache these responses.

Now it gets a bit tricky because the query params are involved. If it’s for an absolute date range (specific `begin` and `end`) in the past, no new data will come in, so that data is valid indefinitely. If it’s for a relative date range (end=now implicitly), then the response should be valid until the start of the next epoch. We can make a rough estimate of when that is based on previous epoch durations (which is already part of the response data anyway).

We can then put the expiration time in an [`Expires` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires).

enriquefynn:
If queries are from the past, set the `Expires` header to `now` + 1 year, otherwise, if we are in the same epoch as the query's end slot, we set the `Expires` to the approximate time until the next epoch.
Closes #526

### DIFF
--- a/cli/listener/src/main.rs
+++ b/cli/listener/src/main.rs
@@ -15,14 +15,14 @@ use rusqlite::{params, Connection, Row};
 use serde::{Deserialize, Serialize};
 use solana_client::rpc_client::RpcClient;
 use solana_sdk::{
-    clock::{Epoch, Slot},
+    clock::{Epoch, Slot, SLOT_MS},
     commitment_config::CommitmentConfig,
     epoch_schedule::EpochSchedule,
     pubkey::Pubkey,
     signature::Keypair,
 };
 use solido_cli_common::{
-    error::{AsPrettyError, Error},
+    error::{Abort, AsPrettyError, Error},
     prometheus::{write_metric, Metric, MetricFamily},
     snapshot::{Config, OutputMode, SnapshotClient, SnapshotClientConfig},
 };
@@ -570,11 +570,41 @@ fn get_success_response(interval_prices: IntervalPrices) -> ResponseBox {
     .boxed()
 }
 
+/// Get a header with `Expires` key.
+/// If the end date for the query is far in the past we set it to expire in 1
+/// year. However, if the end date is in the same epoch, we set it to expire in
+/// the approximate time the next epoch begins.
+fn get_expires_header(
+    now: chrono::DateTime<chrono::Utc>,
+    current_slot: Slot,
+    interval_prices: &IntervalPrices,
+) -> Header {
+    let epoch_sched = EpochSchedule::without_warmup();
+    let (epoch_for_current_slot, _) = epoch_sched.get_epoch_and_slot_index(current_slot);
+
+    // Set the validity of the query until the next epoch
+    let expires_date = if epoch_for_current_slot == interval_prices.end_epoch {
+        let first_slot_next_epoch = epoch_sched.get_first_slot_in_epoch(epoch_for_current_slot + 1);
+        // Does not underflow because `first_slot_next_epoch <= current_slot`.
+        let slots_diff = first_slot_next_epoch - current_slot;
+        // Estimate for how many milliseconds the response is valid.
+        let ms_diff = slots_diff * SLOT_MS;
+        now + chrono::Duration::milliseconds(ms_diff as i64)
+    } else {
+        // 1 year duration if the end is in the past.
+        now + chrono::Duration::days(365)
+    };
+    Header::from_bytes(&b"Expires"[..], &expires_date.to_rfc2822()[..])
+        .expect("Static header value, does not fail at runtime.")
+}
+
 /// Gets a response that will be sent to a requesting client for an APY
 /// query.
 fn get_interval_price_request(
     db_connection: &Connection,
     range_date: &DateBeginEnd,
+    now: chrono::DateTime<chrono::Utc>,
+    current_slot: Slot,
 ) -> ResponseBox {
     let interval_prices = get_interval_price_for_period(
         db_connection
@@ -593,7 +623,11 @@ fn get_interval_price_request(
         Ok(interval_prices_opt) => {
             if let Some(interval_prices) = interval_prices_opt {
                 // Got interval prices.
-                get_success_response(interval_prices)
+                let expires_header = get_expires_header(now, current_slot, &interval_prices);
+                let mut response = get_success_response(interval_prices);
+
+                response.add_header(expires_header);
+                response
             } else {
                 // No interval price could be calculated, probably because of few data points.
                 get_error_response(ResponseError::NotFound(
@@ -638,6 +672,7 @@ fn serve_request(
     db_connection: &Connection,
     request: Request,
     metrics_mutex: &MetricsMutex,
+    current_slot: Slot,
 ) -> Result<(), std::io::Error> {
     let now = chrono::Utc::now();
 
@@ -668,7 +703,7 @@ fn serve_request(
             Response::from_data(out).with_header(content_type).boxed()
         }
         Ok(Endpoint::IntervalPriceRequest(range_date)) => {
-            get_interval_price_request(db_connection, &range_date)
+            get_interval_price_request(db_connection, &range_date, now, current_slot)
         }
         Err(err) => get_error_response(err),
     };
@@ -676,7 +711,11 @@ fn serve_request(
 }
 
 /// Spawn threads that run the http server.
-fn start_http_server(opts: &Opts, metrics_mutex: Arc<MetricsMutex>) -> Vec<JoinHandle<()>> {
+fn start_http_server(
+    opts: &Opts,
+    metrics_mutex: Arc<MetricsMutex>,
+    current_slot: Slot,
+) -> Vec<JoinHandle<()>> {
     let server = match Server::http(opts.listen.clone()) {
         Ok(server) => Arc::new(server),
         Err(err) => {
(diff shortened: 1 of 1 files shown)


enriquefynn:
I'm not so sure about this 😬. So CDNs also should store the query parameters for caching?!
If it's from the past, we set the `Expires` date to `now + 1 year`. Otherwise we compute how many slots are left for the next epoch and estimate with the `clock::SLOT_MS`

ruuda:
> So CDNs also should store the query parameters for caching?!

Yes, they do, the query parameter is part of the url, cdns should treat the entire url as opaque. In fact, that’s a common workaround if you messed up your caching policy and now the cdn is serving stale data: append a `?foobar=«uuid-or-something»` to your url.

cli/listener/src/main.rs
```diff
@@ -570,11 +570,41 @@ fn get_success_response(interval_prices: IntervalPrices) -> ResponseBox {
...
+    let epoch_sched = EpochSchedule::without_warmup();
+    let (epoch_for_current_slot, _) = epoch_sched.get_epoch_and_slot_index(current_slot);
+
+    // Set the validity of the query until the next epoch
+    let expires_date = if epoch_for_current_slot == interval_prices.end_epoch {
+        let first_slot_next_epoch = epoch_sched.get_first_slot_in_epoch(epoch_for_current_slot + 1);
+        // Does not underflow because `first_slot_next_epoch <= current_slot`.
+        let slots_diff = first_slot_next_epoch - current_slot;
+        // Estimate for how many milliseconds the response is valid.
+        let ms_diff = slots_diff * SLOT_MS;
```
ruuda, line 591: We both know that `SLOT_MS` is a lie :joy: It’s defined to be 400ms, but in practice it’s more like 550–600ms. In this case it’s not an issue, it just causes us to underestimate how long the data will be valid for. But if you want to be more accurate, we could also base it on the observed block time, as

```rust
if interval_prices.duration_epochs() > 0 {
    let epoch_duration = interval_prices.duration_wall_time() / interval_prices.duration_epochs();
    let slot_duration = epoch_duration / epoch_sched.get_slots_in_epoch(interval_prices.end_epoch);
    now + slot_duration * slots_diff
}
```

We can even do `now + slot_duration * (slots_diff * 8 / 10)` to underestimate the duration again, so we don’t overshoot and cache for a long time when the epoch suddenly ends sooner than expected.

Also, we should have a minimum time as well. Even if we are very close to the epoch boundary, we don’t need to let the value expire instantly, we can cache it for 60 seconds or so. The APY is not *that* important that you need to know the new value within seconds after the new epoch starts.

cli/listener/src/main.rs
```diff
@@ -668,15 +703,19 @@ fn serve_request(
...
     };
     request.respond(response)
 }
 
 /// Spawn threads that run the http server.
-fn start_http_server(opts: &Opts, metrics_mutex: Arc<MetricsMutex>) -> Vec<JoinHandle<()>> {
+fn start_http_server(
+    opts: &Opts,
+    metrics_mutex: Arc<MetricsMutex>,
+    current_slot: Slot,
```
ruuda, line 717: This makes the current slot a constant that gets set at server startup and never updates!

But so let’s think about this a bit more ... on the one hand I like this approach of basing the `Expires` on just the current slot and the `IntervalPrices`, then we can have one uniform way of handling things. But on the other hand, I think it would be better to base it on the request:

* If you request with an absolute begin and end range, if you actually got the range you asked for, then that value will never change any more, even if the end happens to fall in the current epoch.
* If you request with `?days=30`, then it’s a moving target, and we do need to make it expire at the epoch boundary.

Maybe what we can do, instead of including the current slot, is just to pass in whether the request used a relative date range or not, and base the decision for how long to cache on that?

cli/listener/src/main.rs
```diff
@@ -570,11 +570,41 @@ fn get_success_response(interval_prices: IntervalPrices) -> ResponseBox {
...
+    let expires_date = if epoch_for_current_slot == interval_prices.end_epoch {
+        let first_slot_next_epoch = epoch_sched.get_first_slot_in_epoch(epoch_for_current_slot + 1);
+        // Does not underflow because `first_slot_next_epoch <= current_slot`.
+        let slots_diff = first_slot_next_epoch - current_slot;
+        // Estimate for how many milliseconds the response is valid.
+        let ms_diff = slots_diff * SLOT_MS;
+        now + chrono::Duration::milliseconds(ms_diff as i64)
+    } else {
+        // 1 year duration if the end is in the past.
+        now + chrono::Duration::days(365)
```
ruuda, line 595: In theory this should be fine, but in practice, if we mess something up, one year is a long time to have to wait for the bad data to go away. (I think there is an option to flush the cache in Cloudflare, but it will be a hassle to get it executed.) In terms of load, it doesn’t matter *that* much when a CDN can save us 50 requests a year, the case where the CDN helps us is when we get thousands of requests a second. So we might also cache it for just 7 days. Then we get ~1 request per week (for a given url) which is still virtually nothing.

cli/listener/src/main.rs
enriquefynn: >This makes the current slot a constant that gets set at server startup and never updates!

True, we need to include the clock in the snapshot then to use it, I suppose.
I think the approach is also good and wouldn't change much the end result. We can do it like you described.

cli/listener/src/main.rs
```diff
@@ -570,18 +611,69 @@ fn get_success_response(interval_prices: IntervalPrices) -> ResponseBox {
...
+                        let slot_duration = epoch_duration
+                            / epoch_sched.get_slots_in_epoch(interval_prices.end_epoch) as i32;
+                        now + slot_duration * (slots_diff * 8 / 10) as i32
+                    } else {
+                        // We can't estimate the epoch duration, use Solana's
+                        // `SLOT_MS`. Empirically, this is lower than than what
+                        // we observe in reality, so we'd cache for less time
+                        // than ideal.
+                        let ms_diff = slots_diff * SLOT_MS;
+                        now + chrono::Duration::milliseconds(ms_diff as i64)
```
enriquefynn, line 650: If people query `since_launch` this is ok, but we could hardcode this expected time calculation. Then we don't need to use the `SLOT_MS` method.
What do you think it's better?

ruuda: Looks good like this!

cli/listener/src/main.rs
```diff
@@ -463,7 +496,8 @@ fn get_date_params<'a, I: IntoIterator<Item = (Cow<'a, str>, Cow<'a, str>)>>(
                     )
                 })?;
 
-                begin_opt.push(t);
+                begin_vec.push(t);
+                request_date_type = DateRequestType::Past;
```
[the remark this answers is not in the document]
ruuda: I think we need to check `end` against `now`, and if `end > now`, return `MovingTarget` as well.

cli/listener/src/main.rs
```diff
@@ -570,18 +611,69 @@ fn get_success_response(interval_prices: IntervalPrices) -> ResponseBox {
...
+                    let first_slot_next_epoch =
+                        epoch_sched.get_first_slot_in_epoch(clock.epoch + 1);
+                    let slots_diff = first_slot_next_epoch - clock.slot;
+                    if interval_prices.duration_epochs() > 0 {
+                        // We can estimate the epoch duration.
+                        let epoch_duration = interval_prices.duration_wall_time()
+                            / interval_prices.duration_epochs() as i32;
+                        let slot_duration = epoch_duration
+                            / epoch_sched.get_slots_in_epoch(interval_prices.end_epoch) as i32;
+                        now + slot_duration * (slots_diff * 8 / 10) as i32
```
ruuda, line 643:
```suggestion
                        // Cache until we estimate that 80% of the current epoch will be complete,
                        // so that we don’t cache for too long if the epoch suddenly ends faster than expected.
                        now + slot_duration * (slots_diff * 8 / 10) as i32
```

cli/listener/src/main.rs
```diff
@@ -570,18 +611,69 @@ fn get_success_response(interval_prices: IntervalPrices) -> ResponseBox {
...
+                        // than ideal.
+                        let ms_diff = slots_diff * SLOT_MS;
+                        now + chrono::Duration::milliseconds(ms_diff as i64)
+                    }
+                }
+            }
+        }
+    };
+
+    // Set minimum expiration time to 1 hour.
```
ruuda, line 657: 1h is still quite a long time to have to wait for bad data to go away if we mess something up, let’s do something like 5 minutes? The same argument as for reducing it from 1y to 7d applies: those 11 additional requests are not what’s going to kill the app in cases of high load. It’s when we get into multiple requests per second territory that load _might_ become a problem (but actually I expect we can easily handle hundreds of requests/sec), and even caching for 1s in that case will reduce the load from an unbounded number of requests/sec to 1 request/sec. (Per set of query params, but if somebody is deliberately varying those, that’s a DoS attack.)

cli/listener/src/main.rs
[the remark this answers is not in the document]
enriquefynn: Great! I removed the mutable variable and just checked `end >= now` at the end.

cli/listener/src/main.rs
enriquefynn: Ok, 5 minutes seems reasonable, thanks!

cli/listener/src/main.rs
```diff
@@ -1021,4 +1140,140 @@ mod test {
...
+            &interval_prices,
+            Some(current_clock),
+            &DateRequestType::Fixed,
+        );
+        assert_eq!(
+            expiration_header.field,
+            HeaderField::from_str("Expires").unwrap()
+        );
+        let next_year_expiration = begin_epoch_293_datetime + chrono::Duration::weeks(1);
+        assert_eq!(expiration_header.value, next_year_expiration.to_rfc2822());
```
ruuda, line 1185:
```suggestion
        let next_week_expiration = begin_epoch_293_datetime + chrono::Duration::weeks(1);
        assert_eq!(expiration_header.value, next_week_expiration.to_rfc2822());
```

ruuda (approved):
LGTM! I did find one more typo :grimacing:

PR #548, opened 2022-03-29
Include token swap accounts in `anker show` output

ruuda:
I need to donate some funds into one of these accounts on devnet to work around a too-eager assertion that we have since removed, but which is still included in version that I deployed to devnet earlier. So print the addresses so I know where the accounts are.


enriquefynn (approved):
LGTM!

ruuda:
I’ll wait with merging this until the audit of `anker-limit-slippage` is complete to not disturb that, this is not essential anyway.

PR #549, opened 2022-03-31
Enforce minimum distance between last price sample and `SellRewards`

ruuda:
Addresses https://github.com/ChorusOne/solido/pull/511/files#r839805004. Without this, you can put the final fetch in the same transaction as the sale, and sandwich the final sale by sandwiching two additional fetches (so you control 3 of the 5 price samples, so you control the median). With this added check, you need to sandwich one additional fetch before you can sandwich the final sale.

### DIFF
--- a/anker/src/error.rs
+++ b/anker/src/error.rs
@@ -63,6 +63,9 @@ pub enum AnkerError {
 
     /// Value of `sell_rewards_min_out_bps` is greater than 100% (1_000_000).
     InvalidSellRewardsMinOutBps = 4016,
+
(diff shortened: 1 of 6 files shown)


enriquefynn (approved):
LGTM!

enriquefynn:
Just one Nit: we need to add:
```rust
wait_for_slots(100)
```
in `tests/test_anker.py` just after we finish updating the pool price.

ruuda:
Good find, added, and I also confirmed that the test failed previously and passes now.

PR #551, opened 2022-04-04
Allow `SellRewards` just before `FetchPoolPrice`

ruuda:
As pointed out by Neodyme, a malicious cranker could spam `FetchPoolPrice` transactions, and although eventually the `SellRewards` should win if there is some probability for either transaction to win the race, this could delay selling for the rewards for some time. To mitigate that, give `SellRewards` a head start of one slot, so if there are two crankers that are spamming `FetchPoolPrice` and `SellRewards`, the `SellRewards` would win.


ruuda:
(fmt CI failure is due to an unrelated issue for which we have a pending fix)

enriquefynn (approved):
LGTM!

Issue #529, opened 2022-02-25
Add fuzzer for APY daemon


ruuda:
The request handler for the `/apy` endpoint is exposed to the internet. And https://github.com/ChorusOne/solido/pull/528/files#diff-608e8353e99d9e81d7e040834c6b6a167d60064f1eee6c5d8bdce129e23fa464R466 made me realize that even though it is quite a simple endpoint, I wouldn’t be surprised if we have a subtle overflow bug somewhere.

It should be easy to test the request handler with [`cargo-fuzz`](https://lib.rs/crates/cargo-fuzz) and the [`arbitary` crate](https://lib.rs/crates/arbitrary).

PR #552, opened 2022-04-06
Update Clap, add skeleton for fuzzer for APY daemon

ruuda:
I wanted to split out the APY daemon into a library, so I can fuzz functions in it more easily (#529). Then I ran into compilation issues with Clap for some reason. So I decided to update it.

This does change the `solido multisig token` command to no longer be a subcommand, it inlines the `transfer` subcommand directly now. I think that also resolves the panic we had there when you don’t provide any arguments, I think it was because it was an enum in an enum, while Clap expects enums to be in structs and structs in enums.

This does unfortunately create a duplicate dependency, on Clap 2.33 and Clap 3.1.6. The 2.33 dependency is deep in the tree of all Solana stuff, and Solana hasn’t updated, so we cannot get rid of it unfortunately.

GitHub does not recognize the rename, but `listener/main.rs` was renamed to `lib.rs`, the only change is to make `fn main` a `pub fn main` and to add the missing copyright header.

This also adds the files for a fuzz target, but nothing is being fuzzed yet.


enriquefynn (commented):
Looking good,
You forgot to update the clap dependency to `{ version = "3.1.6", features = ["derive"] }` in `cli/maintainer/Cargo.toml`, so clippy/build are complaining.

ruuda:
Good catch, fixed now.

enriquefynn:
This commit changed  the subcommand `'token transfer'` to `'token-transfer'`, but the test at https://github.com/ChorusOne/solido/blob/1181183e8b5b84fa77bbd362ab0a558fd44853be/tests/test_multisig.py#L512-L513
was not updated, can you change this to `'token-transfer'`?

ruuda:
Ah another good catch, will fix. Also, I now notice that as an unintended side-effect this makes the help output of `solido` colorized.

PR #553, opened 2022-04-06
Fuzz APY daemon http handler

### LINKED ISSUE #529: Add fuzzer for APY daemon

ruuda:
This adds a fuzzer to fuzz the http handler, as it will be exposed to the public internet, and we don’t want bad inputs to be able to crash the application.

The fuzzer generates instances of `Action`, which can be one of two cases:

* Insert a row into the database.
* Make an http request for a particular url.

This ensures that we can also test tricky cases that might only be reachable when the database is in a certain state.

Closes #529


cli/listener/fuzz/fuzz_targets/apy_endpoint.rs
```diff
@@ -1,6 +1,102 @@
...
+                price_lamports_numerator,
+                price_lamports_denominator,
+            } => {
+                let timestamp = match chrono::Utc.timestamp_millis_opt(*timestamp_millis) {
+                    chrono::LocalResult::Single(t) => t,
+                    _ => continue,
+                };
+                let exchange_rate = ExchangeRate {
+                    id: 0, // id is not used for inserts, only for reads.
+                    timestamp: timestamp,
```
enriquefynn, line 70: clippy: `replace with `timestamp`.

cli/listener/fuzz/fuzz_targets/apy_endpoint.rs
```diff
@@ -1,6 +1,102 @@
+// SPDX-FileCopyrightText: 2022 Chorus One AG
+// SPDX-License-Identifier: GPL-3.0
+
 #![no_main]
+
+use std::sync::{Arc, Mutex};
+
+use arbitrary::Arbitrary;
+use chrono;
```
enriquefynn, line 9: clippy: remove it entirely.

enriquefynn (approved):
LGTM!
I just have a few comments & nits

We're not running clippy on the new fuzz lib, we should include `cargo clippy --manifest-path cli/listener/fuzz/Cargo.toml -- --deny warnings` at `.github/workflows/build.yml`. I also marked them in this review

TIL: a bunch of things on fuzzers and how to use them in Rust, thanks @ruuda!

cli/listener/fuzz/README.md
```diff
@@ -0,0 +1,30 @@
...
+      -- -dict=fuzz/dictionary.txt
+
+Or run a faster fuzzer but with less intricate coverage:
+
+    cargo +nightly-2022-03-22 fuzz run apy_endpoint \
+      --jobs=«num-cores» \
+      --release \
+      --sanitizer=none \
+      --debug-assertions \
+      -- -dict=fuzz/dictionary.txt
```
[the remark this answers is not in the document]
ruuda: > This feature is experimental, right?

No it’s not experimental, it has been there for a long time. Note that the arguments after the `--` are for libFuzzer and not for `cargo-fuzz`, you can run it with `-- -help=1` to learn more. About dictionaries, [the docs are here](https://llvm.org/docs/LibFuzzer.html#dictionaries).

> Wonder which fields it tries to use the dict for

It doesn’t try “fields”, libFuzzer only generates bytestrings. It’s the `arbitrary` crate that generates struct instances from the bytestring. It does it in such a way that the mutations that libFuzzer applies tend to create, usually correspond to interesting mutations in the generated struct. You can think of `Arbitrary` as a specialized deserializer.

libFuzzer takes the dictionary into account in the mutations that it applies to the bytestring. It will sometimes splice entries from the dictionary into the bytestring, but it still has all of its other mutations, the dictionary just makes it faster. For example, for the string "begin", how would libFuzzer discover it? It would try random inputs until it hits a “b” and the `memcmp` gets one step further. That’s a new program path so this input is interesting, we mutate it further, and at some point it will hit the “e”, etc. But it will take `128 * len("begin")` iterations on average, if it was mutating at the right place, and it gets worse for longer strings, so we can kickstart it a bit with a dictionary. But even without the dictionary, it would eventually discover those paths.

cli/listener/fuzz/Cargo.toml
```diff
@@ -9,12 +9,20 @@ edition = "2018"
 cargo-fuzz = true
 
 [dependencies]
+arbitrary = { version = "1.1.0", features = ["derive"] }
```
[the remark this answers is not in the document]
ruuda: I don’t think it matters much, we will depend on the crate either way, I’ll keep it listed explicitly.

cli/listener/fuzz/fuzz_targets/apy_endpoint.rs
ruuda: Done, thanks.

cli/listener/fuzz/fuzz_targets/apy_endpoint.rs
ruuda: Done, thanks.

PR #555, opened 2022-04-08
Fix overflow in APY daemon discovered by fuzzer

ruuda:
In case of a very high epoch number, the computation for `start_slot + QUERY_SLOT_OFFSET` could overflow. This will not occur in practice, and it’s not triggered by a user input. It could only be triggered by a malicious RPC node. Still, we should fix it, if only to unblock the fuzzer for finding other issues.

### DIFF
--- a/cli/listener/src/lib.rs
+++ b/cli/listener/src/lib.rs
@@ -203,7 +203,13 @@ pub fn get_interval_price_for_period(
             Ok(epoch) => epoch,
             Err(_) => return Ok(None),
         };
-        let minimum_slot = epoch_schedule.get_first_slot_in_epoch(epoch) + QUERY_SLOT_OFFSET;
+        let minimum_slot = match epoch_schedule
+            .get_first_slot_in_epoch(epoch)
+            .checked_add(QUERY_SLOT_OFFSET)
+        {
+            Some(s) => s,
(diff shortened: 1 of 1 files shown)


enriquefynn:
Hum... I wonder if we should start fixing such edge-cases. Of course a malicious RPC node could answer bogus data, but a malicious RPC node also has opportunity to mess with the system in a lot of levels.
At a normal conditions, epoch is overflow will happen in billion of years, a time in which I hope there's something better than Solana out there 🙈

ruuda:
> Hum... I wonder if we should start fixing such edge-cases.

Yes, we should still fix it, even if this is not a problem in practice, for a few reasons:

* Most importantly, once the fuzzer discovers a crash, it will keep hitting that path, and it prevents further exploration. So fixing it is worth it even if the only thing it does is unblock the fuzzer.
* Things that appear hypothetical at first sight sometimes turn out to be exploitable in a different way after all. (A nice example of this: https://blog.neodyme.io/posts/lending_disclosure)
* Bounty hunters are going to be all over this and it will take a lot of energy to argue with them, it’s easier to just fix the issue.

enriquefynn (approved):
Fair enough, let's fix it then :)
LGTM!

PR #556, opened 2022-04-11
Add metrics for Anker deposit and withdrawals

enriquefynn:
Add and test metrics for deposits and withdrawals.
Add 3 fields for each function:
- Amount of stSOL
- Amount of bSOL
- count


enriquefynn:
Hum.. Now thinking about this. This PR makes Anker writable, so its deposit/withdraw transactions have to be serialized just for metrics, before this wasn't the case.

ruuda:
> This PR makes Anker writable, so its deposit/withdraw transactions have to be serialized just for metrics, before this wasn't the case.

Hmm, yeah that’s a fair point. But we can always stop collecting metrics later on if it turns out to be a problem, we can’t (easily) go back and collect the data in hindsight.

PR #558, opened 2022-04-13
Update solana checksum

enriquefynn:
The binaries were probably changed, would be better if we hosted ourselves, because if Solana updates past versions the build is not reproducible anymore.


ruuda (approved):
This is really scary ... does it change the hash of the on-chain programs?

ruuda:
We should upload the container to our container registry, so in case something like this happens again, we still have the image. We should have done that from the start ... Docker is only useful for making things reproducible if we save the image.

ruuda:
Ah, actually, I don’t think the upstream files changed ... we just forgot to update the checksum in https://github.com/ChorusOne/solido/commit/1bf16cf33a5f9dd9f8a74bb04dc9d04b73836407 :sweat_smile: 

Look at the annotated view for https://github.com/ChorusOne/solido/blame/471ea66243d3d02981005e9b8e9b45e65aa07bb2/docker/Dockerfile.dev, the version was changed recently but the checksum wasn’t.

PR #561, opened 2022-04-14
Write Anker metrics to prometheus

### LINKED ISSUE #450: [Anchor] Record statistics about deposits and withdrawals
In Solido we keep a histogram of deposits. This is quite useful, we should do the same for Anker. We may need to generalize the histogram over the token type then.

enriquefynn:
Write Anker metrics to prometheus so we can observe deposits/withdrawals

Closes #450

### DIFF
--- a/cli/common/src/prometheus.rs
+++ b/cli/common/src/prometheus.rs
@@ -336,7 +336,7 @@ pub fn write_anker_metrics_as_prometheus<W: io::Write>(
         &MetricFamily {
             name: "anker_swapped_rewards_st_sol_total",
             help: "Total amount of stSOL rewards swapped by our Anker instance.",
-            type_: "counter",
+            type_: "gauge",
             metrics: vec![Metric::new_st_sol(metrics.swapped_rewards_st_sol_total).at(at)],
         },
     )?;
@@ -345,7 +345,7 @@ pub fn write_anker_metrics_as_prometheus<W: io::Write>(
         &MetricFamily {
             name: "anker_swapped_rewards_ust_total",
             help: "Total amount of UST rewards swapped by our Anker instance.",
-            type_: "counter",
+            type_: "gauge",
             metrics: vec![Metric::new_ust_sol(metrics.swapped_rewards_ust_total).at(at)],
         },
     )?;
@@ -356,7 +356,7 @@ pub fn write_anker_metrics_as_prometheus<W: io::Write>(
         &MetricFamily {
             name: "anker_deposit_st_sol_total",
             help: "Total amount stSOL deposited into Anker",
-            type_: "counter",
+            type_: "gauge",
             metrics: vec![Metric::new_st_sol(metrics.deposit_metric.st_sol_total).at(at)],
         },
     )?;
@@ -365,7 +365,7 @@ pub fn write_anker_metrics_as_prometheus<W: io::Write>(
         &MetricFamily {
             name: "anker_deposit_b_sol_total",
             help: "Total amount bSOL minted in response to a deposit into Anker",
-            type_: "counter",
+            type_: "gauge",
             metrics: vec![Metric::new_b_sol(metrics.deposit_metric.b_sol_total).at(at)],
         },
     )?;
@@ -374,7 +374,7 @@ pub fn write_anker_metrics_as_prometheus<W: io::Write>(
         &MetricFamily {
             name: "anker_deposit_count_total",
             help: "Total number of deposits made by users on Anker.",
-            type_: "counter",
+            type_: "gauge",
             metrics: vec![Metric::new(metrics.deposit_metric.count).at(at)],
         },
     )?;
@@ -385,7 +385,7 @@ pub fn write_anker_metrics_as_prometheus<W: io::Write>(
         &MetricFamily {
             name: "anker_withdraw_st_sol_total",
             help: "Total amount of stSOL withdrawn in response from burning bSOL from Anker",
-            type_: "counter",
+            type_: "gauge",
             metrics: vec![Metric::new_st_sol(metrics.withdraw_metric.st_sol_total).at(at)],
         },
     )?;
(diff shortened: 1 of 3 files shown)


cli/common/src/prometheus.rs
```diff
@@ -85,6 +91,18 @@ impl<'a> Metric<'a> {
         Metric::new(MetricValue::Nano(amount.0))
     }
 
+    /// Construct a metric that measures an amount of UST.
+    pub fn new_ust_sol(amount: MicroUst) -> Metric<'a> {
```
ruuda, line 95:
```suggestion
    pub fn new_ust(amount: MicroUst) -> Metric<'a> {
```
?

cli/common/src/prometheus.rs
```diff
@@ -307,6 +326,91 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
...
+    metrics: &anker::metrics::Metrics,
+    at: SystemTime,
+    out: &mut W,
+) -> io::Result<()> {
+    write_metric(
+        out,
+        &MetricFamily {
+            name: "anker_swapped_rewards_st_sol_total",
+            help: "Total amount of stSOL rewards swapped by our Anker instance.",
+            type_: "counter",
```
ruuda, line 339:
```suggestion
            type_: "gauge",
```

This is data that comes from the chain, not from inside the daemon itself, right? In that case it should be a gauge, not a counter. The value might go down for example when the daemon connects to a loadbalanced RPC node, and one of the RPC nodes behind the loadbalancer is a few slots behind. Then we can get this flip-flopping behavior, and Prometheus should not treat that as a counter reset.

cli/common/src/prometheus.rs
```diff
@@ -307,6 +326,91 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
...
+            metrics: vec![Metric::new_st_sol(metrics.swapped_rewards_st_sol_total).at(at)],
+        },
+    )?;
+    write_metric(
+        out,
+        &MetricFamily {
+            name: "anker_swapped_rewards_ust_total",
+            help: "Total amount of UST rewards swapped by our Anker instance.",
+            type_: "counter",
```
ruuda, line 348:
```suggestion
            type_: "gauge",
```
Also here.

cli/common/src/prometheus.rs
```diff
@@ -307,6 +326,91 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
...
+        },
+    )?;
+
+    // Deposit metrics
+    write_metric(
+        out,
+        &MetricFamily {
+            name: "anker_deposit_st_sol_total",
+            help: "Total amount stSOL deposited into Anker",
+            type_: "counter",
```
ruuda, line 359:
```suggestion
            type_: "gauge",
```

cli/common/src/prometheus.rs
```diff
@@ -307,6 +326,91 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
...
+            metrics: vec![Metric::new_st_sol(metrics.deposit_metric.st_sol_total).at(at)],
+        },
+    )?;
+    write_metric(
+        out,
+        &MetricFamily {
+            name: "anker_deposit_b_sol_total",
+            help: "Total amount bSOL minted in response to a deposit into Anker",
+            type_: "counter",
```
ruuda, line 368:
```suggestion
            type_: "gauge",
```

cli/common/src/prometheus.rs
```diff
@@ -307,6 +326,91 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
...
+            metrics: vec![Metric::new_b_sol(metrics.deposit_metric.b_sol_total).at(at)],
+        },
+    )?;
+    write_metric(
+        out,
+        &MetricFamily {
+            name: "anker_deposit_count_total",
+            help: "Total number of deposits made by users on Anker.",
+            type_: "counter",
```
ruuda, line 377:
```suggestion
            type_: "gauge",
```

cli/common/src/prometheus.rs
```diff
@@ -307,6 +326,91 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
...
+        },
+    )?;
+
+    // Withdraw metrics
+    write_metric(
+        out,
+        &MetricFamily {
+            name: "anker_withdraw_st_sol_total",
+            help: "Total amount of stSOL withdrawn in response from burning bSOL from Anker",
+            type_: "counter",
```
ruuda, line 388:
```suggestion
            type_: "gauge",
```

cli/common/src/prometheus.rs
```diff
@@ -307,6 +326,91 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
...
+            metrics: vec![Metric::new_st_sol(metrics.withdraw_metric.st_sol_total).at(at)],
+        },
+    )?;
+    write_metric(
+        out,
+        &MetricFamily {
+            name: "anker_withdraw_b_sol_total",
+            help: "Total amount of bSOL burned in response from withdrawing stSOL from Anker",
+            type_: "counter",
```
ruuda, line 397:
```suggestion
            type_: "gauge",
```

cli/common/src/prometheus.rs
```diff
@@ -307,6 +326,91 @@ pub fn write_solido_metrics_as_prometheus<W: io::Write>(
...
+            metrics: vec![Metric::new_b_sol(metrics.withdraw_metric.b_sol_total).at(at)],
+        },
+    )?;
+    write_metric(
+        out,
+        &MetricFamily {
+            name: "anker_withdraw_count_total",
+            help: "Total number of withdrawals made by users on Anker.",
+            type_: "counter",
```
ruuda, line 406:
```suggestion
            type_: "gauge",
```

ruuda (commented):
I have a few small comments, LGTM aside from those!

cli/common/src/prometheus.rs
enriquefynn: Copy-paste bug

PR #563, opened 2022-05-04
Make dependencies compatible with "cargo vendor"

ruuda:
We need to patch all the dependencies from `solana-*`, to make sure we don't mix packages from crates.io and our fork. Without this, `cargo vendor` will complain with an error like this:

```
$ cargo vendor
error: failed to sync

Caused by:
  found duplicate version of package `solana-metrics v1.7.8` vendored from two sources:

  	source 1: registry `crates-io`
  	source 2: https://github.com/ChorusOne/solana?branch=program-test-178#af358e1a
```

I listed all of these under patches until Cargo stopped complaining.

This is needed in particular to make this work with `buildRustPackage` in Nix, because it relies on `cargo vendor`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1950,7 +1950,7 @@ dependencies = [
  "solana-program",
  "solana-program-test",
  "solana-sdk",
- "solana-vote-program 1.7.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


ruuda:
Opened #565 for updating Solana.

PR #564, opened 2022-05-04
Add changelog for v1.3.2

ruuda:
Also bump the version in the Cargo.toml files.


enriquefynn (approved):
LGTM!

I thought that nix was failing because of the patched deps. But you didn't change those here.

enriquefynn:
Oh, I see this is done after #563, Great!

PR #567, opened 2022-05-12
Remove extra call to `check_mint`

### LINKED ISSUE #560: `process_withdraw` duplicates `check_mint` check.
Speaking with [kkonevets](https://github.com/kkonevets),  thanks for the report 👍, he found out that we check twice for the mint at:

https://github.com/ChorusOne/solido/blob/0d9a19cc030bafc86a138524bf56918e68acf1e5/anker/src/processor.rs#L410

It doesn't impact our deployment 😅 but it does consume a bit more computing, or maybe the compiler is smart enough to remove/optimize this.
Anyways we should remove the duplicate check.

enriquefynn:
Closes #560

### DIFF
--- a/anker/src/processor.rs
+++ b/anker/src/processor.rs
@@ -407,7 +407,6 @@ fn process_withdraw(
     anker.check_is_st_sol_account(&solido, accounts.reserve_account)?;
     anker.check_mint(accounts.b_sol_mint)?;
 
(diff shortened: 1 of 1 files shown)


ruuda (approved):
LGTM!

I thought this was going to be super tricky to review, removing an authorization check, but it turns out it literally had the same check two lines above, so this is safe.

PR #569, opened 2022-06-04
Transition to public validators with arbitrary commission

kkonevets:
We need to be able to add public validators to our pull with arbitrary commission. Now validation fee is 100%, which means all rewards go to validator vote account which Solido has withdraw authority from. Users should see a reasonable operational fee, e.g. 7%, which consists of validator, developer and treasury fees. 
The main idea is to let Solana pay to a public validator vote account directly and keep the rest of the rewards in stake accounts which are governed by Solido. Solido will have no control over the vote account. Treasury and developer fees are minted per validator as a percentage of a sum over rewards in validator stake accounts. User’s appreciation fee is kept in stake accounts, so it’s automatically compounded.

**Main changes**:
- add `max_validation_fee` to Solido state to limit maximum validation rewards
- remove rewards withdraw authority from Solido state
- remove `fee_credit` and `fee_address` from Validator,  let Solana pay validation fees directly to vote accounts in SOL
- remove `ClaimValidatorFee` and `CollectValidatorFee` instructions. Developer and treasury fee minting logic is moved from `CollectValidatorFee` to `WithdrawInactiveStake` instruction
- calculate  rewards in `WithdrawInactiveStake` per validator and mint developer and treasury fees for this validator as a portion of rewards in stake accounts. If we do so for all validators we will mint all developer and treasury fees in sum. We mint per validator because Solana has a limit of approximately max 20 accounts referenced in a single transaction, so we can’t reference all vote accounts and their stake accounts in a single transaction. Appreciation fees are kept at stake accounts as is, which means they are automatically compounded.
- add new permissionless instruction `CheckMaxCommissionViolation` which marks validator as inactive if he sets his commission over `max_validation_fee`, add maintenance function that periodically triggers `CheckMaxCommissionViolation`, add CLI entry and python test for this instruction.
- modify all tests and cli tools to take account of the changes

**TODO**:
- Add an instruction to convert on-chain state from previous version to the new which will arrive as a separate pull request
- Add a multisig instruction to change `max_validation_fee`
- We need to bypass a CPU and memory threshold on a number of possible validators in the pull, currently 64. For this we will integrate Solana stake-pool [BigVec](https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/src/big_vec.rs) data structure to efficiently serialise/deserialise lists of validators and maintainers.
- We will not know Solido validator fee metrics (e.g. `fee_validation_total_lamports`), because a public validator can have stake accounts beyond Solido. So we can approximately compute it from amount of Solido stake


program/src/accounts.rs
```diff
@@ -340,9 +340,7 @@ macro_rules! accounts_struct {
 
                     // Also consume the iterator, so the no-excess-accounts check
                     // below does not trigger.
-                    for _ in 0..$multi_account.len() {
-                        accounts_iter.next();
-                    }
+		    let mut accounts_iter = accounts_iter.skip($multi_account.len());
```
ruuda, line 343: You are re-defining `accounts_iter` here ... does it actually remain in scope (or rather, escape the `$(...)?` scope), such that the next `accounts_iter` reference below references the skipped version, and not the original one?

I found the former version to be easier to understand, because `.skip` does not actually mutate the existing iterator to skip that many elements, it wraps the iterator in a _new_ one.

program/src/instruction.rs
```diff
@@ -32,6 +31,8 @@ pub enum LidoInstruction {
         max_validators: u32,
         #[allow(dead_code)] // but it's not
         max_maintainers: u32,
+        #[allow(dead_code)] // but it's not
+        max_validation_fee: u8,
```
ruuda, line 35: If we add a new configuration variable, we also need to add an instruction so the manager can change it, and add all the boilerplate in the CLI so we can propose this instruction to the multisig and pretty-print it.

program/src/instruction.rs
```diff
@@ -101,6 +100,12 @@ pub enum LidoInstruction {
     AddMaintainer,
     RemoveMaintainer,
     MergeStake,
+
+    /// Check if validator increased his commission over maximum allowed
+    /// and deactivate him if he did
+    ///
+    /// Requires no permission
+    CheckMaxCommissionViolation,
```
ruuda, line 108:
```suggestion
    DeactivateValidatorIfCommissionExceedsMax,
```

How about naming it after what it does?

program/src/instruction.rs
```diff
@@ -109,7 +114,7 @@ impl LidoInstruction {
         // `Borsh::serialize`, which takes an arbitrary writer, and which can
         // therefore return an IoError. But when serializing to a vec, there
         // is no IO, so for this particular writer, it should never fail.
-        self.try_to_vec()
+        self.try_to_vec() // actually it can fail on OutOfMemory error
```
ruuda, line 117: Out of memory should be unrecoverable, so that is fine.

(For very specialized use cases, a long-running daemon that is essential to the system might want to handle memory exhaustion, e.g. to do an emergency shutdown and persist some state to disk. But for programs running on a blockchain, it makes no sense to handle OOM; if you can edit the program to handle OOM, you might as well edit the program to make it more resource-efficient in the first place.)

program/src/instruction.rs
```diff
@@ -69,11 +70,9 @@ pub enum LidoInstruction {
     /// Observe any external changes in the balances of a validator's stake accounts.
     ///
     /// If there is inactive balance in stake accounts, withdraw this back to the reserve.
+    /// Distribute fees.
     WithdrawInactiveStake,
 
-    /// Claim rewards from the validator account and distribute rewards.
-    CollectValidatorFee,
-    ClaimValidatorFee,
```
ruuda, line 76: We can’t remove elements from this enum, because it would change the serialization format. A Lido instruction starts with a discriminant byte that indicates the instruction; removing these two would change that byte for all instructions that follow here, which would break everything that submits transactions.

Instead, we should mark them as deprecated. We can replace the handlers for these instructions with no-op handlers that make the transaction fail with a message like “this instruction is no longer supported”. But we need to keep them around in the enum forever.

program/src/instruction.rs
```diff
@@ -140,7 +145,6 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
-        const sysvar_rent = sysvar::rent::id(),
```
ruuda, line 143: In general we can’t modify existing instructions, but because `initialize` can’t be called on an existing instance anyway, in this case this is fine.

program/src/instruction.rs
```diff
@@ -419,8 +425,6 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
-        const sysvar_clock = sysvar::clock::id(),
-        const sysvar_rent = sysvar::rent::id(),
```
ruuda, line 423: We can’t modify existing instructions; that is a breaking change for all existing things that call this instruction due to https://github.com/ChorusOne/solido/blob/b4ba3f7919b9f07b4dad6be7fca1c290467a5328/program/src/accounts.rs#L237-L240

The proper way to deal with this is to append a new instruction, `UpdateExchangeRateV2`, that is the same as v1, but with these two input accounts omitted. In this case, because the new implementation needs only a subset of the accounts, we can transparently handle this in the instruction dispatcher, a call to `UpdateExchangeRate` v1 would construct an `UpdateExchangeRateV2AccountsInfo` struct and then call the implementation for v1.

program/src/instruction.rs
```diff
@@ -463,55 +467,6 @@ accounts_struct! {
             is_writable: true,
         },
 
-        // We only allow updating balances if the exchange rate is up to date,
```
ruuda, line 466: Also here, we can’t modify existing instructions because that is a breaking change, we need to add _new_ instructions instead.

If it is possible to upgrade the old call in-place in the instruction dispatcher that is nice, but if that is not possible or not worth the effort, we can also make the old instruction fail. But we should not change the meaning of the existing instruction.

program/src/instruction.rs
```diff
@@ -537,41 +492,36 @@ accounts_struct! {
...
-            // Is writable due to withdraw to reserve (vote_instruction::withdraw)
-            is_writable: true,
-        },
-        // Used to get the rewards out of the validator vote account.
-        pub rewards_withdraw_authority {
-            is_signer: false,
-            is_writable: false,
-        },
+        // Needed for minting rewards.
+        const spl_token_program = spl_token::id(),
```
ruuda, line 496: Also here; we can’t change the meaning of the existing instruction, we should add a new instruction instead.

program/src/instruction.rs
```diff
@@ -629,11 +579,6 @@ accounts_struct! {
             is_signer: false,
             is_writable: false,
         },
-        pub validator_fee_st_sol_account {
-            is_signer: false,
-            is_writable: false,
-        },
-        const sysvar_rent = sysvar::rent::id(),
```
ruuda, line 636: In this case, because nobody external can call this instruction anyway, it _might_ be fine to edit it in-place, but it will still cause trouble. For example, if you use the `solido multisig show-transaction` command to inspect one of the transactions that added the initial Lido validators (to v1), then v2 of the `solido` program will no longer be able to parse that transaction.

We could accept this and say that you need `solido` 1.x to be able to inspect the transaction, but there is no way for people to figure that out from just looking at the multisig transaction's bytes.

It would still be nicer to not modify the existing transactions, and instead add a `AddValidatorV2` instruction. The implementation of `AddValidator` v1 can be replaced with just a panic that explains the instruction is deprecated, but we should keep the definition around.

ruuda (commented):
I looked at `instruction.rs`. One thing that stands out is that you modified the existing instructions. We can’t do that — if something external is calling these instructions, it would change the meaning under their feet. It also complicates things for block explorers, e.g. if you see discriminant 0x06, what kind of instruction it is depends on whether it happend before or after slot _x_ ...

Instead, we should leave the existing instructions untouched, and only add _new_ instructions at the end. We can replace the implementations of some instructions to fail the transaction, now that they are deprecated, but we need to keep the definition around.

I haven’t looked at the other files yet, I think this will cause a rather large (but trivial) diff so I’ll continue the review after this has been addressed.

program/src/accounts.rs
kkonevets: It shadows previous definition of `accounts_iter`. 
`std::slice::Skip` also implements same Iterator trait as `std::slice::Iter`, so no difference and it's easier to read. But if you insist, I will revert it back

program/src/instruction.rs
ruuda: It’s fine to do it in a follow-up, that makes this PR easier to review as well. We just shouldn’t forget it.

kkonevets:
@ruuda @enriquefynn Please check this PR. I had to add `SetMaxValidationFee` instruction to this PR, otherwise juggling interdependent PRs would be cumbersome.
@Popeye1357

ruuda (commented):
I looked at most of the files now, but mostly the boilerplate and routine stuff. I haven’t verified the changes to the logic and processor yet. That needs a very careful review and I need to do that with a clear head with my full attention, will try to do that first thing tomorrow morning.

The changes I saw so far all look good. I have some superficial comments, but I didn’t find any structural issues in the parts I looked at, the logic is sound. Thanks for also updating the maintainer to immediately unstake from ill-behaved validators.

Thanks also for adding the v2 instructions. The diff also looks a lot cleaner like this :-)

program/src/instruction.rs
```diff
@@ -815,6 +848,169 @@ pub fn merge_stake(program_id: &Pubkey, accounts: &MergeStakeMeta) -> Instructio
...
+            is_writable: false,
+        },
+        pub validator_vote_account {
+            is_signer: false,
+            is_writable: false,
+        },
+    }
+}
+
+pub fn add_validator_v2(program_id: &Pubkey, accounts: &AddValidatorMetaV2) -> Instruction {
```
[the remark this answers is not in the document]
kkonevets: Good point, haven't noticed that. Will fix in next commit.

program/src/processor.rs
```diff
@@ -818,76 +825,16 @@ pub fn process_withdraw_inactive_stake(
...
-            accounts.vote_program.clone(),
-        ],
-        &[&[
-            accounts.lido.key.as_ref(),
-            REWARDS_WITHDRAW_AUTHORITY,
-            &[lido.rewards_withdraw_authority_bump_seed],
-        ]],
-    )?;
-    lido.save(accounts.lido)
+    Err(LidoError::InstructionIsDeprecated.into())
```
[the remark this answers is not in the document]
kkonevets: Ok

cli/maintainer/src/daemon.rs
```diff
@@ -108,12 +106,8 @@ impl MaintenanceMetrics {
                         .with_label("operation", "UpdateExchangeRate".to_string()),
                     Metric::new(self.transactions_withdraw_inactive_stake)
                         .with_label("operation", "WithdrawInactiveStake".to_string()),
-                    Metric::new(self.transactions_collect_validator_fee)
-                        .with_label("operation", "CollectValidatorFee".to_string()),
                     Metric::new(self.transactions_merge_stake)
                         .with_label("operation", "MergeStake".to_string()),
-                    Metric::new(self.transactions_claim_validator_fee)
-                        .with_label("operation", "ClaimValidatorFee".to_string()),
```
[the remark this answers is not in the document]
kkonevets: yeah, forgot to add this:
```rust
                    Metric::new(self.transactions_deactivate_validator_if_commission_exceeds_max)
                        .with_label(
                            "operation",
                            "DeactivateValidatorIfCommissionExceedsMax".to_string(),
                        ),
                ],

```

cli/maintainer/src/config.rs
```diff
@@ -458,6 +454,42 @@ cli_opt_struct! {
...
+    DeactivateValidatorIfCommissionExceedsMaxOpts {
+        /// Address of the Solido program.
+        #[clap(long, value_name = "address")]
+        solido_program_id: Pubkey,
+
+        /// Account that stores the data for this Solido instance.
+        #[clap(long, value_name = "address")]
+        solido_address: Pubkey,
+    }
+}
```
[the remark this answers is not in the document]
kkonevets: If you are talking about `DeactivateValidatorIfCommissionExceedsMax` then as a user I would like to have an explicit control over deactivation of validators that exceed max validation fee. Also it prints such validators, which could be useful to someone. It's better to have more options than less. And it's convenient for testing - I can explicitly run the instruction with python.
I have removed a signer. Will be visible in a next commit.

cli/maintainer/src/commands_solido.rs
```diff
@@ -346,14 +347,14 @@ pub struct ShowSolidoOutput {
...
-    pub rewards_withdraw_authority: Pubkey,
-
     /// Identity account address for all validators in the same order as `solido.validators`.
     pub validator_identities: Vec<Pubkey>,
 
     /// Contains validator info in the same order as `solido.validators`.
     pub validator_infos: Vec<ValidatorInfo>,
+
+    /// Contains validator fees
+    pub validator_fees: Vec<u8>,
```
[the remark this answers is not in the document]
kkonevets: Accepted. Yes, I thing we should rename `validation fee` to `commission`. But for now I just renamed `validator_fees ` to `validator_commission_percentages `

cli/maintainer/src/commands_solido.rs
```diff
@@ -346,14 +347,14 @@ pub struct ShowSolidoOutput {
...
-    #[serde(serialize_with = "serialize_b58")]
-    pub rewards_withdraw_authority: Pubkey,
-
     /// Identity account address for all validators in the same order as `solido.validators`.
     pub validator_identities: Vec<Pubkey>,
 
     /// Contains validator info in the same order as `solido.validators`.
     pub validator_infos: Vec<ValidatorInfo>,
+
+    /// Contains validator fees
```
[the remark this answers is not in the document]
kkonevets: Accepted

cli/maintainer/src/commands_solido.rs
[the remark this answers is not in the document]
kkonevets: No. I renamed it everywhere.

cli/maintainer/src/commands_solido.rs
```diff
@@ -892,3 +877,96 @@ pub fn command_withdraw(
...
+
+    let mut violations = vec![];
+    for pubkey_entry in solido.validators.entries {
+        let validator = pubkey_entry.entry;
+        let vote_pubkey = pubkey_entry.pubkey;
+        let validator_account = config.client.get_account(&vote_pubkey)?;
+        let commission = get_vote_account_commission(&validator_account.data)
+            .ok_or_else(|| CliError::new("Validator account data too small"))?;
+
+        // dbg!(pubkey_entry.pubkey, commission, validator.active);
```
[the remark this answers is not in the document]
kkonevets: done

cli/maintainer/src/commands_solido.rs
```diff
@@ -892,3 +877,96 @@ pub fn command_withdraw(
...
+            validator_vote_account: vote_pubkey,
+            validation_fee: commission,
+        });
+    }
+
+    Ok(DeactivateValidatorIfCommissionExceedsMaxOutput {
+        entries: violations,
+        max_validation_fee: solido.max_validation_fee,
+    })
+}
```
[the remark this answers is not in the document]
kkonevets: I answered earlier why we should keep it

cli/maintainer/src/commands_multisig.rs
```diff
@@ -655,6 +655,16 @@ impl fmt::Display for ShowTransactionOutput {
                         print_changed_reward_distribution(f, current_solido, reward_distribution)?;
                         print_changed_recipients(f, current_solido, fee_recipients)?;
                     }
+                    SolidoInstruction::SetMaxValidationFee {
+                        solido_instance,
+                        max_validation_fee,
+                        manager,
+                    } => {
+                        writeln!(f, "It sets a maximun validation fee")?;
```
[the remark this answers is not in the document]
kkonevets: ok

cli/maintainer/src/commands_multisig.rs
```diff
@@ -655,6 +655,16 @@ impl fmt::Display for ShowTransactionOutput {
...
+                        writeln!(f, "    Solido instance:    {}", solido_instance)?;
+                        writeln!(f, "    Manager:            {}", manager)?;
+                        writeln!(f, "    Max validation fee: {}", max_validation_fee)?;
```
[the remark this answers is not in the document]
kkonevets: done

cli/maintainer/src/commands_multisig.rs
```diff
@@ -655,6 +655,16 @@ impl fmt::Display for ShowTransactionOutput {
                         print_changed_reward_distribution(f, current_solido, reward_distribution)?;
                         print_changed_recipients(f, current_solido, fee_recipients)?;
                     }
+                    SolidoInstruction::SetMaxValidationFee {
+                        solido_instance,
+                        max_validation_fee,
```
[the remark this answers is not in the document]
kkonevets: I renamed it to `max_validation_commission`

program/src/state.rs
```diff
@@ -26,15 +26,15 @@ use crate::{
     account_map::{AccountMap, AccountSet, EntryConstantSize, PubkeyAndEntry},
     MINIMUM_STAKE_ACCOUNT_BALANCE, MINT_AUTHORITY, RESERVE_ACCOUNT, STAKE_AUTHORITY,
 };
-use crate::{REWARDS_WITHDRAW_AUTHORITY, VALIDATOR_STAKE_ACCOUNT, VALIDATOR_UNSTAKE_ACCOUNT};
+use crate::{VALIDATOR_STAKE_ACCOUNT, VALIDATOR_UNSTAKE_ACCOUNT};
 
 pub const LIDO_VERSION: u8 = 0;
```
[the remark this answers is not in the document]
kkonevets: O yeah! I forgot it. Done

program/tests/tests/max_validation_fee.rs
```diff
@@ -0,0 +1,37 @@
...
+    assert_eq!(solido.max_validation_fee, context.max_validation_fee + 1);
+
+    let result = context.try_deactivate_validator_if_commission_exceeds_max(validator.pubkey);
+    assert_eq!(result.await.is_ok(), true);
+
+    // check validator is not deactivated
+    let validator = &context.get_solido().await.validators.entries[0];
+    assert_eq!(validator.entry.active, true);
+
+    // increase max_validation_fee abouve 100%
```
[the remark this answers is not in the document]
kkonevets: done

program/tests/tests/max_validation_fee.rs
```diff
@@ -0,0 +1,37 @@
...
+    assert_eq!(context.try_set_max_validation_fee(101).await.is_err(), true);
```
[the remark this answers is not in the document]
kkonevets: done

program/src/vote_instruction.rs
```diff
@@ -13,9 +13,7 @@ use serde_derive::{Deserialize, Serialize};
 use solana_program::{
     clock::{Slot, UnixTimestamp},
     hash::Hash,
-    instruction::{AccountMeta, Instruction},
     pubkey::Pubkey,
-    vote,
```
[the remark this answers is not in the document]
kkonevets: You are right, we don't need `vote_instruction.rs` anymore. I deleted it.

kkonevets:
There is weird error in CI on tests::fetch_pool_price::test_fail_fetch_pool_price_too_early `panicked at 'Could not send transaction to fetch pool price.: IoError(Custom { kind: Other, error: "the request exceeded its deadline" })`
On my local machine no errors are observed.

kkonevets:
Yes, the same commit on following push was successful. All bpf tests succeeded. So GitHub CI was a bottleneck. I guess GitHub's test workload is not constant and depends on a time of a day.
@ruuda I have fixed everything you commented on, please review.

program/src/processor.rs
```diff
@@ -649,17 +649,17 @@ pub fn check_address_and_get_balance(
...
 /// the reserve account.
-/// Updates the validator's balance.
+/// Updates the validator's balance and distribute rewards.
 /// This function is permissionless and can be called by anyone.
 pub fn process_withdraw_inactive_stake(
     program_id: &Pubkey,
     raw_accounts: &[AccountInfo],
 ) -> ProgramResult {
-    let accounts = WithdrawInactiveStakeInfo::try_from_slice(raw_accounts)?;
+    let accounts = WithdrawInactiveStakeInfoV2::try_from_slice(raw_accounts)?;
```
ruuda, line 658: Now that rewards show up primarily as increased _active_ balance of the stake accounts, this instruction does more than just withdrawing inactive stake, it is _the_ primary mechanism through which fees get distributed. I think it would be clearer for readers if we reflect that in the name, maybe `UpdateStakeAccountBalance` or something like this.

program/src/processor.rs
```diff
@@ -723,18 +723,14 @@ pub fn process_withdraw_inactive_stake(
...
+    // withdraw from these accounts, so we should not observe a decrease in
+    // balance.
+    Validator::observe_balance(
+        stake_observed_total,
+        validator.entry.effective_stake_balance(),
+        "Stake",
+    )?;
 
     // We tracked in `stake_accounts_balance` what we put in there ourselves, so
     // the excess is a donation by some joker.
```
ruuda, line 736: This comment is no longer true, now that we no longer work with 100%-commission accounts, the excess also represents staking rewards that are paid out by Solana.

program/src/processor.rs
```diff
@@ -1049,20 +982,31 @@ pub fn process(program_id: &Pubkey, accounts: &[AccountInfo], input: &[u8]) -> P
         }
         LidoInstruction::Unstake { amount } => process_unstake(program_id, amount, accounts),
         LidoInstruction::UpdateExchangeRate => process_update_exchange_rate(program_id, accounts),
-        LidoInstruction::WithdrawInactiveStake => {
+        LidoInstruction::WithdrawInactiveStake
+        | LidoInstruction::CollectValidatorFee
+        | LidoInstruction::ClaimValidatorFee
+        | LidoInstruction::AddValidator => {
+            msg!("Please use a new instruction {:?}", instruction);
```
ruuda, line 989:
```suggestion
            msg!("{:?} is no longer supported since v2. Please check the changelog in the repository for update instructions.", instruction);
```

We should also list those breaking changes in `CHANGELOG.md` so we don’t forget to do it later. We can add a preliminary section there, something like

```markdown
## v2.0.0 (unreleased)

New features:

* Solido no longer requires that validators use a 100%-commission account of which Solido
  is the withdraw authority. Any vote account can now be used, as long as its commission does
  not exceed Solido’s configured maximum commission percentage.

**Compatibility**

* The `AddValidator` instruction is no longer supported and has been superseded by `AddValidatorV2`.
* ...
```

program/tests/tests/withdraw_inactive_stake.rs
```diff
@@ -69,4 +71,94 @@ async fn test_withdraw_inactive_stake() {
...
+    let rewards = (validator_after.entry.stake_accounts_balance
+        - validator_before.entry.stake_accounts_balance)
+        .expect("Does not underflow, because we received rewards.");
+    assert_eq!(rewards, Lamports(arbitrary_rewards));
+
+    let validation_commission = (vote_account_after - vote_account_before).unwrap();
+    // validation commission is 5% of total rewards, solido_rewards is 95% of total rewards
+    assert_eq!(
+        validation_commission,
+        Lamports((5f64 * (rewards.0 as f64) / 95f64) as u64)
```
ruuda, line 144:
```suggestion
        Lamports((5.0 * (rewards.0 as f64) / 95.0) as u64)
```
I think this should also work because the inner `f64` will make the float literals infer to `f64` as well. In my opinion that's a bit more readable than the `f64` suffix.

program/tests/tests/withdraw_inactive_stake.rs
```diff
@@ -69,4 +71,94 @@ async fn test_withdraw_inactive_stake() {
...
+    );
+
+    // The treasury balance increase, when converted back to SOL, should be equal
+    // to 3% of the rewards. Three lamports differ due to rounding errors.
```
ruuda, line 148: If we swap the order of the division and multiplication below, the result will be more accurate, because the division is an integer divison that truncates downward, so divide-then-multiply multiplies the error in the division.

In general when working with fixed-point math, you want to do the divisions as late as possible, as long as the prior calculations don’t overflow.

program/tests/tests/withdraw_inactive_stake.rs
```diff
@@ -69,4 +71,94 @@ async fn test_withdraw_inactive_stake() {
...
+    let treasury_fee = (treasury_after - treasury_before).unwrap();
+    let treasury_fee_sol = solido_after
+        .exchange_rate
+        .exchange_st_sol(treasury_fee)
+        .unwrap();
+    assert_eq!(treasury_fee_sol, Lamports(rewards.0 / 100 * 3 + 1));
```
ruuda, line 154:
```suggestion
    assert_eq!(treasury_fee_sol, Lamports(rewards.0 * 3 / 100 + 1));
```

program/tests/tests/withdraw_inactive_stake.rs
```diff
@@ -69,4 +71,94 @@ async fn test_withdraw_inactive_stake() {
...
+
+    // The developer balance increase, when converted back to SOL, should be equal
+    // to 2% of the rewards. Two lamport differ due to rounding errors.
+    let developer_fee = (developer_after - developer_before).unwrap();
+    let developer_fee_sol = solido_after
+        .exchange_rate
+        .exchange_st_sol(developer_fee)
+        .unwrap();
+    assert_eq!(developer_fee_sol, Lamports(rewards.0 / 100 * 2));
```
ruuda, line 163:
```suggestion
    assert_eq!(developer_fee_sol, Lamports(rewards.0 * 2 / 100));
```

program/tests/tests/solana_assumptions.rs
```diff
@@ -110,7 +110,8 @@ async fn measure_staking_rewards(mode: StakeMode) -> Lamports {
     // Move ahead one epoch so the stake becomes active.
     context.advance_to_normal_epoch(1);
 
-    let balance_t0 = context.get_sol_balance(vote_account).await;
+    let mut balance_t0 = context.get_sol_balance(vote_account).await;
+    balance_t0 = (balance_t0 + context.get_sol_balance(stake_account).await).unwrap();
```
ruuda, line 114: No need for this to be mutable, we can define it as a sum right away.

program/tests/tests/solana_assumptions.rs
```diff
@@ -128,7 +129,8 @@ async fn measure_staking_rewards(mode: StakeMode) -> Lamports {
         .increment_vote_account_credits(&vote_account, 1);
     context.advance_to_normal_epoch(2);
 
-    let balance_t1 = context.get_sol_balance(vote_account).await;
+    let mut balance_t1 = context.get_sol_balance(vote_account).await;
+    balance_t1 = (balance_t1 + context.get_sol_balance(stake_account).await).unwrap();
```
ruuda, line 133: Also here.

ruuda (commented):
I have two files left to review, `test_solido.py` and `state.rs`. The logic related to fee distribution looks good from what I can tell. I still need to sit down and think a bit about how fee distribution and the exchange rate update interact now, if this changes anything. I think it should be fine, but I want to confirm that still.

program/src/state.rs
```diff
@@ -315,10 +317,7 @@ impl Lido {
 
     /// Checks if the passed maintainer belong to the list of maintainers
     pub fn check_maintainer(&self, maintainer: &AccountInfo) -> ProgramResult {
-        if !&self.maintainers.entries.contains(&PubkeyAndEntry {
-            pubkey: *maintainer.key,
-            entry: (),
-        }) {
+        if self.maintainers.get(maintainer.key).is_err() {
```
ruuda, line 320: Nice simplification :+1:

program/src/state.rs
```diff
@@ -1251,60 +1173,54 @@ mod test_lido {
...
+                developer_amount: Lamports(150),
                 st_sol_appreciation_amount: Lamports(0),
             },
         );
 
         assert_eq!(
             // In this case the amount cannot be split exactly, all fees are
             // rounded down.
-            spec.split_reward(Lamports(1_000), 4).unwrap(),
+            spec.split_reward(Lamports(1_000)).unwrap(),
```
ruuda, line 1194:
```suggestion
            spec.split_reward(Lamports(1_003)).unwrap(),
```

The point of this test case was to have an amount that is not a multiple of the denominator, such that the split cannot be exact. In that case all fees should round down, and the remainder goes to stSOL appreciation.

Previously the fee shares summed to 6, and 1000 is not a multiple of 6, so this test exercised the rounding. But now the fee shares sum to 4, and 1000 is a multiple of 4. We can change the amount to split to e.g. 1003. Then we have 3 additional lamports to distribute. 3/4 = 2.25 of those should go to the treasury, which then rounds down to 2. 1/4 = 0.75 of those should go to the developer, which rounds down to 0. But then so far we distributed 2 lamports, and we had 3 to distribute, so the remaining 1 lamport goes to stSOL appreciation.

tests/test_solido.py
```diff
@@ -208,47 +209,34 @@ def approve_and_execute(
...
-    f'tests/.keys/validator-token-account-key.json', st_sol_mint_account
-)
-print(f'> Validator stSol token account: {fee_account}')
-
 
 class Validator(NamedTuple):
     account: TestAccount
     vote_account: TestAccount
-    fee_account: str
+    withdrawer_account: TestAccount
```
ruuda, line 220: I think I mentioned this elsewhere already, but we can also make the authorized withdrawer equal to the vote account itself, that should simplify the tests a bit.

ruuda (commented):
I reviewed the final two files. Phew, it is done! :grinning:  I only found a few minor things.

cli/maintainer/src/commands_solido.rs
```diff
@@ -892,3 +875,100 @@ pub fn command_withdraw(
...
+
+        let instruction = lido::instruction::deactivate_validator_if_commission_exceeds_max(
+            opts.solido_program_id(),
+            &lido::instruction::DeactivateValidatorIfCommissionExceedsMaxMeta {
+                lido: *opts.solido_address(),
+                validator_vote_account_to_deactivate: vote_pubkey,
+            },
+        );
+        let signers: Vec<&dyn Signer> = vec![];
+        config.sign_and_send_transaction(&[instruction], &signers)?;
```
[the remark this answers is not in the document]
kkonevets: Agree, that's a problem. Why do you think that it is not worth dealing with "we do need to be careful that the transaction can fit that many instructions"? Because it is improbable to have so many misbehaved validators?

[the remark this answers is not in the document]
ruuda: Yes exactly, the probability that we hit that case in production is extremely low because:

* There need to be multiple misbehaving validators at the same time, and even a single validator misbehaving should be very rare.
* Maintainers would already deactivate the misbehaving validator before we can run this, and we have 4 independent parties operating an instance.
* The instruction is small, it needs to reference only 3 accounts, so you can probably fit more than 3 of them in a single transaction.

And even if it does happen, it’s not the end of the world if we need to edit the code a bit and recompile and it takes a few more hours before those validators get deactivated. I would say it’s not worth complicating the code for right now.

cli/maintainer/src/main.rs
```diff
@@ -137,21 +138,21 @@ enum SubCommand {
...
-      V: Validation fee share (this is for all validators combined)
       D: Developer fee share
       A: stSOL value appreciation share
 
-    For example, if the reward distribution is set to '5 : 3 : 2 : 90', then 90%
-    of the rewards go to stSOL value appreciation, and 10% go to fees. Of those
-    fees, 50% go to the treasury, 30% are divided among validators, and 20% goes
-    to the developer.
+    For example, if the reward distribution is set to '5 : 3 : 92', then 92%
+    of the rewards go to stSOL value appreciation, and 8% go to fees. Of those
```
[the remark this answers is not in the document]
kkonevets: I think adding `(after validators take their commission)` is enough, so that Lido would have a hint to do this math.

tests/util.py
```diff
@@ -206,19 +208,23 @@ def create_stake_account(keypair_fname: str) -> TestAccount:
...
+    Generate a vote account for the validatora and authorized withdrawer account
     """
     test_account = create_test_account(vote_key_fname, fund=False)
+    withdrawer_account = create_test_account(authorized_withdrawer_key_fname, fund=True)
     solana(
         'create-vote-account',
         vote_key_fname,
         validator_key_fname,
-        authorized_withdrawer,
+        authorized_withdrawer_key_fname,
```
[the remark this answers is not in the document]
kkonevets: If I set `authorized_withdrawer_key_fname  = vote_key_fname ` I get
`Error: Bad parameter: Authorized withdrawer pubkey is identical to vote account pubkey, an unsafe configuration`
If I set `authorized_withdrawer_key_fname  = validator_key_fname` I get
`Error: Bad parameter: Authorized withdrawer pubkey is identical to identity account pubkey, an unsafe configuration`

program/src/processor.rs
kkonevets: Yes, `UpdateStakeAccountBalance ` sounds good. It distributes fees and it also withdraws inactive stake, but the main thing it does is updating a state variable `stake_accounts_balance`. Fixed

program/src/processor.rs
kkonevets: But still there could be a joker who transfers to stake account and it is considered as rewards by solido. So excess is a sum of rewards + joker transfer. When donation arrives to stake account it is inactive, but when rewards are payed of they become active immediately. That's the only difference. I will fix the comment

program/src/processor.rs
kkonevets: fixed

program/tests/tests/withdraw_inactive_stake.rs
kkonevets: done

program/tests/tests/withdraw_inactive_stake.rs
kkonevets: Yes, but since we can manipulate with small Lamports to make assert happy it does not matter. Secondly, that is your code initially. I will fix it anyway.

program/tests/tests/withdraw_inactive_stake.rs
kkonevets: done

program/tests/tests/withdraw_inactive_stake.rs
kkonevets: done

program/tests/tests/solana_assumptions.rs
kkonevets: Yes, but it looks ugly IMHO

```rust
    let balance_t0 = (context.get_sol_balance(vote_account).await
        + context.get_sol_balance(stake_account).await)
        .unwrap();
```

Anyway, I fixed it.

program/tests/tests/solana_assumptions.rs
kkonevets: done

program/src/state.rs
kkonevets: done

tests/test_solido.py
kkonevets: Yes, you mentioned it in `util.py` but I answer in there

kkonevets:
Thank you for your work, appreciate it. It's ain't easy for me too to answer all comments.
I fixed all I could, please check it out.

cli/maintainer/src/commands_solido.rs
```diff
@@ -935,14 +936,19 @@ pub fn command_deactivate_validator_if_commission_exceeds_max(
...
         violations.push(ValidatorViolationInfo {
             validator_vote_account: vote_pubkey,
             commission,
         });
     }
 
+    let signers: Vec<&dyn Signer> = vec![];
+    // Due to the fact that Solana has a limit on number of instructions in a transaction
+    // this can fall if there would be alot of misbehaved validators each
+    // exceeding `max_commission_percentage`. But it is very improbable scenario.
```
ruuda, line 949:
```suggestion
    // exceeding `max_commission_percentage`. But it is a very improbable scenario.
```

cli/maintainer/src/daemon.rs
```diff
@@ -104,7 +104,7 @@ impl MaintenanceMetrics {
                         .with_label("operation", "StakeDeposit".to_string()),
                     Metric::new(self.transactions_update_exchange_rate)
                         .with_label("operation", "UpdateExchangeRate".to_string()),
-                    Metric::new(self.transactions_withdraw_inactive_stake)
+                    Metric::new(self.transactions_update_stake_account_balance)
                         .with_label("operation", "WithdrawInactiveStake".to_string()),
```
ruuda, line 108: We should update the label as well, this is what gets exposed in the end.
```suggestion
                        .with_label("operation", "UpdateStakeAccountBalance".to_string()),
```

program/src/process_management.rs
```diff
@@ -55,7 +55,7 @@ pub fn process_add_validator(program_id: &Pubkey, accounts_raw: &[AccountInfo])
     // Deserialize also checks if the vote account is a valid Solido vote
     // account: The vote account should be owned by the vote program, the
     // withdraw authority should be set to the program_id, and it should
-    // sattisfy commission limit.
+    // sattisfy the commission limit.
```
ruuda, line 58:
```suggestion
    // satisfy the commission limit.
```
D'oh, another typo

ruuda (approved):
The latest round of changes looks good to me. I caught a few more typos.

tests/util.py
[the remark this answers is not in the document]
ruuda: Ah, it is no longer allowed, never mind then.

cli/maintainer/src/daemon.rs
kkonevets: sharp eye, fixed

ruuda:
Merged! The next step now would be the compatibility layer — a way for the new program to read the old state, and upgrade it in place. I also realized that `Lido::deserialize_lido` does not check the version yet. It should check that the version matches the expected version, and fail otherwise, because it will not be able to read the v0 state.

> We will not know Solido validator fee metrics (e.g. fee_validation_total_lamports), because a public validator can have stake accounts beyond Solido. So we can approximately compute it from amount of Solido stake

Ah yes, this would be interesting to add a metric for, the current fee percentages of all enrolled validators.

kkonevets:
Thanks. Currently I prepare a PR to integrate [BigVec](https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/src/big_vec.rs). That's a hole lot of changes. Only after that it makes sence to implement `UpdateState` instruction, because `BigVec` brings additional changes in data structures.
Thanks for the hint about `Lido::deserialize_lido`.

kkonevets:
`BigVec` does exactly that. It does not deserialize with Borsh, but [reinterprets](https://github.com/solana-labs/solana-program-library/blob/57c5fd80f60ed4d6053989e64882f9002e352910/stake-pool/program/src/big_vec.rs#L207) list entries with an unsafe pointer cast. More over it can cheaply select entries with a binary predicate. For example, we can select a validator with a specific pubkey. So that's just a number of `strcmp` calls, which is not that long, but indexing would be a lot faster for sure.

ruuda:
> So if the second maintainer removes a validator from the beginning of the list indexes will shift by 1 and the first maintainer will have no way to know that. Indexes are not "unique", pubkeys are much better

You would still verify the pubkey of course, but you can skip the linear search. If the pubkey at the user-provided index is not the expected one, simply fail the transaction.

PR #570, opened 2022-06-09
Update some dependencies

ruuda:
This should silence some of the dependabot reports, although most of them are transitive dependencies of Solana, which is harder to update. @kkonevets is working on that.


kkonevets:
Yeah, I was panning to do that, thanks.

ruuda:
Ugh, this breaks the tests, I think we need to update Solana as well: https://github.com/ChorusOne/solido/runs/6811251135?check_suite_focus=true#step:5:631

ruuda:
This has been superseded by #571.

PR #571, opened 2022-06-10
Switching to Solana v1.9.28

kkonevets:
- Updating solana to v1.9.28
- Updating build.yml to solana 1.9.28
- Patching Anchor 0.13 with solana 1.9.28
- Patching multisig with Anchor 0.13 to be able to build a listner
- calling `util::ensure_orca_program_exists()` once so that SOlana does  not block us by IP when running lots of tests
- refactoring activating, deactivating, effective in stake account according to solana 1.9.28
- using `get_latest_blockhash` instead of `RecentBlockhashes `
- fixing tests
- using `get_new_latest_blockhash` in a normal way
- updating licenses


cli/common/Cargo.toml
```diff
@@ -7,20 +7,20 @@ name = "solido-cli-common"
 version = "1.3.2"
 
 [dependencies]
-anchor-lang = "0.13.2"
+anchor-lang = "0.13.0"
```
ruuda, line 10: Why .0? Couldn’t we patch 0.13.2? Or is the fork based off of 0.13.2?

cli/common/src/snapshot.rs
```diff
@@ -260,16 +257,9 @@ impl<'a> Snapshot<'a> {
...
-        self.get_bincode(&sysvar::recent_blockhashes::id())
-    }
-
     /// Return the most recent block hash at the time of the snapshot.
-    pub fn get_recent_blockhash(&mut self) -> crate::Result<Hash> {
-        let blockhashes = self.get_recent_blockhashes()?;
-        // The blockhashes are ordered from most recent to least recent.
-        Ok(blockhashes[0].blockhash)
+    pub fn get_latest_blockhash(&mut self) -> crate::Result<Hash> {
+        Ok(self.rpc_client.get_latest_blockhash()?)
```
ruuda, line 262: The point of the snapshot is that you can consistently read many things from a single atomic snapshot of the chain state. That means that the snapshot is not allowed to do any RPC calls, except for fetching the snapshot itself. If we make a new RPC call, then the blockhash that call returns can be a different block hash than the one the snapshot was built from.

(I recall something about Solana wanting to remove this sysvar. I also recall something about it being available in the raw RPC response. But not sure if my memory is correct here.)

program/tests/tests/add_remove_validator.rs
```diff
@@ -108,7 +108,7 @@ async fn test_remove_validator_with_unclaimed_credits() {
     let vote_account = solido.validators.entries[0].pubkey;
     assert_eq!(
         solido.validators.entries[0].entry.fee_credit,
-        StLamports(62_301_530_769)
+        StLamports(998_758_598)
```
ruuda, line 111: Did you manage to find out what caused the difference? Is it a difference in the inflation schedule?

program/tests/tests/limits.rs
```diff
@@ -79,7 +80,11 @@ async fn test_max_validators_maintainers() {
         let maintainer = context.add_maintainer().await;
         context.maintainer = Some(maintainer);
 
-        let validator = context.add_validator().await;
+        validator = Some(context.add_validator().await);
+    }
+
+    // take out of the loop to reduce the time to wait
```
ruuda, line 86: But now this is testing a weaker property, it only tests deposit/stake/unstake with a single validator.

testlib/src/solido_context.rs
```diff
@@ -123,27 +125,11 @@ pub async fn send_transaction(
...
-            .get_new_blockhash(&context.last_blockhash)
-            .await
-            .expect("Failed to get a new blockhash.")
-            .0;
-    }
+    context.last_blockhash = context
+        .banks_client
+        .get_new_latest_blockhash(&context.last_blockhash)
+        .await
+        .expect("Failed to get a new blockhash.");
```
ruuda, line 132: :+1: Nice that we no longer need this hack.

testlib/src/solido_context.rs
```diff
@@ -589,7 +559,7 @@ impl Context {
         let rent = self.context.banks_client.get_rent().await.unwrap();
         let rent_voter = rent.minimum_balance(VoteState::size_of());
 
-        let initial_balance = Lamports(42);
+        let initial_balance = Lamports(rent.minimum_balance(0));
```
ruuda, line 562: Ah, so just entering the minimum rent-exempt balance did the trick, nice.

tests/check_licenses.py
```diff
@@ -37,6 +38,8 @@
     'ring',
     # All runtime code that we depend on is covered by an ISC-style license.
     'webpki',
+    # https://github.com/hsivonen/encoding_rs#licensing
```
ruuda, line 41: Please explain the actual situation in the comment, the link might die at some point in the future.

ruuda (commented):
Whew, thanks for this heroic effort! It looks almost good to me aside from one thing, the `get_latest_blockhash` in the snapshot. It should really produce the latest blockhash _from that snapshot_, it should not query the RPC again.

cli/common/Cargo.toml
kkonevets: Patching 0.13.0 and 0.13.2 is a twice more work than just patching 0.13.0

program/tests/tests/add_remove_validator.rs
kkonevets: No, I did't do a research yet. Just wanted do deliver ASAP

program/tests/tests/limits.rs
kkonevets: We add a new validator in the loop with `context.add_validator().await`, so plenty of validators. And at the end we just test that `deposit` and `unstake` work properly with a list of new validators.

cli/common/src/snapshot.rs
kkonevets: So maybe we should cache result of get_latest_blockhash() and reuse on new sign_transaction() calls?

cli/common/Cargo.toml
kkonevets: 0.13.0 is used in multisig, so I took a lower bound

tests/check_licenses.py
kkonevets: Ok

cli/common/src/snapshot.rs
kkonevets: Done https://github.com/lidofinance/solido/blob/b07f2b32475850d6e167fb1f64620d57e0e58964/cli/common/src/snapshot.rs#L277-L289

ruuda:
b07f2b3 does not fix the problem, it still makes an RPC call that is separate from the `GetMultipleAccounts` call. We *have* to get block hash from that `GetMultipleAccounts` call, either through a sysvar, or from elsewhere in the response.

We can get the slot number from `GetMultipleAccounts` (both from the sysvar and in the `context` part of the response). Maybe what we can do to get the block hash is to fetch the block at that slot? This also performs a separate RPC call, but because there is a unique block hash for every (finalized) slot, the output there is not a moving target, so that is fine to do in a separate request.

kkonevets:
Good point. Fixed. Please approve the CI build.

ruuda (approved):
LGTM!

ruuda:
[The integration test failed :confused:](https://github.com/ChorusOne/solido/runs/6860194327?check_suite_focus=true#step:9:50)

    Command '('target/debug/solido', '--cluster', 'http://127.0.0.1:8899/', '--output', 'json', 'multisig', 'create-multisig', '--multisig-program-id', 'AM83tkD6iukH4LtkCR1TbP9Xp1xUJh2kCRriYR4vKvaH', '--threshold', '2', '--owners', 'GjhPQBR1wA7H9efpbFeXsRJZVkroLsUNg8Wve2fheePG,7F1ET9qn9Sm2ULkxXSfaYa99X8w3B9ufpdKPdrNDCSUV')' returned non-zero exit status 1.

We’ve seen some flakes in the integration tests in the past but it would usually fail earlier, can you check what’s up?

kkonevets:
RPC should be done with commitment "confirmed", otherwise we get "block not available for slot ..." error

kkonevets:
Can we pull to main now?

ruuda (approved):

kkonevets:
By the way, should we increment a version number of solido to 1.3.3 ?

kkonevets:
Soon I will edit this PR https://github.com/ChorusOne/solido/pull/569. So we should probably wait

ruuda:
Actually, we should probably update the maintainer daemon to remove the Anker steps again and then make a new release, before the larger "2.0" update.

PR #572, opened 2022-06-27
Stop calling `Anker::SendRewards` in maintainer

ruuda:
The maintenance daemons are still trying to call `Anker::SendRewards`, and generating an error most of their duty time. It’s not really harmful, but it would be nice if they didn’t error so we can see when there is actually a problem.

We can make this a small 1.3.3 release just for the maintainer daemon, independent of the v2 stuff that @kkonevets is working on. Nobody is required to update, but I would like to update ours.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## v1.3.3
+
+Released 2022-07-08.
+
+The on-chain Solido program remains functionally unchanged since v1.0.0. The
+Anker program remains unchanged since v1.3.0.
+
+Changes:
+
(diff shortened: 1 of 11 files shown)


enriquefynn (approved):
LGTM

kkonevets:
O, I have just noticed you requested my review. Need some time to finish current task and then I will check it, thanx

kkonevets (approved):
LGTM

ruuda:
Tagged this change as v1.3.3, which is on top of v1.3.2 directly, and not reachable from `main`. For merging to `main`, there is a conflict in `CHANGELOG.md`, I will resolve it and then merge it, just beware that the resulting commits will be different from the one that is tagged.

PR #573, opened 2022-06-29
Integrate BigVec data structure

kkonevets:
The main goal of this PR is to increase maximum number of validators and maintainers in a pool and optimize overall performance.
- `AccountMap` was replaced by [BigVec](https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/src/big_vec.rs) which was rewritten to add `get_mut(index)` and `remove(index)` methods
- optimized some places in `processor.rs`
- added two new accounts `validator_list` and `maintainer_list` which hold the lists
- `validator_index` and `maintainer_index` are passed to instructions to precisely get/remove a value from a list, with credit to @ruuda 
- deletion from a list is done with `swap_remove` in O(1), with credit to @ruuda 
- `Validator`, `Maintainer` and `Lido` structures got `lido_version` field, which is checked 
- Account type is written on serialization and checked on deserialization
- current `max_validators` and `max_maintainers` is 6700, previously 64


ruuda:
It looks like you are introducing new accounts that hold state. This is opens up a big can of worms security-wise, because now you need to reason about a set of accounts, and be very _very_ careful every time to check that the accounts are related in the right way. Also, it becomes a harder to get a consistent view of the state from the RPC — now `GetMultipleAccounts` is the only safe option, but people will forget that and introduce torn reads that lead to subtle bugs.

If you read [the Neodyme report](https://github.com/ChorusOne/solido/blob/7d74b094c6a96486257e14d3c89e011c5d00fce6/audit/2021-08-03-neodyme.pdf), they confirm:

> Every program on Solana has to store data somehow. There are multiple approaches to this, and solido chose one of the safest ones: each instance of the solido program has a single unique account that stores all data. While solido allows for multiple instances to exist simultaneously, it is written in a way that rules out many typical Solana vulnerabilities by design. By storing all state in a single account that is subsequently required in every operation, all attacks that work by confusing between multiple program-owned accounts are by definition impossible.

I don’t think these additional accounts are needed; the lists can go into the main Solido account, you can still have the memory layout match the serialization format and transmute the input.

---

Also, it looks like much of the code in `state.rs` was copied verbatim from https://github.com/solana-labs/solana-program-library/blob/b5cec79ee4ae0dab014e7d5d9551b30f959245af/stake-pool/program/src/state.rs#L525. I am not a lawyer, but I am pretty sure that this is a violation of section 4a, 4b, and 4c of the Apache 2.0 license that covers the Solana Program Library. If you do want to use any third-party code at all, I recommend starting with a verbatim copy of the original source, in a separate file, don’t mix it with our existing code. Then add a very clear header that says who the original author and copyright holder is, and the license, and that you changed these files. And do the changes in a separate commit, so it is clear what the changes are versus the original.

kkonevets:
One drawback of keeping validator list inside Lido struct is that you increase stack (or heap if you choose) when deserializing Lido.

```rust
pub struct Lido {
    pub lido_version: u8,
    pub account_type: AccountType,
    pub manager: Pubkey,
    pub st_sol_mint: Pubkey,
    pub exchange_rate: ExchangeRate,
    pub sol_reserve_account_bump_seed: u8,
    pub stake_authority_bump_seed: u8,
    pub mint_authority_bump_seed: u8,
    pub reward_distribution: RewardDistribution,
    pub fee_recipients: FeeRecipients,
    pub metrics: Metrics,

    pub validator_list_data: [u8; VALIDATOR_LIST_SIZE],
    pub maintainer_list_data: [u8; MAINTAINER_LIST_SIZE],

    pub max_commission_percentage: u8,
}
```

So when we do this: 
```rust
let lido = try_from_slice_unchecked::<Lido>(&accounts.lido.data.borrow())?;
```
So `lido.validator_list_data` and `lido.maintainer_list_data` would be copied.

If we don't want to copy we should implement deserialization that does `borsh` for us by our self. This means every Lido field should be manually tracked in `accounts.lido.data`. That is cumbersome and inconvenient. 
Next, as we know Solana has a stack limit of 4Kb and 32kb of heap size. And validator list plus maintainer list size is about 0.8Mb. So it does not fit in to stack nor in to heap.
So keeping list data in a separate account is not a bad idea.

ruuda:
Re: original message, what I would do is give the `Lido` struct a field `validators: &'a [(Pubkey, Validator)]`. That will give `Lido` itself a lifetime parameter though, which may be a very invasive change. Then in the deserializer that takes the `raw_account_data: &[u8]`, deserialize the known-size part of the `Lido` struct from the first _n_ bytes, then transmute the remainder of the raw data into a `&[(Pubkey, Validator)]` and complete the struct.

If that is too invasive, or if it’s not directly possible to transmute the raw bytes into `(Pubkey, Validator)`, an alternative is to only put the size of the list in the `Lido` struct, and then in every method that needs access to the validator list, pass in the `&[u8]` separately, and deserialize the single validator at that offset. That approach also doesn’t need an additional lifetime parameter, but I think it’s a bit more error-prone. Also, some operations require walking the validator list, so if some deserialization is still required per validator that will be expensive; I think a transmute would be the best option.

kkonevets:
@ruuda Now I place maintainers and validators next to Lido structure in same account. Please take a look ASAP

ruuda (commented):
I reviewed `big_vec.rs`; I haven’t looked at the other files yet.

program/src/big_vec.rs
```diff
@@ -0,0 +1,490 @@
...
+
+    /// Get element at position
+    pub fn get_mut<T: Pack>(&mut self, index: u32) -> Result<&mut T, ProgramError> {
+        let (start_index, end_index) = self.get_slice_bounds::<T>(index)?;
+        let ptr = self.data[start_index..end_index].as_ptr();
+        Ok(unsafe { &mut *(ptr as *mut T) })
+    }
+
+    /// Removes and returns the element at position index within the vector, shifting all elements after it to the left.
+    pub fn remove<T: Pack + Clone>(&mut self, index: u32) -> Result<T, ProgramError> {
```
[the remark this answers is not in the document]
kkonevets: We don't need it now. But in future we could

program/src/big_vec.rs
```diff
@@ -0,0 +1,490 @@
...
+pub struct Iter<'data, 'vec, T> {
+    len: usize,
+    current: usize,
+    current_index: usize,
+    inner: &'vec BigVec<'data>,
+    phantom: PhantomData<T>,
+}
+
+impl<'data, 'vec, T: Pack + 'data> Iterator for Iter<'data, 'vec, T> {
+    type Item = &'data T;
```
ruuda, line 235: I think the item type will need to be `T`, not `&T`, to be able to back the iterator by [`read_unaligned`](https://doc.rust-lang.org/std/ptr/fn.read_unaligned.html).

program/src/big_vec.rs
```diff
@@ -0,0 +1,490 @@
...
+    },
+    std::marker::PhantomData,
+};
+
+/// Contains easy to use utilities for a big vector of Borsh-compatible types,
+/// to avoid managing the entire struct on-chain and blow through stack limits.
+#[derive(Debug)]
+pub struct BigVec<'data> {
+    /// Underlying data buffer, pieces of which are serialized
+    pub data: &'data mut [u8],
```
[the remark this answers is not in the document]
kkonevets: A vector capacity is defined [here](https://github.com/lidofinance/solido/blob/1b8647a9b5637f4c126807a570327adac24fd8de/program/src/state.rs#L98)
`max_entries` is placed at the header of a vector, this declares ahead possible number of elements. When deserializing Solido first reads `max_entries` and so it knows whre is the beginning of a next vector. Vector deserialization is done [here](https://github.com/lidofinance/solido/blob/1b8647a9b5637f4c126807a570327adac24fd8de/program/src/state.rs#L1074) 
So, no we do not use the entire remainder of the account data for the validators vec.

program/src/big_vec.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        // if not last element replace it with last
+        if index != self.len() - 1 {
+            unsafe {
+                sol_memmove(
+                    self.data[start_index..end_index].as_mut_ptr(),
+                    self.data[data_end_index - T::LEN..data_end_index].as_mut_ptr(),
+                    T::LEN,
+                );
+            }
```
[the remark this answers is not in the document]
kkonevets: John states [here](https://github.com/solana-labs/solana-program-library/blob/024ba3ad410fef2d31e500c0f1a30db0c222a6a8/stake-pool/program/src/big_vec.rs#L48) that it's not that cheap. If we will ever remove a bunch of values from a vector it will matter. I tried the safe variant, it works well. But I left unsafe just for speed reason. Our workflow is fine for this function and performance will not suffer for now. So I can replace it with safe code.

ruuda (commented):
Reviewed `instruction.rs` and `process_management.rs` now.

program/src/instruction.rs
```diff
@@ -132,6 +123,82 @@ pub enum LidoInstruction {
...
+        validator_index: u32,
+        // Index of a maintainer in maintainer list
+        #[allow(dead_code)] // but it's not
+        maintainer_index: u32,
+    },
+
+    /// Update the exchange rate, at the beginning of the epoch.
+    ///
+    /// This can be called by anybody.
+    UpdateExchangeRateV2,
```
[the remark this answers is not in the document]
kkonevets: Yes, have not noticed, thanks. First, there were `validator_list` and `maintainer_list` which I removed and forgot to remove V2 postfix.

program/src/big_vec.rs
```diff
@@ -0,0 +1,490 @@
...
+        let mut vec_len = u32::try_from_slice(vec_len_ref)?;
+
+        let start_index = VEC_SIZE_BYTES + vec_len as usize * T::LEN;
+        let end_index = start_index + T::LEN;
+
+        vec_len += 1;
+        vec_len.serialize(&mut vec_len_ref)?;
+
+        if self.data.len() < end_index {
+            return Err(ProgramError::AccountDataTooSmall);
```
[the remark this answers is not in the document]
kkonevets: Nice. Actually, that's John's code, but I make same mistake later. Fixed.

program/src/big_vec.rs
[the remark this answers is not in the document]
ruuda: Then let’s delete it, we can add it when we need it.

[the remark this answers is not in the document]
kkonevets: done, will commit soon

program/src/big_vec.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        let new_len = self.len() - 1;
+        let mut vec_len_ref = &mut self.data[0..VEC_SIZE_BYTES];
+        new_len.serialize(&mut vec_len_ref)?;
+
+        Ok(value)
+    }
+
+    /// Extracts a slice of the data types
+    pub fn deserialize_mut_slice<T: Pack>(
```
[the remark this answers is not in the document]
kkonevets: Deleted this function

program/src/state.rs
```diff
@@ -3,58 +3,547 @@
...
+    ) -> Result<&'data mut T, ProgramError> {
+        let element = self.big_vec.get_mut::<T>(index)?;
+        Self::check_pubkey(element, pubkey)?;
+        Ok(element)
+    }
+
+    /// Removes element with pubkey at index
+    pub fn remove(&'data mut self, index: u32, pubkey: &Pubkey) -> Result<T, ProgramError> {
+        let element = self.big_vec.swap_remove::<T>(index)?;
+        Self::check_pubkey(&element, pubkey)?;
```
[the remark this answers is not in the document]
kkonevets: fixed

program/src/processor.rs
```diff
@@ -125,16 +131,18 @@ pub fn process_initialize(
...
 
     // Confirm that the fee recipients are actually stSOL accounts.
     lido.check_is_st_sol_account(accounts.treasury_account)?;
     lido.check_is_st_sol_account(accounts.developer_account)?;
 
-    lido.save(accounts.lido)
+    // first save Lido, then maintainer list and then validator list
+    let lido_data = &mut *accounts.lido.data.borrow_mut();
+    lido.save(lido_data)?;
+    maintainers.save(&mut lido_data[LIDO_CONSTANT_SIZE..])?;
```
ruuda, line 144: What does it need to save here? The `ValidatorList` and `MaintainerList` mutate the backing account directly, so there is nothing to save, right?

program/src/processor.rs
```diff
@@ -183,71 +191,69 @@ pub fn process_deposit(
...
-    let validator = lido.validators.get(accounts.validator_vote_account.key)?;
-
-    if !validator.entry.active {
-        msg!(
-            "Validator {} is inactive, new deposits are not allowed",
-            validator.pubkey
-        );
-        return Err(LidoError::StakeToInactiveValidator.into());
-    }
+    let mut validators = Lido::deserialize_account_list::<Validator>(lido_data)?;
```
ruuda, line 214: I think I mentioned it before, but I think it would be nice to package this together with `Lido::deserialize_lido`, so it would return a tuple of the Lido instance, and also the validator list and maintainer list. Then we know for sure that they are consistent with one another.

program/src/processor.rs
```diff
@@ -300,7 +306,8 @@ pub fn process_stake_deposit(
     // record that here; we will discover it later in `WithdrawInactiveStake`,
     // and then it will be treated as a donation.
     msg!("Staked {} out of the reserve.", amount);
-    validator.entry.stake_accounts_balance = (validator.entry.stake_accounts_balance + amount)?;
+    validator.stake_accounts_balance = (validator.stake_accounts_balance + amount)?;
+    validator.effective_stake_balance = validator.get_effective_stake_balance();
```
ruuda, line 310: Hmm, if there is a field `effective_state_balance` and also a method to get it, they can be confused easily, and there may be a risk of the field going out of sync with the getter. Maybe we can make the method mutate the validator instead? Something like `validator.recompute_effective_stake_balance()`?

program/src/processor.rs
```diff
@@ -505,13 +512,14 @@ pub fn process_unstake(
...
             return Err(LidoError::InvalidAmount.into());
         }
-        validator.entry.stake_seeds.begin += 1;
+        validator.stake_seeds.begin += 1;
     }
 
-    validator.entry.unstake_accounts_balance = (validator.entry.unstake_accounts_balance + amount)?;
-    validator.entry.unstake_seeds.end += 1;
+    validator.unstake_accounts_balance = (validator.unstake_accounts_balance + amount)?;
+    validator.effective_stake_balance = validator.get_effective_stake_balance();
```
ruuda, line 519: Also here, it seems safer if it would be a method `recompute_effective_stake_balance()`, also so you cannot assign it to the wrong field or forget to assign it.

program/src/processor.rs
```diff
@@ -977,36 +983,77 @@ pub fn process(program_id: &Pubkey, accounts: &[AccountInfo], input: &[u8]) -> P
...
+        | LidoInstruction::AddValidator
+        | LidoInstruction::AddMaintainer
+        | LidoInstruction::DeactivateValidator
+        | LidoInstruction::RemoveMaintainer
+        | LidoInstruction::RemoveValidator
+        | LidoInstruction::StakeDeposit { .. }
+        | LidoInstruction::Unstake { .. }
+        | LidoInstruction::Withdraw { .. } => {
+            msg!("{:?} is no longer supported since v2. Please check the changelog in the repository for update instructions.", instruction);
+            Err(LidoError::InstructionIsDeprecated.into())
```
ruuda, line 1056: Can you add the newly deprecated functions to the changelog as well? The main thing to explain is that many of these functions now take the validator and maintainer index as transaction data, to make these instructions more efficient, so clients need to compute that index, similar to pre-computing a bump seed so it only has to be verified on-chain.

ruuda (commented):
Reviewed `processor.rs`, the changes there are pretty straightforward, that looks good, I only have minor comments. Next up is `state.rs`.

program/src/big_vec.rs
```diff
@@ -0,0 +1,490 @@
...
+    solana_program::{
+        program_error::ProgramError, program_memory::sol_memmove, program_pack::Pack,
+    },
+    std::marker::PhantomData,
+};
+
+/// Contains easy to use utilities for a big vector of Borsh-compatible types,
+/// to avoid managing the entire struct on-chain and blow through stack limits.
+#[derive(Debug)]
+pub struct BigVec<'data> {
```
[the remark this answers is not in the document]
kkonevets: Agree, `BigVec<Foo>` that is much better. Fixed

program/src/big_vec.rs
kkonevets: That means you make a copy on each iteration step. This is bad. Remember that our protocol works on-chain and if some clients will not be able to run cli on M1 it is totally fine. We can't sacrifice for convenience in protocol

kkonevets: One thing only matters - that unaligned code works on-chain.

program/src/big_vec.rs
```diff
@@ -0,0 +1,490 @@
...
+            len: self.len() as usize,
+            current: 0,
+            current_index: VEC_SIZE_BYTES,
+            inner: self,
+            phantom: PhantomData,
+        }
+    }
+
+    /// Find matching data in the array
+    pub fn find<T: Pack>(&self, data: &[u8], predicate: fn(&[u8], &[u8]) -> bool) -> Option<&T> {
```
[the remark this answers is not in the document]
kkonevets: Not quite, `Iterator::find` transmutes every value to `T` and compares to pubkey. Find with `predicate` will just memcmp first 32 bytes of a string with pubkey and transmute once on return. We eliminate almost all transmuting this way. Don't know if this gives a lot of performance boost, should benchmark it, but it will give some definitely.

UPDATE:
I benched it. I was able to handle only 4000 of validators vs 6700 previously. So we should leave it as is.

program/src/big_vec.rs
kkonevets: Another workaround is this. We can use unaligned code in protocol, but use aligned code on a client side. We can **implement is later** without updating protocol. But protocol should be with unaligned code due to performance issues.

program/src/big_vec.rs
```diff
@@ -0,0 +1,490 @@
...
+        }
+
+        Ok((start_index, end_index))
+    }
+
+    /// Get element at position
+    pub fn get_mut<T: Pack>(&mut self, index: u32) -> Result<&mut T, ProgramError> {
+        let (start_index, end_index) = self.get_slice_bounds::<T>(index)?;
+        let ptr = self.data[start_index..end_index].as_ptr();
+        Ok(unsafe { &mut *(ptr as *mut T) })
```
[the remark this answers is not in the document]
kkonevets: It is safe to cast unaligned pointer in bpf. I already answered question about nonalignment somewhere here. But in essence we can leave unaligned cast in protocol and implement aligned code on client, so those hipsters with macs be happy. Even if we would not implement a client can always find a PC with Linux. 
We can implement a client side code later without updating protocol. But protocol should be with unaligned code due to performance issues I mentioned in another comment.

program/src/state.rs
```diff
@@ -3,58 +3,546 @@
...
+
+    fn new(pubkey: Pubkey) -> Self;
+    fn pubkey(&self) -> Pubkey;
+
+    /// Performs a very cheap comparison, for checking if entry
+    /// info matches the account address.
+    /// First PUBKEY_BYTES of a ListEntry data should be the account address
+    fn memcmp_pubkey(data: &[u8], pubkey: &[u8]) -> bool {
+        sol_memcmp(&data[..PUBKEY_BYTES], pubkey, PUBKEY_BYTES) == 0
+    }
```
ruuda, line 117: This looks like something that will be very error-prone to use. What if you pass in the wrong data or wrong pubkey array? I would on the existing `Eq` impl for `Pubkey`. There are no cases left where we need to do a linear scan and compare the pubkey of every entry, now that we provide indexes in the instructions, right? So performance of a single `Eq` of 32 bytes per instruction call, that should be negligible compared to the other authorization checks we already do.

ruuda (commented):
I looked at a part of `state.rs`, will continue tomorrow.

program/src/big_vec.rs
kkonevets: It could be or it could not. But it brings complexity anyway. I suggest to not bother and postpone aligned client code implementation to a next PR. Anyway, this code is in rust tradition of doing things. We implement small pieces of unsafe code and then use them. All rust  stdis like that.

program/src/process_management.rs
```diff
@@ -38,12 +38,12 @@ pub fn process_change_reward_distribution(
     lido.fee_recipients.treasury_account = *accounts.treasury_account.key;
     lido.fee_recipients.developer_account = *accounts.developer_account.key;
 
-    lido.save(accounts.lido)
```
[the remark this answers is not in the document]
kkonevets: The main point is that you need to have data reference `&mut [u8]` on stack to be able to use BigVec. So the function can't return `ValidatorList<'a>` because a temporary data reference `&mut [u8]` will be destroyed. I could think about a better interface but now it looks convenient to use in instructions.

program/src/processor.rs
kkonevets: Yes. But we have to serialize empty `AccountList`'s once on initialization. From now on we use BigVec only. We use `borsh` here so that we do not have to implement manually BigVec initial serialization: serializing `ListHeader` struct and putting number of elements in `BigVec` ahead. So `borsh` helps us to init. That's John's idea, pretty good

program/src/processor.rs
kkonevets: O, now I get what you meant. The problem is that you can't mutably access `accounts.lido.data` simultaneously by maintainers and validators, the almighty borrow checker will not be happy. So we can't return them both from `deserialize_account_list`. So the only option is to access `accounts.lido.data` sequentially.

program/src/processor.rs
kkonevets: Same here, answered earlier. To be considered...

kkonevets: Maybe I should rename `get_effective_stake_balance` to `recompute_effective_stake_balance` that does not mutate anything?

```rust
    pub fn recompute_effective_stake_balance(&self) -> Lamports {
        (self.stake_accounts_balance - self.unstake_accounts_balance)
            .expect("Unstake balance cannot exceed the validator's total stake balance.")
    }
```

program/src/processor.rs
kkonevets: done

program/src/state.rs
```diff
@@ -3,58 +3,546 @@
...
-pub type Validators = AccountMap<Validator>;
+/// Types of list entries
+/// Uninitialized should always be a first enum field as it catches empty list data errors
+#[derive(Clone, Debug, PartialEq, Eq, BorshDeserialize, BorshSerialize, Serialize, BorshSchema)]
+pub enum ListEntryType {
+    /// If the account has not been initialized, the enum will be 0
+    Uninitialized,
+    Validator,
+    Maintainer,
+}
```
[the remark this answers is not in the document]
kkonevets: To know a type of serialized list elements. Otherwise when deserializing there is now way to know a type of list elements. Although we could use a knowledge that first goes Lido, then maintainer list and then validator list in raw data `[Lido, validators, maintainers]` but it is error prone due to forgetting that and swapping the order. So the list type is always checked.
Also to produce generic messages: ` msg!("Pubkey {} is duplicated in a {:?} list", pubkey, T::TYPE)`
Also to choose an algorithm as in `fn deserialize_account_list<T: ListEntry>`

program/src/state.rs
```diff
@@ -3,58 +3,546 @@
...
+#[derive(
+    Clone, Debug, Default, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema, Serialize,
+)]
+pub struct AccountList<T> {
+    /// Data outside of the list, separated out for cheaper deserializations
+    #[serde(skip_serializing)]
+    pub header: ListHeader<T>,
+
+    /// List of accounts in the pool
+    pub entries: Vec<T>,
```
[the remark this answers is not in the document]
kkonevets: `pub entries: Vec<T>` does not hold a vector of pointers, `T` is a `Maintainer` or a `Validator`, which are regular structs. `AccountList<T>` is used in `Initialize` instruction once and then in CLI and tests.

program/src/state.rs
```diff
@@ -3,58 +3,546 @@
...
+/// Helper type to deserialize just the start of AccountList
+#[repr(C)]
+#[derive(
+    Clone, Debug, Default, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema, Serialize,
+)]
+pub struct ListHeader<T> {
+    /// Maximum allowable number of elements
+    pub max_entries: u32,
+
+    pub entry_type: ListEntryType,
```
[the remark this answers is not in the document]
kkonevets: Type parameter does not know nothing about serialization format. When you have raw bytes how would type parameter verify that the thing in raw bytes is actually a Maintainer or a Validator? The `ListHeader` is serialized on part with `BigVec` and is used too keep list type and a size of the following list. `ListHeader` is an info written ahead of the list.

program/src/state.rs
kkonevets: `BigVec` has `find` and `find_mut` that use `sol_memcmp` to efficiently search an element with given pubkey. It works much faster than using iterator with transmuting (already discussed it up the list). 
We can't do nothing if someone passes wrong data. Even Solana gives us access to raw data through `accounts.lido.data`. If someone misuses the data it's his fault. Actually `memcmp_pubkey` is used only by BigVec internally and is not is not a public function `fn memcmp_pubkey`. So it is OK.

ruuda: Aah, it’s for finding duplicates.

But still, do we need it? Do we run into the compute budget otherwise? This level of unsafety feels like a high price to pay for a sanity check like preventing duplicates. (I’m not saying that this particular implementation is unsafe, but there are only so many things you can do that need extremely careful review, until somebody misses something somewhere. The only way to confidently keep things secure, is to fight complexity, and try very very hard to keep things as simple as possible.)

program/src/state.rs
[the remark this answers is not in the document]
ruuda: > When you have raw bytes how would type parameter verify that the thing in raw bytes is actually a Maintainer or a Validator?

You know it from its position, it’s defined out of band by the enclosing type.

Similar to when you Borsh-serialize a `struct Foo { a: u64, b: u64 }`, how would you tell from the raw bytes that it’s two 64-bit integers? The format is not self-descriptive, but if you know that the byte array contains a Borsh-serialized `Foo`, then you know what the bytes mean.

program/src/state.rs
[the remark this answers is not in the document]
ruuda: Hmm, can we remove it then? During initialization the vec should be empty, so we can just write the zero len manually.

program/src/state.rs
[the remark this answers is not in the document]
ruuda: > but it is error prone due to forgetting that and swapping the order.

I would argue it’s the other way around, by introducing this, you introduce new representable states that are invalid. A big part of leveraging type systems to eliminate bugs is to make invalid states unrepresentable. You *still* depend on the order with the tag, but now you also need to verify that the tag matches the expectation, and add error handling logic for when it does not.

program/src/processor.rs
ruuda: Hmm, so are there cases where the field and the computed value diverge? Or is the field only an optimization?

program/src/processor.rs
ruuda: > The problem is that you can't mutably access accounts.lido.data simultaneously by maintainers and validators, the almighty borrow checker will not be happy

You can, with [`slice::split_at_mut`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at_mut).

kkonevets:
Sure bpf does not have alignment requirements, it's like x86_64 allows unaligned loads. That's why my code works. Also your ams snippet does not contain `ALIGN` directive.
> So there is no overhead to using unaligned_read for BPF targets, at least not in a setup like this, where we immediately access the fields

Immediately accessing a filed does the trick, agree. I have a main performance bottleneck - looping through validators and summing effective_stake_balance. Seems like it will work this way. But we should bench it.

In other scenarios `read_unaligned` will make a copy.  Docs say `read_unaligned creates a bitwise copy of T, regardless of whether T is Copy`
I will maybe check it out, but will postpone to the very end.

program/src/processor.rs
kkonevets: I have done it multiple times with all tests passing in different variants, but my aesthetic sense tells me it looks ugly. Instead one line I should add 3 new:

```rust
-    let mut lido = Lido::deserialize_lido(program_id, accounts.lido)?;

+    let lido_data = &mut *accounts.lido.data.borrow_mut();
+    let mut slices = LidoSlices::new(lido_data)?;
+    let (mut lido, _, _) = Lido::deserialize_lido(program_id, accounts.lido, &mut slices)?;
```

where
```rust
pub struct LidoSlices<'a> {
    pub lido: &'a mut [u8],
    maintainers: &'a mut [u8],
    validators: &'a mut [u8],
}
```
So I quit it. Don't want to spend more of my time on syntactic sugar, although  they could be inconsistent as you mention here. But I don't think it's a problem at all.

program/src/processor.rs
kkonevets: There are no such places in protocol. But I guess it is in `balance.rs`.

program/src/processor.rs
kkonevets: No, reverting it back. It is ugly.

program/src/state.rs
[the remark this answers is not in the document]
kkonevets: I would not advice to remove. If we do than BigVec will leak out to CLI and tests. It will be so inconvenient to stick to data on stack for BigVec: `struct BigVec<'data, T>`. It is much more convenient to just copy data to a regular Vec. I know that I can iter BigVec and collect it to Vec. But it means I should also reimplement all methods `AccountList` has for BigVec. Because when you will copy BigVec to Vec it should have methods `AccountList` had. So, no way

program/src/state.rs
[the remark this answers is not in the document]
kkonevets: My approach is similar to protopuf.

program/src/process_management.rs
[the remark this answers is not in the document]
kkonevets: `split_at_mut` will create temp variables on stack that will be destroyed after `Lido::deserialize_lido` returns. So you can't call `lido_account.data.borrow_mut` in that function. You should split the data before calling it. And it looks ugly, I tried.
Two make everything beautiful I suppose we have to use BigVec with `data: Rc<RefCell<&'a mut u64>>`. But it is overdo.

program/src/state.rs
kkonevets: I agree, partially. The compute budget is reduced by half of it's current value.
Rust all over the std uses unsafe code, so not use rust? A little bit of tested unsafe code will not harm. By the way `fund` uses a single line of unsafe code at return `Some(unsafe { &*(current_slice.as_ptr() as *const T) })`. We know bpf is OK with unalignment reads. It's not a problem.
A bigger problem is to over complicate code.
Marinade [also](https://github.com/marinade-finance/liquid-staking-program/blob/a309057f1eb3413070846d34e8fd5d83e99dc1c6/programs/marinade-finance/src/state.rs#L83-L87) uses unsafe but not much. 
We have just 4 unsafe declarations which are all in a single file.
Even if we use `read_unaligned` it will still be unsafe by definition. By the way I used `read_unaligned` in `swap_remove` [here](https://github.com/lidofinance/solido/blob/387d93d99ed23c0f376444559f2ddead3c78ed7a/program/src/big_vec.rs#L78)

program/src/state.rs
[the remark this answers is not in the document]
kkonevets: Here is a real example. I switched of the type check. It should panic on the last line, but it does not.
```rust
// add one validator with default value
let accounts = ValidatorList::new_default(1);

// allocate space for future elements
let mut buffer: Vec<u8> = vec![0; ValidatorList::required_bytes(accounts.header.max_entries)];
let mut slice = &mut buffer[..];
// serialize empty list to buffer, which serializes a header and length
BorshSerialize::serialize(&accounts, &mut slice).unwrap();

// deserialize to BigVec but with a different account type
let slice = &mut buffer[..];
ListHeader::<Maintainer>::deserialize_vec(slice).unwrap();
```

kkonevets:
About validator list account. In instruction that updates Lido state I need to increase lido.data space with reallocation, but I could not find an instruction that reallocates account space. Is it possible?
Looks like [allocate](https://github.com/solana-labs/solana/blob/bf225bae738c071bd02ec79ec0b8ddfc1679a766/runtime/src/system_instruction_processor.rs#L90) does not allow that

kkonevets:
Update on Lido account as a PDA account.
Here is one pitfall. PDA accounts have limit on max data size and it is less than 1Mb (can’t remember exact value). About a month ago I have experimented with validator_list account as a PDA account and found the limit. Also I read about the limit somewhere on the internet. The point is that validator and maintainer lists won’t feet the limit, I have tried, I am 100% sure.
So the only possibility I see now is to use a regular account. Hence it should be the current Lido account. As we can't resize it we should withdraw all lamports from it, wait for epoch boundary and create it again with new data size. So we should save the state somewhere and int the new account with it after epoch boundary. Sound very complicated.
Or we can create two new accounts: validator_list account and list account account. This is what I have done initially on the pull request.
I this we should go with two new accounts even if it less secure. Do we have a choice?

ruuda:
> PDA accounts have limit on max data size and it is much less than 1Mb

Hmm, that’s unfortunate :confused: 

> Create a new regular Lido account and this will break all client integrations.

We can still put all the state in a new (non-PDA) account and keep the old state at the old location. Whether it’s a PDA or not does not make such a big difference. A PDA is nice because then we can enforce there is only a single canonical instance, but for the current version the state is also a regular account. Of course old clients will not know where to find the new account, but even if we kept it at the same address they wouldn’t know how to parse the v2 data, so clients need to update either way. ~~With a new account, at least the data layout will not change under their feet.~~*

For doing the migration, we can put it in the migrate instruction. If would take both the old and new address, and it must be signed by both the manager of the instance and the new address.

\* I now realize, we shouldn’t allow the instance to be migrated more than once (we could trust the manager to not do that, but it would be better to make it impossible). The best way to achieve that, I think, is to clear out the original account after completing the migration. So yes, then it does break existing clients, but still, an update would do that any way.

ruuda:
> By the way, lesson learned - when creating a state account create it with maximum data size

This is somewhat expensive though, about 7 SOL for 1 MB. But yeah, it sounds like it would be worth it. (And it’s not as bad now that the price crashed :grimacing:)

ruuda:
> If we save in the state an old Solido address and derive from that it breaks all security, I guess

D'oh, yes, you are right. Hmm :frowning_face:

> So we should recreate all PDA accounts. Do you think that will work?

Hmm, yes ... I think it would!

> Also creating a new reserve account means we should transfer funds from the old. Sounds complicated.

Hey, now I have a new idea ... I think this might be quite nice actually, because we don’t need to care about backwards compatibility _at all_ in the new version! Maybe this even simplifies things!

* Create v2 without regard for backwards compatibility (though I would keep the version byte, just to avoid ambiguity).
* Deploy the program to a new address, keep the v1 program at the old address.
* The v2 program would have a new PDA for its reserve, and also for the mint authority.
* Build a v1.4 on top of v1, with only one new additional instruction: `Migrate`. Migrate would transfer everything in the v1 reserve to the v2 reserve, and change the mint authority of the mint from the v1 authority to the v2 authority.
* After the migration happens, v1 can no longer mint or burn stSOL, so deposts and withdraws on it would stop working. It should enforce then, that `Migrate` can only be called after there are no active validators any more.
* All of the complexity of the migration then stays in this v1.4 program, which is basically a single-use program.
* This makes the v2 program simpler, so easier to audit and reason about, and it also makes it a bit easier to review the migration in isolation.

What do you think?

kkonevets:
I think it is too complicated. We discussed it internally and decided to reset everything back to maintainer and validator list accounts. This is a simple safe change and it will work for sure.

kkonevets:
@ruuda I returned back validator and maintainer list accounts. Please check only correctness and security, not code style.

PR #579, opened 2022-08-09
Expose deactivating stake

### LINKED ISSUE #578: Deactivating stake is not exposed properly in Prometheus metrics
2022-08-08-solido-deactivating-stake

The stake should have converted from “active” into “deactivating”, but instead it disappeared from the metrics only to re-appear once it made its way back into the reserve after the epoch boundary.

enriquefynn:
Closes #578
Sum inactive stake on stake and unstake accounts.
Use deactivating stake on `unstake_accounts`.

Please confirm that these are the combinations allowed here

### DIFF
--- a/cli/maintainer/src/maintenance.rs
+++ b/cli/maintainer/src/maintenance.rs
@@ -1243,12 +1243,7 @@ impl SolidoState {
 
             let stake_balance: StakeBalance = stake_accounts
                 .iter()
-                .map(|(_addr, stake_account)| stake_account.balance)
-                .sum();
-
-            // Stake here can only be inactive or deactivating.
-            let unstake_balance: StakeBalance = unstake_accounts
-                .iter()
+                .chain(unstake_accounts.iter())
                 .map(|(_addr, stake_account)| stake_account.balance)
                 .sum();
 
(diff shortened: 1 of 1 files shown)


cli/maintainer/src/maintenance.rs
```diff
@@ -1239,15 +1246,25 @@ impl SolidoState {
                 .map(|(_addr, stake_account)| stake_account.balance)
                 .sum();
 
+            // Stake here can only be inactive or deactivating.
+            let unstake_balance: StakeBalance = unstake_accounts
+                .iter()
+                .map(|(_addr, stake_account)| stake_account.balance)
+                .sum();
```
ruuda, line 1253: I believe this works, but simpler would be to add the `unstake_balance` to the `stake_balance` and keep the metrics as-is. It’s already separated out into activating/active/deactivating/inactive, which account we found it in should not matter.

The diff could be as simple as

```diff
            let stake_balance: StakeBalance = stake_accounts
                .iter()
+               .chain(&unstake_accounts)
                .map(|(_addr, stake_account)| stake_account.balance)
                .sum();
```

ruuda (approved):
LGTM

PR #580, opened 2022-08-11
Improve validator name sanitizer

ruuda:
One of the validators in the last onboarding wave has U+26A1 HIGH VOLTAGE SIGN and U+FE0F VARIATION SELECTOR-16 in its name. In my terminal and also in Grafana in my browser, these render as a lightning bolt using an emoji font.

Also, due to the way we render labels when rendering the metrics, the variation selector turns into the literal text "\u{fe0f}" (so a
backslash and hex between curly brackets, not an U+FE0F itself).

Strip code points from both blocks to fix this.

### DIFF
--- a/cli/maintainer/src/maintenance.rs
+++ b/cli/maintainer/src/maintenance.rs
@@ -420,6 +420,8 @@ fn sanitize_validator_name(name: &str) -> String {
     // adds no information, strip it here to leave more space for graphs in
     // dashboards, and not waste so much space on the redundant part of the name.
     match name.strip_prefix("Lido / ") {
+        // Negated range.contains syntax only makes things more cryptic below.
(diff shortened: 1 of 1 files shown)


enriquefynn:
Isn't it better to have the names in ASCII? There's probably some built-in function to do that

enriquefynn:
> This is a very hard problem and depends on what you want to do and assumptions about the source language and script too ... (E.g. do you convert ü to ue, ß to ss? Or ü to just u? Or do you drop it entirely? What about code points that have no obvious ascii alternative?) But there is https://lib.rs/crates/deunicode which does an impressive job nonetheless.

That's fair, probably is better to do what's in this PR to depend on this crate 👍

enriquefynn (approved):

PR #581, opened 2022-08-12
Add changelog for 1.3.4

ruuda:
Let’s make a new release with the few small bugfixes for the maintainer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Changelog
 
+## v1.3.4
+
+Released 2022-09-27.
+
+The on-chain Solido program remains functionally unchanged since v1.0.0. The
+Anker program remains unchanged since v1.3.0.
+
+Changes:
+
+ * Update Solana SDK to 1.9.28.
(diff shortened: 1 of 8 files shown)


ruuda:
@enriquefynn, I now also included @kkonevets changes to handle the error that the maintainer is running into, let’s make a new release and deploy this today, can you take a look?

enriquefynn (approved):
LGTM, but we need to do another version to include changes from #583 or we could try to include them here

ruuda:
> we need to do another version to include changes from https://github.com/ChorusOne/solido/pull/583

Ah yes, good point. I will cherry-pick it onto `current` and include it in the changelog.

kkonevets:
@ruuda [this commit](https://github.com/ChorusOne/solido/pull/583) is not valid, don't cherry pick it please

ruuda:
Okay, I reverted back to `e83ce9bf`, the deactivation is not going to happen in the very near future either way, let’s get the current fixes deployed before the end of the epoch.

PR #582, opened 2022-08-17
Pass empty seeds when withdrawing

enriquefynn:
`process_withdraw` was failing on recent tests, we passed a slice of one element before `&[&[]]` to `invoke_signed`, probably Solido was trying to call a signed instruction with the empty slice as a bump seed, in fact we should have passed an empty slice `&[]` so it just ignores it. `invoke` calls `inkove_signed` with an empty slice.

### DIFF
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -944,7 +944,7 @@ pub fn process_withdraw(
             stake_program: accounts.stake_program,
         },
         sol_to_withdraw,
-        &[&[]],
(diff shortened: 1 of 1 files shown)


kkonevets (approved):

ruuda (approved):
I don’t understand why this matters, I thought Solana concatenated all the seeds anyway ... but if it fixes the problem, LGTM :woman_shrugging:

PR #583, opened 2022-09-22
Withdraw inactive stake before unstaking

enriquefynn:
When the validator is flagged as inactive, we try to unstake everything from it, but we accumulate some inactive stake when we merge stake accounts, this needs to be withdrawn even if it does not exceed the `MINIMUM_WITHDRAW_AMOUNT`.

### DIFF
--- a/cli/maintainer/src/maintenance.rs
+++ b/cli/maintainer/src/maintenance.rs
@@ -1015,8 +1015,15 @@ impl SolidoState {
             // of Lamports, we don't bother withdrawing. We try to do this
             // so we don't pay more for fees than the amount that we'll
             // withdraw. Or if we have stake to remove from unstake accounts.
-            if expected_difference_stake > SolidoState::MINIMUM_WITHDRAW_AMOUNT
-                || removed_unstake > Lamports(0)
+            // If we're dealing with an inactive validator, we try to withraw
+            // all inactive stake.
+            let minimum_withdraw_amount = if validator.entry.active {
+                SolidoState::MINIMUM_WITHDRAW_AMOUNT
+            } else {
+                Lamports(0)
+            };
+
+            if expected_difference_stake > minimum_withdraw_amount || removed_unstake > Lamports(0)
             {
                 // The balance of this validator is not up to date, try to update it.
                 let mut stake_account_addrs = Vec::new();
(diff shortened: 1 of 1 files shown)


ruuda (approved):
LGTM!

kkonevets:
please take a look (https://github.com/lidofinance/solido/commits/v1_compute_budget) at commits starting at 22f3db7 and to the end. This is a cumulative update of v1 maintainer that includes lots of recent bug fixes also the one that caused a crash after validator deactivation - we need it to be deployed to be able to migrate to v2

cli/maintainer/src/maintenance.rs
```diff
@@ -1015,8 +1015,15 @@ impl SolidoState {
...
-                || removed_unstake > Lamports(0)
+            // If we're dealing with an inactive validator, we try to withraw
+            // all inactive stake.
+            let minimum_withdraw_amount = if validator.entry.active {
+                SolidoState::MINIMUM_WITHDRAW_AMOUNT
+            } else {
+                Lamports(0)
+            };
+
+            if expected_difference_stake > minimum_withdraw_amount || removed_unstake > Lamports(0)
```
kkonevets, line 1026: this will not pass if validator is inactive
```rust
if expected_difference_stake > minimum_withdraw_amount
```
`minimum_withdraw_amount` would be 0 also as `expected_difference_stake`, so `0 > 0` is false.
Take a look [here](https://github.com/lidofinance/solido/commit/578d650dd3dcf3426afb1896e87479cf2f8ba848)
We should recompute effective stake balance after stake merging

kkonevets (changes requested):
I tested this commit and it does not do what we expect

cli/maintainer/src/maintenance.rs
ruuda: Why would `expected_difference_stake` be zero?

> if validator is inactive

Let’s distinguish two things. There is the `.active` field on `Validator`. And there are the stake accounts. The case that was failing happens when the `active` field is `false`, but there is active stake in the stake accounts, which the maintainer tries to deactivate. So in this case, `current_stake_balance` definitely greater than zero. `effective_stake_balance` is the difference between the recorded balance in stake accounts (which will be greater than zero) and unstake accounts (which will still be zero). If there is excess stake to withdraw, then `expected_difference_stake` is going to be greater than zero.

Am I missing something?

kkonevets: Did you test this commit? I did. After stakes are merged the effective stake balance does not change by itself. It is changed only in WithdrawInactiveStake, but this is what we try to call here. Merging two stakes does not account additional rent exempt amount

PR #586, opened 2022-12-05
Verification script

Shi-eld:


ruuda:
The new upstream repository is at https://github.com/lidofinance/solido, I think you might want to open this there instead.

